PR #68, opened 2016-02-17
Added new image querying functions.

kwagyeman:
Now you can get the image width/height. No need to fill your code with constants anymore.

I also made get and set pixel robust. They will not break anymore if the user asks for a pixel off the image. Note... I changed the point argument to a tuple. This was done to keep the API consistent with other objects like rects which are always passed as a tuple.


iabdalkader:
I'm merging this one:

"Merging upstream changes."
No merge commits please, you should rebase or stash changes:
Try
`git pull --rebase`

Also get/set pixel should accept (x, y) why use a list ? It's just more typing to do.

PR #81, opened 2016-02-25
Update blob_detection example for the case there are no blobs

deshipu:
The image.find_blobs function now returns None when there are no blobs,
instead of an empty list. Since we can't iterate over a None, we need to
replace it with an empty tuple in that case.

### DIFF
--- a/usr/examples/blob_detection.py
+++ b/usr/examples/blob_detection.py
@@ -25,7 +25,7 @@ while (True):
     binary.erode(3)
 
     # Detect blobs in image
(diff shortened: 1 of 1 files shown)


iabdalkader:
Hi,
I'm going to fix this in the C wrappers , will make it return an empty list.

deshipu:
Great, I think that it will actually be better, but since the documentation explicitly that it returns None on no match, I assumed there is some reason for that. Thanks!

PR #87, opened 2016-02-28
Dilate and errode working.

kwagyeman:
The old code did not actually implement the erode and dilate kernels
correctly. However, it might have been a little faster because it avoided
the boundary problem.

In the future we can optimize all the kernel code to have different loops
for doing the edges of image versus the center. But, for now, this is
good enough. QVGA color tracking with kernels will be slow, but, the
speed can be improved with QQVGA resolution. Using a 3x3 kernel is
plenty fast. Larger ones are slower.

I also added the ability for you to set the threshold for erode and
dilate. This lets you make the kernel a little bit smarter so that it
won't erode or dilate a pixel unless the threshold is met. Meaning,
you'll be able to use erode to erode an image down to 1 pixel wide
lines.

### DIFF
--- a/src/omv/img/imlib.c
+++ b/src/omv/img/imlib.c
@@ -126,36 +126,6 @@ static void imlib_read_pixels(FIL *fp, image_t *img, int line_start, int line_en
     }
 }
 
-// Main Ram = 196,608 bytes
-// -
-// struct framebuffer = 20 bytes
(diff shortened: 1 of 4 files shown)


iabdalkader:
I think speed is more important than boundaries, anyway thanks! Could you move the script to examples ?

PR #93, opened 2016-03-05
Add support for color gif.

kwagyeman:
Color gifs look very good for how bad you'd expect them to be with just
7 bits of color (rgb232) - quite amazing. Also, I  hardened the gif
module to make it "user ready".

I was going to try to make color gifs using the LAB color space... and that doesn't make sense to do given how many calculations that have to be performed to scale pixel values and etc. Just using the RGB color space seems to work fine.

### DIFF
--- a/src/omv/img/gif.c
+++ b/src/omv/img/gif.c
@@ -4,35 +4,39 @@
  * This work is licensed under the MIT license, see the file LICENSE for details.
  *
  * A super simple GIF encoder.
- * Only support Grayscale/128 colors/No compression.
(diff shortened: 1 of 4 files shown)


iabdalkader:
We could make the LAB script generate 4 bytes per pixel, the 4th byte for RGB232, this brings the table size to 256KBs.

kwagyeman:
Not really necessary. Color gifs look pretty good. I'm happy.

iabdalkader:
For speed.

PR #103, opened 2016-03-24
Redid MLX Code

kwagyeman:
First, a few things:

The MLX 16x4 sensor has just too low of a resolution for mass appeal for
the price. The product is not going to sell very well. We need to look
into supporting sensors with a better res. Like the FLIR 1. The MLX
module was renamed to the "flir" module with this idea in mind.

The flir code now takes care of doing scaling and blending itself. I did
this to get rid of the user having to scale the image themselves and
blend themselves. Its too easy to run out of memory given our current
ultra small heap. In general, anything that requires multiple images in
RAM has got to go. When we do another OpenMV Cam with external RAM in
the MB range then maybe such functions will be safe. But, right now they
are definitely not.

Anyway, moving on, I fixed a few bugs with the MLX math code. But, for
the most part was correct. I also added recommended polling code for
brownouts as required by the datasheet.

Last, I designed this code like the LCD code to support a type value
when inited. This will allow the system to user a different sensor in the
future without any API changes to the user.

I will add test scripts for this next. Basic usage follows:

import flir
flir.init()
flir.display_ir(sensor.snapshot())

And that's it. Super easy. If the user wants the raw temp values they
can use flir.read_ir() to get the ta and to values. The display function
has a hidden alpha and scale argument for controlling blending and the
min/max scaling.

The previous way we worked out scaling kinda sucked... it was a good
shot, but, controllable min and maxes that autoscale by default just
work better. If the user knows the temp range then they can just set the
min and max.

One more thing - scaling is just done using nearest neighbor. Better scaling would have been cool, but, I think, pointless since the res is so low its not really buying anyone anything. We have to get that 80x60 FLIR 1 sensor in the future.

Anyway, longest commit ever done.


kwagyeman:
import sensor, image, time, fir

sensor.reset() # Initialize the camera sensor.
sensor.set_pixformat(sensor.RGB565) # or sensor.GRAYSCALE
sensor.set_framesize(sensor.QVGA) # or sensor.QQVGA (or others)

clock = time.clock() # Tracks FPS.

fir.init()

while(True):
    clock.tick()
    fir.display_ir(sensor.snapshot())
    print(clock.fps())

PR #130, opened 2016-04-23
Improved JPEG capabilities.

kwagyeman:
* Added the ability to control the quality on JPEG functions... However,
due to our JPEG implementation this doesn't seem to help. 90% JPEG
quality images and regular images should be about equal. But, you can
see heavy degredation with 90% still. E.g. text is unreabable. Not
exactly sure why this is happening but it can be fixed later.

* Changed the compress() function to compressed(). Also, it now
compresses using FB_Alloc to prevent realloc issues when compressing.

* Added new compress() function. This function compresses an image in
place and if that image is the frame bufffer then it will update the
frame buffer bpp value to reflect the image was compressed. Users can use
this function to basically finalize the frame buffer and then pass the FB
to functions that need to send image bytes. The benefit of using this
function is that it should allow higher quality JPEGs and let everything
run at a faster speed while connected to the IDE.

I made this function to speed up WiFi. However, I encountered a bug with
the winc.send() method. It appears to zero the bytes it sends. I didn't
debug further except to verify that the image data became zero after
calling send.


iabdalkader:
Did you test this ? If you compress the FB, then sensor_snapshot will re-compress JPEG data on the next frame, the IDE will read garbage. Also I'm not sure how the new compress makes WiFi faster ?

kwagyeman:
Yes, I always test my PRs. I modified the sensor snapshot function not to compress images if they are JPEG already. The code works.

The new compress makes WiFi faster when connected to the IDE because you don't have to compress twice... however, until the send() zeroing the data buffer issue is fixed you can't use the command and have to use a temporary buffer.

Anyway, please merge, I've got more stuff in the works.

Issue #147, opened 2016-08-04
Allow configuration of FIR refresh rate & ADC resolution


wminarik:
(for the Thermopile Shield)

Currently within py_fir.c the MLX90621 initialization is set to:
64 Hz refresh rate
18-bit ADC resolution

I suggest allowing these parameters to be set on the call to fir_init(), either by adding arguments to this function or by adding a new function (fir_config()?) that takes a valid refresh rate (0.5 to 512 Hz) and a valid ADC resolution (15-18 bits).

Reasoning: the noise level of MLX90621 corner pixels decreases from ~ 1°C at 64 Hz to less than 0.1°C at 1 Hz. When the frame rate is slower than 64 Hz, the thermal images are noisier than need be. Stationary objects would benefit from a lower noise floor.

The current 18-bit ADC resolution gives the smallest temperature increment within range, but limits the maximum object temperature to ~450°C. Objects within the field hotter than this may return invalid data due to measurement overflow (according to the datasheet). There may be applications in which it's desired to have hotter objects within the field of view (up to a maximum of ~950°C with a 15-bit ADC resolution setting).

kwagyeman:
Yes, we can easily do this. However, it you want it done sooner than later can you submit a code patch for this feature? I wrote the MLX code such that all the calcs are done during the init time. It shouldn't really be that much work to add this feature in.

Otherwise... this is rather low priority for me and it won't get done for a long time.

iabdalkader:
Closing this because we're not supporting FIR MLX anymore.

PR #149, opened 2016-08-18
changes fir_init() to take specified refresh rate and ADC resolution

wminarik:
Changes fir_init() (file py_fir.c) to take three arguments:
fir_init(*[refresh,resolution, type=1]*)
where:
refresh: refresh rate in Hz, between 1 and 64 Hz (64 Hz default)
resolution: ADC resolution in bits, between 15 and 18 (18-bits default)

following the rational in [#147](https://github.com/openmv/openmv/issues/147)

The existing *type=1* argument was placed last, as I believe that dictionary arguments (`*kw_args`) must follow scalar args (`*args`)

I'm unpracticed with writing C for Python, especially within a large calling framework -please make suggestions!

### DIFF
--- a/src/omv/py/py_fir.c
+++ b/src/omv/py/py_fir.c
@@ -208,6 +208,18 @@ static mp_obj_t py_fir_deinit()
     return mp_const_none;
 }
 
+/*
+Allows the refresh rate to be set in the range 1 Hz and 48 Hz, in powers of 2. (48 Hz default)
+The MLX90621 sensor is capable of a larger range but these extreme values are probably not useful with OpenMV.
+
+Allows the ADC precision to be set in the range of 15 to 18 bits. (18 bit default). Lower ADC precision
+allows a large maximum temperature within the scene without sensor overflow.
+ADC 18-bits: max scene temperature ~450C, 15-bits: max scene temperature ~950C
+
+calling:
+ fir.init()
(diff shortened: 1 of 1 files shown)


iabdalkader:
merged, I'll use keywords instead of positional args, anyway thanks!

wminarik:
Thanks -you've taken my awkward code and made it elegant; thanks also for introducing me to built in intrinsic functions.

One bug: The current algorithm works well for user input that is a power of 2. For arbitrary integers, the algorithm gives less straightforward results -in fact all odd numbers set the refresh rate to 0.5 Hz. I suggest the following change:

under "//sanitize values":
change:
`IR_refresh_rate = 14 - __CLZ(__RBIT((IR_refresh_rate > 512) ? 512:(IR_refresh_rate < 1)? 1:IR_refresh_rate));`
to
`IR_refresh_rate = __CLZ((IR_refresh_rate > 512) ? 512:(IR_refresh_rate < 1)? 1:IR_refresh_rate) - 17;`

this saves a bit reversal step and sets any user input between two powers of 2 to the lower valid refresh rate (that is, refresh=10 Hz is set to 8 Hz).

Cheers,
Bill

iabdalkader:
Hi, sorry got busy with other issues, feel free to send another PR if you want.

PR #150, opened 2016-08-26
modified refresh rate processing code to accept inputs not = 2^n

wminarik:
Changes the code in order to set any user input between two powers of 2 to the lower valid refresh rate which must be a power of two (that is, an input of refresh=10 Hz is set to 8 Hz).


iabdalkader:
@wminarik Hi, thanks for the PR, however I can't merge this because it has a "merge commit" you need to rebase `git pull --rebase` in the future to void of that.

iabdalkader:
Closing this because we're not supporting the old FIR shield anymore.

PR #157, opened 2016-12-25
QR Code Detection + Start of New LIB

kwagyeman:
This PR brings in the start of the New LIB. An effort on my part to cleanup the OpenMV code, make it faster, and in general address problems that have been simmering for a while. New LIB integration will be done slowly and surely to replace old code with better code. Anyway, I put all New LIB changes in the first two commits. Integration of the New LIB with the OpenMV code is only done in the last commit.

The last commit, the integration commit, brings in new blob detection code which uses a scanline based flood fill algorithm that is faster than the previous algorithm - see this for a comparison http://will.thimbleby.net/scanline-flood-fill/. Additionally, find_blobs returns a list of blob objects now which have named values so you don't have to use array accesses everywhere anymore. The new blob objects can be sliced however so that old code continues to work.

And... for the main course, qr code detection. This works now... and it is awesome. I've created a demo script to show off this feature. The qrcode function also returns a new qrcode object like blob detection which features named properties now. The qrcode detection algorithm can find multiple qrcodes in view and does not slow the frame rate down unless there are a lot of QR codes in view. Additionally, it does not require the qr code to be aligned with the camera. It can handle qrcodes that are rotated, scaled, sheared, etc. all with respect to the camera.

In fact, after reading the qr code I think I know how to do a lot of stuff better now... in particular, I re-did the blob code with the flood fill algorithm after reading the QR code.


iabdalkader:
Sorry I can't merge all that. Please keep the PRs very specific, 1 commit of related changes max. Also I don't like all the changes to image library stuff, it all works fine, if you see a problem, bug, improvement just send PRs to fix those one at a time.

PR #182, opened 2017-01-30
Add flag for SWD programmer.

kwagyeman:
It can now check if the self-test passed.

### DIFF
--- a/src/omv/main.c
+++ b/src/omv/main.c
@@ -66,6 +66,8 @@
 #include "py_lcd.h"
 #include "py_fir.h"
 
+#include "framebuffer.h"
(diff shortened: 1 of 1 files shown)


iabdalkader:
Sure you want to use BPP and not the CRC reg ? I'm not entirely sure how this might affect code that checks MAIN_FB->BPP, like sensor and fb_alloc.

kwagyeman:
Can't use the CRC reg without resetting the hardware for that. I tried the CRC reg first and all writes to it failed.

As for any side affects. Not that this only runs after the self test. Afterwards it doesn't run again.

PR #195, opened 2017-03-13
Fix volatile issue.

kwagyeman:
This flag has to be set for the SWD debugger to pass. This worked for
the M4 without volatile but failed for the M7 without the keyword.

As for how this worked before. I don't know. Anyway, good this was caught now.


iabdalkader:
I'll add cache clean after setting the flag in main.c

PR #219, opened 2017-04-28
Add chinese readme

rmokerone:
Modified the README syntax error, and added the Chinese translation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ### OpenMV (Open Machine Vision Module)
 
-![简体中文版](README_CN.md )
+[简体中文版](README_CN.md )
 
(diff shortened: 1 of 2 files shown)


iabdalkader:
Hi,

Thanks for the PR, what's oopy.org and why do you add it to our readme ?

rmokerone:
oopy.org is an open source hardware organization in China. We add this URL in this document, only to illustrate that this document is translated(翻译) by the oopy organization:-D

iabdalkader:
I'm sorry but we can't reference an organization which we're not affiliated with, plus the README is really outdated, I'm going to update it in the following days, if you would like to translate the new one without references you're welcome.

rmokerone:
I am very happy to contribute to this project.I did not notice the issue of the organization. If there is a mistake, please forgive me.Thanks again.

PR #220, opened 2017-04-28
Add image.raw() function.

rmokerone:
Sometimes we need to get raw pixels. So, I wrote such a function, can return   the original information. The function returns bytes and prints the number of bytes.
```
>>> img = sensor.snapshot()
>>> dat = img.raw()
153600
>>> type(dat)
<class 'bytes'>
>>> data[0:10]
b'\x84\x91\x84\x92\x9c\x92\x9c\x92\x9c\x92'
```


kwagyeman:
Interesting. Um, question:. Images can be accessed directly using [] indexing which returns the pixels. What's the rational behind raw byte access versus raw pixel access which is already built-in?

iabdalkader:
@kwagyeman @rmokerone That's true, I implemented the `subscr` function in image class long time ago and forgot about it, I'll revert this PR.

kwagyeman:
I'm okay with adding a raw function for total byte access. Just want to know why though. If you want to serialize images you can just pass them to functions which send data because images are treated as buffer objects for uart.write()/spi.send() etc.

rmokerone:
@iabdalkader Thank you for your reply.I have seen ``subscr`` function, but this function can only return one pixel at a time.I want to access multiple pixels at once. Perhaps I can rewrite the ``subscr`` function, let it return a ``bytes`` type.

rmokerone:
@kwagyeman Thank you very much, I did not know image types can be passed directly to the ``spi.write()`` .Now know, thanks again.

kwagyeman:
Hi, I'll make the subscr function allow splices so you can get multiple bytes at a time. However, this will use a lot of ram for a large splice. So be careful. I'll push a PR for this tomorrow.

rmokerone:
@kwagyeman See this feature very happy. thank you~

PR #232, opened 2017-05-26
Add more resolution combinations.

kwagyeman:
With the new frame rate speed increase folks will be asking for smaller
resolutions to get 85 FPS or so when running an algorithm. This commit
adds all scaled modes of frame sizes we already support. We should be
good now on frame sizes for the present and future now.

Todo - skip frames does not run long enough anymore for auto white
balance and gain to stabilize before they are turned off in some scripts.
This needs to be adjusted.

---

Yes, this is tested.


kwagyeman:
Removed the extra QQVGA2 reses.

PR #241, opened 2017-06-11
Refactor framebuffer

kwagyeman:
Moved structs along with image copying code from sensor into
framebuffer.c so that we can use the new copy_fb_to_jpeg_fb() function
in the image library for methods with "copy_to_fb" so that they update
the IDE preview when called. In particular, I need for for my new ImageWriter/Reader classes.
Will send them once this PR is accepted.

Also, I noticed that the MAIN_FB_SIZE() value is not calculated
correctly in all cases. Will fix later. Trying to keep this commit clean
for just the refactoring.

All changes have been tested. Too.

### DIFF
--- a/src/omv/framebuffer.c
+++ b/src/omv/framebuffer.c
@@ -1,0 +1,75 @@
+/*
+ * This file is part of the OpenMV project.
+ * Copyright (c) 2013/2014 Ibrahim Abdelkader <i.abdalkader@gmail.com>
+ * This work is licensed under the MIT license, see the file LICENSE for details.
+ *
+ * Framebuffer stuff.
+ *
+ */
+#include "imlib.h"
+#include "omv_boardconfig.h"
+#include "framebuffer.h"
+// If buffer size is bigger than this threshold, the quality is reduced.
+// This is only used for JPEG images sent to the IDE not normal compression.
+#define JPEG_QUALITY_THRESH     (160*120*2)
+
+extern char _fb_base;
+framebuffer_t *fb = (framebuffer_t *) &_fb_base;
+
+extern char _jpeg_buf;
+jpegbuffer_t *jpeg_fb = (jpegbuffer_t *) &_jpeg_buf;
+
+void copy_fb_to_jpeg_fb()
+{
+    static int overflow_count = 0;
+
+    if ((fb->bpp > 3) && JPEG_FB()->enabled) {
+        // Lock FB
+        if (mutex_try_lock(&JPEG_FB()->lock, MUTEX_TID_OMV)) {
+            if(OMV_JPEG_BUF_SIZE < fb->bpp) {
+                // image won't fit. so don't copy.
+                JPEG_FB()->w = 0; JPEG_FB()->h = 0; JPEG_FB()->size = 0;
+            } else {
+                memcpy(JPEG_FB()->pixels, fb->pixels, fb->bpp);
+                JPEG_FB()->w = fb->w; JPEG_FB()->h = fb->h; JPEG_FB()->size = fb->bpp;
+            }
+
+            // Unlock the framebuffer mutex
+            mutex_unlock(&JPEG_FB()->lock, MUTEX_TID_OMV);
+        }
+    } else if ((fb->bpp > 0) && JPEG_FB()->enabled) {
+        // Lock FB
+        if (mutex_try_lock(&JPEG_FB()->lock, MUTEX_TID_OMV)) {
+            // Set JPEG src and dst images.
+            image_t src = {.w=fb->w, .h=fb->h, .bpp=fb->bpp,            .pixels=fb->pixels};
+            image_t dst = {.w=fb->w, .h=fb->h, .bpp=OMV_JPEG_BUF_SIZE,  .pixels=JPEG_FB()->pixels};
+
+            // Note: lower quality saves USB bandwidth and results in a faster IDE FPS.
+            bool overflow = jpeg_compress(&src, &dst, JPEG_FB()->quality, false);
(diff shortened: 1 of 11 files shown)


src/omv/framebuffer.c
```diff
@@ -0,0 +1,75 @@
...
+ */
+#include "imlib.h"
+#include "omv_boardconfig.h"
+#include "framebuffer.h"
+// If buffer size is bigger than this threshold, the quality is reduced.
+// This is only used for JPEG images sent to the IDE not normal compression.
+#define JPEG_QUALITY_THRESH     (160*120*2)
+
+extern char _fb_base;
+framebuffer_t *fb = (framebuffer_t *) &_fb_base;
```
iabdalkader: There's a problem with allocating the buffer this way, it's global, can't have other fbs or jpeg_fbs.

EDIT1: And if you make it static we'll have to use getters fb_get_main_fb() and fb_get_jpeg_fb(). could add those functions and #define MAIN_FB() fb_get_main_fb()

EDIT2: So if you add getters and #define the FB macros, this way it won't break the code and it gets refactored to use the getters. I suggest **fb_get_raw/main_buffer()** and **fb_get_jpeg_buffer()**

src/omv/framebuffer.h
```diff
@@ -36,4 +38,8 @@ static struct jpegbuffer {
 
 // Use this macro to get a pointer to the free SRAM area located after the framebuffer.
 #define FB_PIXELS() (MAIN_FB()->pixels+MAIN_FB_SIZE())
+
+// Transfers the frame buffer to the jpeg frame buffer if not locked.
+void copy_fb_to_jpeg_fb();
```
iabdalkader: If you're going to make framebuffer.c, please prefix fb functions (ex. fb_ or framebuffer_).
Note this function is mislabeled and misused (will explain more in comment on sensors.c). I think it should be called fb_update_jpeg_buffer().

src/omv/sensor.c
```diff
@@ -712,55 +708,7 @@ int sensor_snapshot(image_t *image, line_filter_t line_filter_func, void *line_f
...
-                }
-                // Set FB from JPEG image
-                JPEG_FB()->w = dst.w; JPEG_FB()->h = dst.h; JPEG_FB()->size = dst.bpp;
-            }
-
-            // Unlock the framebuffer mutex
-            mutex_unlock(&JPEG_FB()->lock, MUTEX_TID_OMV);
-        }
-    }
+    if (sensor.pixformat != PIXFORMAT_JPEG) copy_fb_to_jpeg_fb();
```
iabdalkader: So one problem here, this breaks OV2640. If the main FB is already compressed it should be copied to the JPEG buffer for the IDE preview. Remember USBDBG checks and sends JPEG FB, so you should call this unconditionally.

src/omv/framebuffer.c
```diff
@@ -0,0 +1,75 @@
...
+                    JPEG_FB()->quality = IM_MAX(1, (JPEG_FB()->quality/2));
+                }
+                JPEG_FB()->w = 0; JPEG_FB()->h = 0; JPEG_FB()->size = 0;
+            } else {
+                if (overflow_count) {
+                    overflow_count--;
+                }
+                // No buffer overflow, increase quality up to max quality based on frame size
+                if (overflow_count == 0 &&
+                        JPEG_FB()->quality < ((MAIN_FB_SIZE() > JPEG_QUALITY_THRESH) ? 35:60)) {
```
iabdalkader: Maybe QUALITY_THRESH and the low/high factors should be moved to board files.

src/omv/framebuffer.c
kwagyeman: I don't understand why you are requesting this. The previous method was creating a static pointer var in each c file the header was included in that all pointed to the same frame buffer and jpeg buffer. My moving these values into their own C file I'm freeing up a lot of .data waste. As for multiple frame buffers... not sure I follow you there. Is this something we need to support now? I'm just trying to do a re-factor here. Not change how everything works. Anyway, changing the name of the stuff from "fb" will break other code in the firmware. There are quite a few places that don't use MAIN_FB().

src/omv/framebuffer.c
kwagyeman: Will do.

src/omv/sensor.c
kwagyeman: Please understand I was following the original code functionality. This means it was broken before. I can remove the if().

src/omv/framebuffer.h
kwagyeman: Okay.

src/omv/sensor.c
kwagyeman: Oh, I see what was going on now here. I did an update before to make it copy the jpeg buffer. But, kept previous code. This if part should have been dropped before.

iabdalkader: Yes, I don't know if it was broken before or now but it is broken.

src/omv/framebuffer.c
iabdalkader: Yeah I know how the current method works (I wrote it) and I know it takes space (but Not a lot of data, 8 bytes per include), and I wanted to move those to .c but this contaminates the namespace (can't have another variable called fb for example). Just saying consider either prefixing or using getters.
I'd prefer getters, because this allows us in the future to add code (if needed) before getting the FB.
Re code that will break, it should be refactored first, since we're refactoring anyway let's just fix it, I can do that if you want.

kwagyeman: Okay, will go through all code where fb is used and replace with MAIN_FB() which uses a getter.

PR #254, opened 2017-07-10
Support for new Haar XML format

normen:
So this is my PR for the new Haar XML format with a big caveat - it doesn't seem to work (UPDATE - It works now!) I changed the importer in such a way that it still supports the old files just like before, thats why I made a PR already.

- It doesn't support creating the header files for the new format yet because I can't test them
- It does allow showing the info and creating binary files from both the new and old XML format
- It recognizes the format by itself
- (FIXED) The binary files created from the new format load in the OpenMV cam but don't track anything
- (FIXED) The new format seems to have a different count of features in the rects and the weakClassifiers, this might be the issue?

Edit: Oh and I also fixed a bug where the -i parameter would create a .cascade file :)


normen:
I guess the answer is hidden somewhere in this file: https://github.com/opencv/opencv/blob/master/apps/traincascade/cascadeclassifier.cpp

At the bottom where it saves for the old format it uses some kind of index to pick the correct rect and for the new format it seems to leave out "unused" rects/features?

..but its too late over here for me to make sense of it all.. :) If you have a hint I'll try and complete the PR (sans the header generation).

normen:
Its working! Cats! Detected! Me -> Tired & Happy o__O

iabdalkader:
Hi, thanks for the PR, I'll test it tonight :+1:

normen:
Lol, looking at the code after sleeping a while I see I do a completely unnecessary loop over the features array while using a counter in that loop to access the array index.. Instead of simply using the array index ;) I'll clean that up.

iabdalkader:
Hi, I tested the new converter vs the old one using the old format, and it generates the same cascade. Will test the new format later. Thanks a lot of the help :+1:

iabdalkader:
Yep it's working :)

cat

normen:
Aww what a cutie :)

As said I didn't change the header generation but I guess you'll only need that when you plan to add new built-in cascades and then you can change it yourself based on my changes, I guess this is what most users will want.

All the best with the project!

PR #259, opened 2017-08-23
Update to latest openmv API

neurodroid:
Make this script work as a standalone program, and update to latest openmv API

### DIFF
--- a/usr/openmv-fb.py
+++ b/usr/openmv-fb.py
@@ -1,24 +1,47 @@
 #!/usr/bin/env python
 import sys
-import usb.core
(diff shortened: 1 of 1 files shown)


iabdalkader:
Hi,
Thanks for the PR. This script is not used or maintained anymore, I kept it around as an example for using pygame to display the framebuffer. If you're interested in fixing it more, you should add calls stop_script() enable_fb() after calling init:

```
#Interrupt main.py
openmv.stop_script()

# Enable Framebuffer
openmv.enable_fb(True)
```

PR #357, opened 2018-06-22
Make both networks output the same value types.

kwagyeman:
Both CIFAR and LENET work still.

The smile network... I couldn't really get to work before or afterwards.
I noticed the Haar Cascade one has trouble finding my face. Maybe fix via using
the contrast settings of the previous Haar scripts?

Also, does the smile network need the whole face? Because the bounding rect on my face is missing my forehead.

...

Note, I want to get this fix in before the release otherwise we're stuck with the unscaled values for life. Since the scaling is done with floats this mod is reversible in python if the user needs to undo it.


iabdalkader:
>> Also, does the smile network need the whole face?

Not the training images are tightly cropped around the face, basically just needs the eyes + mouth, actually I just made the ROI smaller works a lot better now.

PR #366, opened 2018-08-09
Wifi updates

kwagyeman:
AP mode works with the AP alright. STA mode needs help. Not sure what's wrong and I don't have an SWDIO debugger to find out.

Here's the latest IDE that support everything. OpenMV IDE will create the settings files and etc.

http://upload.openmv.io/openmv-ide-windows-2.1.0/openmv-ide-winodws-2.1.0.exe
http://upload.openmv.io/openmv-ide-linux-x86_64-2.1.0/openmv-ide-linux-x86_64-2.1.0.run

Note that the reason for AP and STA mode settings having different names is so that you don't have to retype them if you want to change modes in the IDE...

### DIFF
--- a/src/omv/boards/OPENMV2/imlib_config.h
+++ b/src/omv/boards/OPENMV2/imlib_config.h
@@ -12,6 +12,9 @@
 // Enable flood_fill()
 //#define IMLIB_ENABLE_FLOOD_FILL
 
+// Enable mean()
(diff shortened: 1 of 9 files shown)


iabdalkader:
Merged 0413d4bbd209b5e93566270124d4b485d61995bf

PR #368, opened 2018-08-16
SPI timeout in slave mode, due to internal chip error

miiicmueller:
See stm32f7 erratasheet, chapter 2.11.2.
This is a proposition that worked for us, with clk speed of 10MHz.
Maybe it wont be sufficient with a higher clk speed...


kwagyeman:
Wow, thanks for spending a PR. I'll let Ibrahim approve but helping on the code is appreciated!

PR #400, opened 2018-10-28
Port ffwrapper and other omv file operations to the micropython vfs

andrewleech:
The existing implementation goes directly to oofatfs which limits file operations to stm32 port  and files on the the fatfs drives only.

This change allows any micropython filesystem to be used for all file IO in the omv subsystem, including virtual file systems defined via a python interface (eg mprepl)

This patch has not been tested explicitly in this submitted form, however it's been used a fair bit in my omv_module repo form (see below). 
This pr is intended as a hopefully useful contribution though even if it needs some further fixes, but also as a conversation starter.

At work I'm building a product which includes much of the omv folder as a submodule built directly against upstream micropython, as described in:  
https://github.com/micropython/micropython/pull/4195
https://gitlab.com/alelec/mp_omv
https://gitlab.com/alelec/omv
This has been restructured as such partly as a test case, but mostly as we need to use other features from mainline micropython (and I'm on a personal mission to reduce the need for micropython forks :-)

I've got a bunch of other changes in the omv repo above which may be of use to you too, most of which are for compatibility both with mainline stm32 but also to add support for use with the unix port for automated unit testing of imaging functions. I'm happy to help with anything there of interest.

Note: my mp_omv system currently breaks support for the openmv ide but I'd like to add it back using the second cdc port now optionally available in mainline micropython, rather than the existing method of interleaving it into the main cdc... this hasn't been started yet though.

On our private work fork we've also got initial support for a 12bit 1.3mp image sensor (ar0130) which I could discuss pushing up if there's any value for you (it needs the sdram support too to fit the larger images). Thanks for your fantastic platform!

### DIFF
--- a/src/omv/ff_wrapper.c
+++ b/src/omv/ff_wrapper.c
@@ -6,99 +6,170 @@
  * File System Helper Functions
  *
  */
-#include <mp.h>
+#include <py/builtin.h>
+#include "mp.h"
 #include "common.h"
 #include "fb_alloc.h"
 #include "ff_wrapper.h"
+
 #define FF_MIN(x,y) (((x)<(y))?(x):(y))
 
-NORETURN static void ff_fail(FIL *fp, FRESULT res)
+NORETURN static void ff_fail(file_t *fp, FRESULT res)
 {
-    if (fp) f_close(fp);
+    if (fp && fp->file) mp_stream_close(fp->file);
     nlr_raise(mp_obj_new_exception_msg(&mp_type_OSError, ffs_strerror(res)));
 }
 
-NORETURN static void ff_read_fail(FIL *fp)
+NORETURN static void ffile_read_fail(file_t *fp)
 {
-    if (fp) f_close(fp);
+    if (fp && fp->file) mp_stream_close(fp->file);
     nlr_raise(mp_obj_new_exception_msg(&mp_type_OSError, "Failed to read requested bytes!"));
 }
 
-NORETURN static void ff_write_fail(FIL *fp)
+NORETURN static void ffile_write_fail(file_t *fp)
 {
-    if (fp) f_close(fp);
+    if (fp && fp->file) mp_stream_close(fp->file);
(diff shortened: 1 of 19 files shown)


kwagyeman:
Hi, could you fix the travis CI issues? Your changes fail to compile on all OpenMV Cam targets. Otherwise, it looks like you just did a lot of find and replace on things which is okay.

andrewleech:
Yeah sorry there are a number of build failures, I'll clean them up.

The bulk of the notable changes are in `ff_wrapper`, reworking that to target the micropython fs interfaces rather than oofatfs. This is a higher level interface present in all micropython ports rather than going direct to oofatfs.
The rest of it really us just renaming as you noted to used the updated ffwrapper layer.

kwagyeman:
Ibrahim will have to approve this after all compiles pass. I'm not sure what the exact implications are. However, we appreciate the PRs and can use any help on features. Thanks!

andrewleech:
Yeah no pressure or anything, I've just been trying to find some things from my tree that make sense to push up in case it's useful. Most of my my changes are to support a larger camera sensor with more pixel formats, which isn't much use if you don't have hardware to suit. 
This PR is also part of a set of changes to allow building the unix port though which may be of interest?

I use mprepl a lot to mount my pc's local dir in micropython (get's exposed as a virtual filesystem), this filesystem change lets me read/write images directly to folder on PC over the usb.

Going via the higher level interface though is surely going to have some overhead though which I haven't profiled. I also know you've got your filesystem buffering, and micropython has it's write pagefile buffer, I'm not sure if these are going to be effectively duplicating/fighting each other now too.

kwagyeman:
A unix port would be very helpful. We'd love to have the software run on unix (which isn't particular hard but we don't have time). As for the pagewrite buffer... that must be new. We added the buffering because things were slow without it. That code is kinda only now however. 2 years now. Anyway, yeah, we love getting PRs that make sense for us to merge. Thanks!

iabdalkader:
Closing this please reopen if there's an update.

PR #401, opened 2018-10-29
jpeg: ensure buffer is deallocated on compression failure

andrewleech:
I've had a number of situations where the jpeg compression fails due to not enough free memory, which is to be expected.
I noticed however that in these cases, the memory used during the compression attempt was not getting freed so I couldn't just re-try the compression at a lower setting.

This PR should address this issue by ensuring the allocated memory is freed before the exception is raised.

### DIFF
--- a/src/omv/py/py_image.c
+++ b/src/omv/py/py_image.c
@@ -1070,7 +1070,11 @@ static mp_obj_t py_image_compressed(uint n_args, const mp_obj_t *args, mp_map_t 
     fb_alloc_mark();
     uint8_t *buffer = fb_alloc_all(&size);
     image_t out = { .w=arg_img->w, .h=arg_img->h, .bpp=size, .data=buffer };
-    PY_ASSERT_FALSE_MSG(jpeg_compress(arg_img, &out, arg_q, false), "Out of Memory!");
+    if(jpeg_compress(arg_img, &out, arg_q, false)) {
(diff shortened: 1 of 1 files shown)


kwagyeman:
Hi, I know what you are trying to fix here. But, the act of throwing an exception triggers the memory deallocator. This isn't actually necessary. If you check the MP code under our fork you'll find that the fb_free_till_mark method is called in the NLR longjump handler.

See here: https://github.com/openmv/micropython/blob/901023b150448c9815a51cbb8782ab8c9809a3c3/py/nlr.h#L131

https://github.com/openmv/micropython/blob/901023b150448c9815a51cbb8782ab8c9809a3c3/py/nlr.h#L143

kwagyeman:
Sorry, if this was kinda hidden :). But, we actually prevent ALL memory from leaking in the system. This is why we don't dealloc memory in our code ever. Because this little hook is here.

andrewleech:
Ah I see, thanks for the clarification!

kwagyeman:
The way this stuff works, it makes it really easy to write the C code. Effectively all malloc calls either succeed or an exception is triggered and memory is freed. Then leaked stuff in the python heap is garbage collected.

andrewleech:
Yep that's really very neat. The joys of working in a python system, it can even make c better!

PR #409, opened 2018-11-04
add motor shield stepper driver and example

kidswong999:


kwagyeman:
Cool, Ibrahim will merge.

iabdalkader:
@kidswong999 Hi, can you move motor_shield_stepper.py to scripts/libraries/ ?

kidswong999:
@iabdalkader OK. 
Emm, but what is the role of script/libraries? I noticed that pca9685.py is in script/examples.

iabdalkader:
@kidswong999 scripts/libraries is for libraries (modules you import). The IDE will search both examples and libraries, but if it can't copy to the camera it's easier to find imports in scripts/libraries. pca9685.py should be in libraries. If you want to send a PR to move pca to libraries go ahead.

kidswong999:
I add a new PR. 
https://github.com/openmv/openmv/pull/410

It is more convenient to directly compile the commonly used libraries into firmware (such as pca9685, tb6612).
If the flash is enough. : )

kwagyeman:
Hey, I just merge libraries into the example folder in the IDE. Doing this for the GitHub matters little.

iabdalkader:
> It is more convenient to directly compile the commonly used libraries into firmware (such as pca9685, tb6612).
> If the flash is enough. : )

The M4 flash is full, so we'll have to disable those libraries for the M4 and then include them with the IDE again as scripts, so probably best to leave them as scripts for now.

PR #431, opened 2019-01-22
Create vl53l1x.py

kidswong999:
I wrote the MicroPython driver for vl53l1x range sensor, it works perfect on OpenMV.


kidswong999:
There is a 90 bytes `VL51L1X_DEFAULT_CONFIGURATION`. 
I don't know if there is any optimization, such as putting the bytes in flash.

kwagyeman:
Should be fine to merge. I'll let Ibrahim do it.

PR #434, opened 2019-01-27
Fix M4 Compile and re-enable methods at the cost of JPEG speed.

kwagyeman:
Made the YUV conversion done in software now versus a LUT on the M4. Speed is lower but we can fix more features. A fair trade for mostly debug functionality.


iabdalkader:
Merged most of the commits.

PR #463, opened 2019-03-03
Kwabena/finalize flir lepton support

kwagyeman:
This PR contains the following features for the FLIR Lepton:

1. Adds the ability to change the color rainbow table LUT to the ironbow FLIR Lepton style LUT to make the image look **cool**. I updated the script that generates the rainbow table too.
2. Add methods to allow the user to get information about the FLIR Lepton attached
3. Add methods to allow the user to execute FLIR Lepton API functions so that they can mess with the module settings easily without us having to do any work on the firmware image.
4. Add the ability to turn the AGC off so that you can set a temperature range for the FLIR Lepton to look for. This allows you to actually use the sensor for real applications.
5. Added a ton of example scripts showing off how to use all this stuff.
6. Tested the software with the FLIR Lepton 1.0/2.0/2.5/3.5.


iabdalkader:
This doesn't work and when ioctl is added (or when PR #465 is merged) it will break this PR anyway.

PR #473, opened 2019-03-07
Avoid invalid mem access by validating post-rounded indices

peitschie:
These bounds checks were incorrect if sourceX/Y rounds up. In this situation, the unrounded source will be smaller than the limit by a factional amount (C will elevate the int bounds to a float for comparison), but the post-rounded version will equal the limit.

E.g.,
sourceX = 10.5
img->w = 11 (i.e., valid indices are [0-10])
sourceX2 = 11, which is invalid memory

### DIFF
--- a/src/omv/img/apriltag.c
+++ b/src/omv/img/apriltag.c
@@ -12078,7 +12078,7 @@ void imlib_rotation_corr(image_t *img, float x_rotation, float y_rotation, float
                     int sourceX2 = round(sourceX);
                     int sourceY2 = round(sourceY);
 
-                    if ((0 <= sourceX) && (sourceX < img->w) && (0 <= sourceY) && (sourceY < img->h)) {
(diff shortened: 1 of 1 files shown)


kwagyeman:
Is this to resolve the bug with the rotation method not doing perfect 90 degree rotations? I fixed another bug involving the round more liberally than I should which makes me guess this is the issue affecting this code. I haven't gotten around to fixing it yet.

peitschie:
The fix was purely to avoid the invalid memory being accessed by this method. The bounds-checks meant that images rotated by unlucky amounts tended to get a line of white pixels (well... uninitialized memory/noise) near the upper bounds of the image.

I didn't look too hard at whether the rounding should have been done another way, just wanted to get rid of the clearly-bad data being added to the image.

kwagyeman:
Okay, it should be merged in then.

PR #486, opened 2019-03-12
Fix Lepton reset.

iabdalkader:


src/omv/lepton.c
```diff
@@ -402,9 +396,9 @@ static int reset(sensor_t *sensor)
         }
     }
 
-    radiometry = rad == LEP_RAD_ENABLE;
```
kwagyeman: This was first because the variable is declared first and was trying to keep order...

PR #502, opened 2019-04-05
Fixup MAVLink optical flow script after testing with ArduPilot

rmackay9:
This PR makes a few small but important changes to the example script for sending optical flow message (via mavlink) to mavlink enabled flight controllers like those running [ArduPilot](http://ardupilot.org/).

The fixes are:

- fix to the direction (one axis was reversed) and scaling (the values were 3 to 5 times too small) of the flow sensor values being sent
- always send packets regardless of whether the sensor has seen much movement.  This is important because the flight controller side performs best with regular updates.  One reason is that without messages it can't determine if the sensor has failed or not (i.e. does no message mean the sensor has failed or that it can't see anyway?).  Also the Extended Kalman Filter does best with regular samples.
- added a green flashing LED which makes it more obvious if the script is running or not.  This is very optional and feel free to drop this commit if you're not happy with it.

By the way, here's a video of the camera in action on my quadcopter: https://www.youtube.com/watch?v=BmixVDBiIbA

### DIFF
--- a/scripts/examples/18-MAVLink/mavlink_opticalflow.py
+++ b/scripts/examples/18-MAVLink/mavlink_opticalflow.py
@@ -49,8 +49,8 @@ def send_optical_flow_packet(x, y, c):
                        0,
                        0,
                        0,
-                       int(x * 10),
-                       int(y * 10),
+                       int(x),
+                       int(y),
                        MAV_OPTICAL_FLOW_id,
                        int(c * 255))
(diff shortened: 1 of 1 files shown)


rmackay9:
thanks!

kwagyeman:
Can you update the AprilTag script to with similar changes in another PR @rmackay9.

rmackay9:
@kwagyeman, yes will do.

PR #523, opened 2019-05-02
optimised solver for cifar10/cifar10_fast

thyrrestrup:
This enables multistep for the solver file and removes the need for multiple solver files.
This also reduces the number of times the setup procedure/network initialisation is done - from 3 to 1.
Also simplifies the train scripts.


kwagyeman:
Looks fine. I'm okay to merge.

PR #530, opened 2019-05-16
pyopenmv extended to support multiple cameras

ekrell:
Instead of a single global '__serial', this version has a list of them. Another list has just the port name. The two lists are parallel in that the index of one corresponds to the index of the other. So, any serial can be accessed by  searching for the port name.

I have been using this version for combining multiple cameras for OpenCV processing. Perhaps others will find it useful.


iabdalkader:
Thanks!

PR #554, opened 2019-07-21
Kwabena/ov2640 fixes

kwagyeman:
* Images are right side up now.
* Picture quality is acceptable (not as good as the OV7725)
* All auto functions work now (the OV2640 ignores exposure control
however)
* Added XGA frame size.
* JPEG mode is enabled but needs work still (not sure if the H7 hardware
can capture the packet stream fast enough not to drop bytes).
* Fix bayer/jpeg image print statement.


src/omv/sensor.c
```diff
@@ -996,7 +1000,13 @@ int sensor_snapshot(sensor_t *sensor, image_t *image, streaming_cb_t streaming_c
                 break;
             case PIXFORMAT_JPEG:
                 // Read the number of data items transferred
-                MAIN_FB()->bpp = (MAX_XFER_SIZE - __HAL_DMA_GET_COUNTER(&DMAHandle))*4;
+                MAIN_FB()->bpp = ((MAX_XFER_SIZE/4) - __HAL_DMA_GET_COUNTER(&DMAHandle))*4;
+                // Remove trailing zeros...
```
iabdalkader: Why is this needed ?

kwagyeman: Removes the trailing zeros in a jpeg image if there are any,

PR #560, opened 2019-08-11
Fix cache maintenance issue in JPEG mode.

iabdalkader:
* In JPEG mode the cacheable framebuffer memory is used directly by the DMA, instead of the line buffer.
Cache maintenance must be performed before the CPU accesses the framebuffer memory.


kwagyeman:
Tested and works. It looks like the end of the jpeg stream is always 0xFFD9 or 0xFFD9D9D9D9D9 etc. E.g. there's extra D9 padding. While this doesn't need to be cleaned up to read the jpeg data is is extra garbage that could be removed.

PR #562, opened 2019-08-18
Kwabena/ov2640 final

kwagyeman:
Two commits
* The first commit makes jpeg mode work even when the jpeg image is larger than the jpeg buffer. Frame rate takes a noise dive however. Also, added methods for sending jpeg data to the IDE over serial.
* The second commit fixes the resolution errors previous with the OV2640. Setting the resolution works right now.

All features of the OV2640 driver appear to work except for AEC control. Turning AEC off/on doesn't appear to do anything. Not sure how to fix. We can definitely change the exposure value but the camera doesn't really care what we write and stays in auto mode. Without new information this is a won't fix.

### DIFF
--- a/src/omv/framebuffer.c
+++ b/src/omv/framebuffer.c
@@ -6,15 +6,54 @@
  * Framebuffer stuff.
  *
  */
-#include "imlib.h"
-#include "omv_boardconfig.h"
+#include "mpprint.h"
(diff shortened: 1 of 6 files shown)


kwagyeman:
Switching to work on DRAM version now.

PR #578, opened 2019-09-03
Kwabena/sdram support

kwagyeman:
The jpeg compression timeout was too low initially for jpeg compressing UXGA images and etc. from the OV2640 (you can also just capture a JPG image from the OV2640 too).

High res apriltag support allows doing apriltag detection on high res images. Performance is poor because of low DRAM speed.

### DIFF
--- a/src/omv/boards/OPENMV2/imlib_config.h
+++ b/src/omv/boards/OPENMV2/imlib_config.h
@@ -111,6 +111,9 @@
 // Enable find_apriltags() (64 KB)
 //#define IMLIB_ENABLE_APRILTAGS
 
+// Enable high res find_apriltags() - uses more RAM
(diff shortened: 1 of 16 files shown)


kwagyeman:
Let me check why that happens. It builds on my machine.

iabdalkader:
> Let me check why that happens. It builds on my machine.

CI is having finding that commit I think, anyway there shouldn't be any edits to src/micropython.

kwagyeman:
Ah, okay, I rebased off of master and it needed MP updated. I think I can fix this.

kwagyeman:
Actually, if you could fix this will get done faster. I can't look at this until much later tonight. Please feel free to cherry pick the commits.

iabdalkader:
> Please feel free to cherry pick the commits.

I can't do that either, can't pull this locally.

PR #582, opened 2019-09-06
Fix printf issue in apriltags, dmtx and lsd.

iabdalkader:
* The root cause for this issue is "define double float". printf functions promote
float to double, which is #define'd as float causing the implicit conversion error.
* Replaced all double with float, and DBL_MIN and FLT_MIN etc... and added explicit
cast to double in printf functions.


iabdalkader:
@kwagyeman Can you test this ? I want to merge it.

PR #583, opened 2019-09-08
Fix binary ops bug.

### LINKED ISSUE #534: Binary ops bug when using BMP
### Steps to reproduce
1. Create a 320x240 black filled BMP and save as mask.bmp
2. Copy the image to SD card and run the following script:
```
import sensor, image
sensor.reset()
sensor.skip_frames(time=1000)
while(True):
   sensor.snapshot().b_and("mask.bmp")
```

### Expected behaviour
The framebuffer should be all black/clear. It works as expected with pgm.

### Actual behaviour
Garbage output black/white dots. This was reported on the forums too http://forums.openmv.io/viewtopic.php?f=6&t=1356

### Describe your setup

**IDE version:**
2.2.0
**Firmware version:**
3.4.1/2
**Operating system:**
Linux
### For developers

**GCC version:**
7.3.1 7-2018-q2-update
**Build target:**
OPENMV3/4

iabdalkader:
* fix #534

### DIFF
--- a/src/omv/img/imlib.c
+++ b/src/omv/img/imlib.c
@@ -594,20 +594,21 @@ void imlib_image_operation(image_t *img, const char *path, image_t *other, int s
         // window is equal to an image in size. However, since this is not the
         // case we shrink the window size to how many lines we're buffering.
         temp.pixels = alloc;
-        temp.h = (size / (temp.w * temp.bpp)); // round down
+        // Set the max buffer height to image height.
+        temp.h = IM_MIN(img->h, (size / (temp.w * temp.bpp)));
         // This should never happen unless someone forgot to free.
         if ((!temp.pixels) || (!temp.h)) {
             nlr_raise(mp_obj_new_exception_msg(&mp_type_MemoryError,
                                                "Not enough memory available!"));
         }
         for (int i=0; i<img->h; i+=temp.h) { // goes past end
-            int can_do = IM_MIN(temp.h, img->h-i);
-            imlib_read_pixels(&fp, &temp, 0, can_do, &rs);
-            for (int j=0; j<can_do; j++) {
+            int lines = IM_MIN(temp.h, img->h-i);
+            imlib_read_pixels(&fp, &temp, 0, lines, &rs);
+            for (int j=0; j<lines; j++) {
                 if (!vflipped) {
                     op(img, i+j, temp.pixels+(temp.w*temp.bpp*j), data, false);
                 } else {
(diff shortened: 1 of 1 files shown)


src/omv/img/imlib.c
```diff
@@ -594,20 +594,21 @@ void imlib_image_operation(image_t *img, const char *path, image_t *other, int s
...
+        // Set the max buffer height to image height.
+        temp.h = IM_MIN(img->h, (size / (temp.w * temp.bpp)));
         // This should never happen unless someone forgot to free.
         if ((!temp.pixels) || (!temp.h)) {
             nlr_raise(mp_obj_new_exception_msg(&mp_type_MemoryError,
                                                "Not enough memory available!"));
         }
         for (int i=0; i<img->h; i+=temp.h) { // goes past end
-            int can_do = IM_MIN(temp.h, img->h-i);
-            imlib_read_pixels(&fp, &temp, 0, can_do, &rs);
```
iabdalkader: @kwagyeman Wouldn't this start reading from zero every time ?

kwagyeman: Um, I don't have the context for this code stored in my head. But, it's just reading a line at a time. You may be breaking the code with this commit. Test it. I don't believe read pixels resets itself. Line start and line end are just how many lines to read. So, the 0 doesn't really mean anything.

kwagyeman:
I don't see how this fixes the bug. Um, so, the thing to note is that BMP images are store upside down sometimes. So, with PPM/PNM you always read the image in top to bottom order which makes writing code for it easy. BMP images can start bottom to top however or top to bottom. This makes working with them harder as you have to handle that and invert the operation direction.

iabdalkader:
@kwagyeman This wasn't the only issue with the code, the temp image height was set to the max lines that fit alloc, so I limit that to image height. The other change line_start is always 0, this is an issue only when the buffer is less than height (otherwise it copies the whole image in one go).

src/omv/img/imlib.c
iabdalkader: I see now, line_end is actually the number of lines to read not a line offset (it should be called line_count). This works because you're reading from files and the file pointer is where you last stopped. I think line_start is misleading and should be removed.

iabdalkader: Okay I reverted this change to 0, but line_start should be removed and line_end renamed to line_count

kwagyeman: Okay, so, basically, the error was that the temp.h buffer could get too big.

PR #620, opened 2019-10-24
License

thibthibaut:
Removed Cube.AI specific files and added instruction to link them from its installation path to avoid dependancy on versions and missmatches between the API and the runtime library.


kwagyeman:
Wait, this makes it.hard to compile the code. Please undo this PR. Our firmware should not require an install path. This is not approved.

iabdalkader:
@kwagyeman Those Cube.AI files are required only if CUBEAI=1 is defined, so it doesn't affect the main firmware build. I trust ST know what's best for their demo that's why I've merged it. That said, I agree it makes it harder to test the example (I liked it more when it was all self-contained). @kwagyeman suggested that you add a step to fetch the required files and libraries if they don't exist @thibthibaut.

PR #621, opened 2019-10-25
Add support for TensorFlow Lite for Microcontrollers

kwagyeman:
This has been a long-time coming, but, it's finally here. TensorFlow lite
runs on the OpenMV Cam now. Better yet, a person detection model
is now built-into all OpenMV Cams too!

Our default code does image classification and supports multi-scale
object detection using a sliding window.

In a coming PR I will add mobilenet examples for the H7 with SDRAM
and image segmentation support.


iabdalkader:
merged in #622

PR #635, opened 2019-11-01
Default CAN baudrate is 125 kbps

jeremitu:
CAN_DEFAULT_PRESCALER is 32. Tested with PCAN Dongle


iabdalkader:
That's right the default is 125Kbps, thanks.

PR #659, opened 2019-12-29
first round of optimizations

bitbank2:
The apriltag linked list search and image thresholding are significantly faster with this set of changes. It doesn't appear to have broken anything. Measuring the effect on the frame rate is harder. With an image full of apriltags, the minimum speed seems to have been accelerated from 7.7fps to 10.6fps.

### DIFF
--- a/src/omv/img/apriltag.c
+++ b/src/omv/img/apriltag.c
@@ -13,6 +13,9 @@
 #include <stdio.h>
 #include "imlib.h"
 
+// Enable new code optimizations
+#define OPTIMIZED
+
 #ifdef IMLIB_ENABLE_APRILTAGS
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunused-variable"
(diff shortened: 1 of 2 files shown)


iabdalkader:
Hi, thank you for the PR, I'll review it. I'm wondering why you changed winc.c ? It looks irrelevant.

bitbank2:
I changed winc.c because the original code gave "punned-type" pointer errors on my compiler.

bitbank2:
The fact that you don't see much of a speedup at the lower resolution sounds like other issues in the camera pipeline are taking more time. The 1.5x speedup (not 1.25x) at VGA resolution means that the changes are having a good effect on speed when most of the CPU is busy working on image data. At lower resolutions there are probably purposeful pauses (frame rate limiter) or other limiting factors in the imaging pipeline which hide the speedup of smaller frame buffers.

iabdalkader:
Hi I also benchmarked the changes on a 250x250 image with 9 apriltrags. It saves about 2ms on average (the test runs 1000 times to account for IRQs etc..). I'll merge this, but can you please first revert winc.c and send it in another PR ? Note we use gcc version 7.3.1 20180622 (7-2018-q2-update), it's probably a warning from a newer gcc but I need to look at it separately. Thank you.

bitbank2:
ok, I removed the winc.c changes. I'll create a new pull request for that

PR #663, opened 2020-01-03
Optimized JPEG MCU capture

bitbank2:
For grayscale and RGB565 JPEG encoding (used to get camera output), the original code  was not optimal. The grayscale code was sped up slightly, but the change isn't very visible because the communication and receiving end of the pipeline are more of a limiting factor.

The RGB565 code was sped up enough to see a change in the max frame rate. On my test rig, the QVGA RGB565 frame rate (as visible in the IDE running the hello world sample) went from 15.6 FPS to 20.4 FPS with this change. Additional speed can be achieved with some SIMD optimization (I'll do on my next pass).

### DIFF
--- a/src/omv/img/jpeg.c
+++ b/src/omv/img/jpeg.c
@@ -73,17 +73,14 @@ static uint8_t *get_mcu()
         case 1: {
             int dx=MCU_W, dy=MCU_H;
             uint32_t *s32, *d32;
-            int bSmall = 0; // truncated MCU
             if (jpeg_enc.x_offset+dx > jpeg_enc.img_w)
                 dx = jpeg_enc.img_w - jpeg_enc.x_offset;
             if (jpeg_enc.y_offset+dy > jpeg_enc.img_h)
                 dy = jpeg_enc.img_h - jpeg_enc.y_offset;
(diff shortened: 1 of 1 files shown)


src/omv/img/jpeg.c
```diff
@@ -70,28 +70,53 @@ static uint8_t *get_mcu()
...
-                    } else {
+        case 1: {
+            int dx=MCU_W, dy=MCU_H;
+            uint32_t *s32, *d32;
+            int bSmall = 0; // truncated MCU
+            if (jpeg_enc.x_offset+dx > jpeg_enc.img_w)
+                dx = jpeg_enc.img_w - jpeg_enc.x_offset;
+            if (jpeg_enc.y_offset+dy > jpeg_enc.img_h)
+                dy = jpeg_enc.img_h - jpeg_enc.y_offset;
+            if (dx != MCU_W || dy != MCU_H) { // partial MCU, fill with 0's to start
```
kwagyeman: Can you remove the bsmall variable and just use the if statement for two loops here. Would be easier to read.

bitbank2: Fixed

PR #667, opened 2020-01-08
Red blue switcheroo

thouseweart:
Tested working with H7 and LCD shield

### DIFF
--- a/scripts/examples/02-Board-Control/spi_control.py
+++ b/scripts/examples/02-Board-Control/spi_control.py
@@ -72,4 +72,3 @@ while(True):
 
     print(clock.fps()) # Note: Your OpenMV Cam runs about half as fast while
     # connected to your computer. The FPS should increase once disconnected.
-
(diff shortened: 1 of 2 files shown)


thouseweart:
According to the [ST7735R datasheet](https://cdn.shopify.com/s/files/1/0803/9211/files/ST7735R.pdf?8137406006175742039), it seems like RGB vs. BGR mode is also controlled by the voltage present at pin "SRGB". So perhaps a manufacturing defect? If it helps, I ordered my shield [(RB-Omv-09) from RobotShop ](https://www.robotshop.com/en/openmv-cam-lcd-shield.html) on ~7/10/19.

kwagyeman:
Yeah, okay, I'll just update the demo scripts and add an argument that allows you to specify this.

iabdalkader:
Hi, I've added an argument to lcd init to switch to BGR mode see https://github.com/openmv/openmv/pull/719
```
lcd.init(bgr=True)
```
Sets MADCTL to 0xC8

PR #668, opened 2020-01-11
histeq and JPEG binary speedup

bitbank2:
This set of changes speeds up the histeq (equalization by histogram) of RGB565 images and the JPEG compression of binary (1-bpp) images.


bitbank2:
I fixed an issue with HQQQVGA JPEG compression and added more detailed comments to the rest of the new JPEG code.

bitbank2:
ok, further cleanup of the comments, variables and braces

PR #700, opened 2020-02-10
Add OV7690 and LSM6DS3 driver support

kwagyeman:
Sensor transposing was added to support auto rotation using the IMU to
detect gravity.

* IMU code paths are all done, tested, and working correctly.
* Sensor auto rotation code paths are all done, tested, and working correctly.
* The OV7690 driver needs more work still. Default register settings are not optimal and some sensor control methods do not have the correct affect in all cases.


kwagyeman:
Please merge PR: https://github.com/openmv/micropython/pull/64 to fix the build.

PR #716, opened 2020-02-13
Auto rotation enabled

kwagyeman:


src/omv/main.c
```diff
@@ -495,6 +496,9 @@ int main(void)
     // mounting the file-system to log errors (if any).
     if (first_soft_reset) {
         sensor_init_ret = sensor_init();
+        #if MICROPY_PY_IMU
+        if ((!sensor_init_ret) && (sensor_get_id() == OV7690_ID)) py_imu_init();
```
iabdalkader: Why check for OV7690 ? It needs to be initialized if it's enabled, agree ?

src/omv/sensor.c
```diff
@@ -433,7 +433,11 @@ int sensor_reset()
     sensor.hmirror       = false;
     sensor.vflip         = false;
     sensor.transpose     = false;
+    #if MICROPY_PY_IMU
+    sensor.auto_rotation = sensor.chip_id == OV7690_ID;
```
iabdalkader: Maybe it's better to move this to ov7690_init ?

kwagyeman:
You need to create a new target for the new camera then.

src/omv/sensor.c
kwagyeman: It can't be in init because reset would clear it...

kwagyeman: Oh, I see... like the color_palette.

kwagyeman: No, I think it makes sense in reset...

iabdalkader: I meant here:
https://github.com/openmv/openmv/blob/master/src/omv/ov7690.c#L616

kwagyeman: It doesn't really make sense to set it there. Instead can you merge the PR and then create a micro target and then we it being set correctly into the micro target.

PR #725, opened 2020-02-18
Move clock config to board files.

iabdalkader:


arunachalam9513:
i micopython code for LED control using open mv pins

PR #747, opened 2020-04-01
optimized sensor copy to allow ful res capture

bitbank2:
The OV5640 sensor produces up to a 2560x1944 image. At the highest res, the dma line copy code wasn't keeping up with the output of the sensor. I optimized this by improving the memory throughput. The Cortex-M7 has a 32-bit memory bus, so the fastest movement of memory is by reading and writing 32-bits at a time on aligned addresses. The memcpy() function already has code which detects if the source and destination addresses are 32-bit aligned, so most of the work was passed to memcpy().

### DIFF
--- a/src/omv/ov5640.c
+++ b/src/omv/ov5640.c
@@ -426,7 +426,7 @@ static int set_pixformat(sensor_t *sensor, pixformat_t pixformat)
             break;
         case PIXFORMAT_GRAYSCALE:
         case PIXFORMAT_YUV422:
-            ret |= cambus_writeb2(sensor->slv_addr, FORMAT_CONTROL, 0x30);
+            ret |= cambus_writeb2(sensor->slv_addr, FORMAT_CONTROL, 0x10);
             ret |= cambus_writeb2(sensor->slv_addr, FORMAT_CONTROL_MUX, 0x00);
             pll = (resolution[sensor->framesize][0] > 2048) ? 0x50 : 0x64; // 32 MHz vs 40 MHz
             break;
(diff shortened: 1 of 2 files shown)


src/omv/ov5640.c
```diff
@@ -612,7 +612,8 @@ static int set_framesize(sensor_t *sensor, framesize_t framesize)
             break;
         case PIXFORMAT_GRAYSCALE:
         case PIXFORMAT_YUV422:
-            pll = (w > 2048) ? 0x50 : 0x64; // 32 MHz vs 40 MHz
```
kwagyeman, line 615: Please verify that RGB565/BAYER/GRAYSCALE all work for 0x64 and then just merge those into one case and remove the dead code. Note that there's another place in the code int eh set_pixformat method that you must change too. Ignore JPEG for now. That needs to be fixed in another way.

src/omv/sensor.c
```diff
@@ -904,33 +904,43 @@ void DCMI_DMAConvCpltUser(uint32_t addr)
...
                     dst += (line - MAIN_FB()->y) * MAIN_FB()->w;
                     src += MAIN_FB()->x;
-                    for (int i = MAIN_FB()->w; i; i--) {
-                        *dst++ = *src++;
-                    }
+                    memcpy(dst, src, MAIN_FB()->w);
                     break;
                 case PIXFORMAT_GRAYSCALE:
                     dst += (line - MAIN_FB()->y) * MAIN_FB()->w;
+                    src += MAIN_FB()->x;
```
kwagyeman, line 911: Move under the if?

bitbank2: This works correctly either as-is or after the if; I can change it for aesthetics.

src/omv/ov5640.c
bitbank2: Will do...

bitbank2: Verified; I just pushed the change which sets the speed to 0x64 for everything. All pixel formats are stable at the highest resolution.

src/omv/sensor.c
iabdalkader: @bitbank2 yes please do so, it took me a while to figure out that it doesn't matter.

bitbank2: done

PR #752, opened 2020-04-13
Fix alpha on imlib_draw, implement palettes

matsondawson:
1/ image.draw_image fails to apply alpha correctly on anything other than grayscale images, this fixes that.
2/ it also adds a new parameter to image.draw_image, color_palette.  The palette is used when the other image is a grayscale image such as thermographic when you want an e.g. rainbow palette to be applied.  This saves allocating extra memory for a separate buffer for the color thermograph then using a function such as image.blend to blend to the camera.  If you pass a color_palette and the other image is not grayscale an exception is raised.

### DIFF
--- a/src/omv/img/draw.c
+++ b/src/omv/img/draw.c
@@ -417,11 +417,11 @@ void imlib_draw_string(image_t *img, int x_off, int y_off, const char *str, int 
     }
 }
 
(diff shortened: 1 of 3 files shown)


matsondawson:
Please cancel, I forgot to branch before the changes.

PR #755, opened 2020-04-15
sped up jpeg encoder by adding a direct bayer-to-ycbcr path

bitbank2:
This change speeds up color JPEG encoding (as measured in the IDE) between 10 and 15% over the previous version. VGA color max frame rate is now around 15 FPS in good lighting with a simple scene (30+ FPS without the IDE).

### DIFF
--- a/src/omv/img/imlib.c
+++ b/src/omv/img/imlib.c
@@ -566,6 +566,293 @@ void imlib_bayer_to_rgb565(image_t *img, int w, int h, int xoffs, int yoffs, uin
     } // faster way
     } // for y
 }
+
+//
+// Convert a row of Bayer pixels into grayscale output
+//
+void bayer_to_y(image_t *img, int x_offset, int y_offset, int width, uint8_t *Y)
+{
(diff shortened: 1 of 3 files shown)


iabdalkader:
> Please move bayer_to_y and bayer_to_binary to the main library as they are re-usable for other algorithms.

I don't see anything calling those functions ? If they are used somewhere, @bitbank2 can you please add the imlib_ prefix ? and function prototypes to the imlib.h ? It's very important to keep the prefix. Otherwise if they are not used please remove them.

kwagyeman:
These methods are for adding bayer support to many of our find algorithms. In particular, the goal is to enable 640x480 processing on color images on the H7.

iabdalkader:
> These methods are for adding bayer support to many of our find algorithms. In particular, the goal is to enable 640x480 processing on color images on the H7.

In this case, we'll need the prototypes in imlib.h and the prefix please.

bitbank2:
ok, will add...

PR #760, opened 2020-04-20
optimized erode/dilate

bitbank2:


bitbank2:
pushed a minor cleanup

PR #765, opened 2020-04-24
Fix send/sendto

kwagyeman:
Both methods can move megabytes of data now without issues.

Also, fixed an uninitialized value returning from recv_from when you poll
for data and there's nothing to get.

### DIFF
--- a/src/winc1500/src/winc.c
+++ b/src/winc1500/src/winc.c
@@ -184,6 +184,8 @@ static void socket_callback(SOCKET sock, uint8_t msg_type, void *msg)
                         pstrRecv->s16BufferSize, rfrom->addr.sin_addr.s_addr, rfrom->addr.sin_port);
 			} else {
                 rfrom->size = pstrRecv->s16BufferSize;
+                rfrom->addr.sin_port = 0;
+                rfrom->addr.sin_addr.s_addr = 0;
 				debug_printf("recvfrom error:%d\n", pstrRecv->s16BufferSize);
 			}
             async_request_done = true;
@@ -820,23 +822,34 @@ int winc_socket_connect(int fd, sockaddr *addr, uint32_t timeout)
 
 int winc_socket_send(int fd, const uint8_t *buf, uint32_t len, uint32_t timeout)
 {
+    uint32_t tick_start = HAL_GetTick();
+
     int bytes = 0;
 
     while (bytes < len) {
         // Split the packet into smaller ones.
         int n = MIN((len - bytes), SOCKET_BUFFER_MAX_LENGTH); 
-        int ret = WINC1500_EXPORT(send)(fd, (uint8_t*)buf + bytes, n, timeout);
+        int ret = WINC1500_EXPORT(send)(fd, (uint8_t*)buf + bytes, n, 0);
 
         if (ret == SOCK_ERR_NO_ERROR) {
             // Do async request
-            ret = winc_async_request(SOCKET_MSG_SEND, &n, timeout);
+            int async_ret;
+            ret = winc_async_request(SOCKET_MSG_SEND, &async_ret, timeout);
 
             // Check sent bytes returned from async request.
-            if (ret != SOCK_ERR_NO_ERROR || n <= 0) {
-                return (n <= 0)? n : ret;
+            if (ret != SOCK_ERR_NO_ERROR || async_ret <= 0) {
+                return (ret != SOCK_ERR_NO_ERROR) ? ret : async_ret;
             }
+
+            bytes += async_ret;
+        } else if (ret == SOCK_ERR_BUFFER_FULL) {
+            // timeout == 0 in blocking mode.
+            if (timeout && ((HAL_GetTick() - tick_start) >= timeout)) {
+                break;
+            }
+        } else { // another error
+            return ret;
         }
-        bytes += n;
     }
 
     return bytes;
@@ -856,8 +869,9 @@ int winc_socket_recv(int fd, uint8_t *buf, uint32_t len, winc_socket_buf_t *sock
             ret = winc_async_request(SOCKET_MSG_RECV, &recv_bytes, timeout);
         }
 
+        // Check received bytes returned from async request.
         if (ret != SOCK_ERR_NO_ERROR || recv_bytes <= 0) {
-            return (recv_bytes <= 0)? recv_bytes : ret;
+            return (ret != SOCK_ERR_NO_ERROR) ? ret : recv_bytes;
         }
 
         sockbuf->size = recv_bytes;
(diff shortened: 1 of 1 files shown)


src/winc1500/src/winc.c
```diff
@@ -820,23 +822,33 @@ int winc_socket_connect(int fd, sockaddr *addr, uint32_t timeout)
...
             // Do async request
-            ret = winc_async_request(SOCKET_MSG_SEND, &n, timeout);
+            int async_ret;
+            ret = winc_async_request(SOCKET_MSG_SEND, &async_ret, timeout);
 
             // Check sent bytes returned from async request.
-            if (ret != SOCK_ERR_NO_ERROR || n <= 0) {
-                return (n <= 0)? n : ret;
-            }
+            if (ret != SOCK_ERR_NO_ERROR) return ret;
```
iabdalkader, line 840: This is basically the same as the old code. There's no reason to change it (except for n<0), and please reuse n.

src/winc1500/src/winc.c
```diff
@@ -820,23 +822,33 @@ int winc_socket_connect(int fd, sockaddr *addr, uint32_t timeout)
...
+            if (async_ret < 0) return async_ret;
+
+            bytes += async_ret;
+        } else if (ret != SOCK_ERR_BUFFER_FULL) {
```
iabdalkader, line 844: Just a small style change:
```
if (ret == SOCK_ERR_NO_ERROR) {
   /// do the async call
} else if (ret == SOCK_ERR_BUFFER_FULL) {
 // wait until timeout
} else { // another error
  return ret;
}
```

src/winc1500/src/winc.c
```diff
@@ -856,9 +868,8 @@ int winc_socket_recv(int fd, uint8_t *buf, uint32_t len, winc_socket_buf_t *sock
             ret = winc_async_request(SOCKET_MSG_RECV, &recv_bytes, timeout);
         }
```
iabdalkader, line 870: This is exactly the same as the removed code, so why is this change needed ? Unless you want to check if winc_recv failed first (it's not likely to happen), but you could just check ret first:
```
if (ret != SOCK_ERR_NO_ERROR || recv_bytes <= 0) {
     return (ret != SOCK_ERR_NO_ERROR)? ret: recv_bytes;
}
```

src/winc1500/src/winc.c
```diff
@@ -872,27 +883,51 @@ int winc_socket_recv(int fd, uint8_t *buf, uint32_t len, winc_socket_buf_t *sock
 
 int winc_socket_sendto(int fd, const uint8_t *buf, uint32_t len, sockaddr *addr, uint32_t timeout)
```
iabdalkader, line 884: Same comments as send.

src/winc1500/src/winc.c
```diff
@@ -872,27 +883,51 @@ int winc_socket_recv(int fd, uint8_t *buf, uint32_t len, winc_socket_buf_t *sock
...
 int winc_socket_recvfrom(int fd, uint8_t *buf, uint32_t len, sockaddr *addr, uint32_t timeout)
 {
+    memset(addr, 0, sizeof(sockaddr));
+
     recv_from_t rfrom;
     int ret = WINC1500_EXPORT(recvfrom)(fd, buf, len, timeout);
     if (ret == SOCK_ERR_NO_ERROR) {
         // Do async request
         ret = winc_async_request(SOCKET_MSG_RECVFROM, &rfrom, timeout);
     }
```
iabdalkader, line 928: Please keep the old code and switch ret and rfrom.size.

src/winc1500/src/winc.c
kwagyeman: I can't do this for all cases because sendto returns 0 for n always. For consistency I don't want to do this.

src/winc1500/src/winc.c
```diff
@@ -820,23 +822,36 @@ int winc_socket_connect(int fd, sockaddr *addr, uint32_t timeout)
...
             // Check sent bytes returned from async request.
-            if (ret != SOCK_ERR_NO_ERROR || n <= 0) {
-                return (n <= 0)? n : ret;
+            if (ret != SOCK_ERR_NO_ERROR || async_ret <= 0) {
+                return (ret != SOCK_ERR_NO_ERROR) ? ret : async_ret;
             }
+
+            bytes += async_ret;
+        } else if (ret == SOCK_ERR_BUFFER_FULL) {
+            // wait until timeout
```
iabdalkader, line 846: The timeout should be here.

src/winc1500/src/winc.c
```diff
@@ -820,23 +822,36 @@ int winc_socket_connect(int fd, sockaddr *addr, uint32_t timeout)
...
+        } else { // another error
+            return ret;
+        }
+
+        // timeout == 0 in blocking mode.
+        if (timeout && ((HAL_GetTick() - tick_start) >= timeout)) {
```
iabdalkader, line 852: This runs in every iteration of the loop.

src/winc1500/src/winc.c
```diff
@@ -872,26 +888,55 @@ int winc_socket_recv(int fd, uint8_t *buf, uint32_t len, winc_socket_buf_t *sock
...
+            // Do async request
+            int async_ret; // sendto always returns 0.
+            ret = winc_async_request(SOCKET_MSG_SENDTO, &async_ret, timeout);
+
+            // Check sent bytes returned from async request.
+            if (ret != SOCK_ERR_NO_ERROR || async_ret < 0) {
+                return (ret != SOCK_ERR_NO_ERROR) ? ret : async_ret;
+            }
+
+            bytes += n;
```
iabdalkader, line 910: bytes += async_ret.

src/winc1500/src/winc.c
```diff
@@ -872,26 +888,55 @@ int winc_socket_recv(int fd, uint8_t *buf, uint32_t len, winc_socket_buf_t *sock
...
+        } else if (ret == SOCK_ERR_BUFFER_FULL) {
+            // wait until timeout
```
iabdalkader, line 912: timeout should be here.

src/winc1500/src/winc.c
```diff
@@ -872,26 +888,55 @@ int winc_socket_recv(int fd, uint8_t *buf, uint32_t len, winc_socket_buf_t *sock
...
+        } else { // another error
+            return ret;
+        }
+
+        // timeout == 0 in blocking mode.
+        if (timeout && ((HAL_GetTick() - tick_start) >= timeout)) {
```
iabdalkader, line 918: This runs in every iteration of the loop.

PR #768, opened 2020-04-24
Optimized datamatrix ops

bitbank2:


kwagyeman:
Hi, what's going on with this code. What are these changes doing? What's the performance gain if any?

bitbank2:
Reducing the de-referencing of member variables in the inner loops increased the speed. This was a measurable, but not major improvement to the perf. A change of algorithm would be needed to see big improvements in the performance.

PR #770, opened 2020-04-28
Improve WiFi performance by 20X

kwagyeman:
Polling for the async response was ruinning WiFi and UDP performance.
This fix removes that polling while not breaking anything. It has been
stress tested while streaming over an hour of wifi video data using the
new RPC scripts.

### DIFF
--- a/src/winc1500/src/winc.c
+++ b/src/winc1500/src/winc.c
@@ -410,6 +410,7 @@ static int winc_async_request(uint8_t msg_type, void *ret, uint32_t timeout)
 
     // Wait for async request to finish.
     while (async_request_done == false) {
+        __WFI();
         // Handle pending events from network controller.
         m2m_wifi_handle_events(NULL);
         // timeout == 0 in blocking mode.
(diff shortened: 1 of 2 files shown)


kwagyeman:
Note that m2m_wifi_handle_events will spin about 5 times internally and then return. Performance could probably be further improved by moving m2m_wifi_handle_events() into an interrupt service routine and called only when needed @iabdalkader. We should probably do this to enable faster streaming on the H7 Plus. Right now we get about 5 Mbps. We should be able to unlock 2x more speed.

src/winc1500/src/winc.c
```diff
@@ -827,21 +827,19 @@ int winc_socket_send(int fd, const uint8_t *buf, uint32_t len, uint32_t timeout)
     int bytes = 0;
 
     while (bytes < len) {
+        // Do async request (clean out any messages - but ignore them).
+        int async_ret;
+        async_request_data = &async_ret;
+        async_request_done = false;
+        async_request_type = SOCKET_MSG_SENDTO;
```
iabdalkader, line 834: This should be send not sendto.

kwagyeman: Fixed

PR #771, opened 2020-04-28
Add disable_fb.

kwagyeman:
Allow user control to disable the frame buffer. Necessary for high manual
jpeg image streaming.


iabdalkader:
1) Do you need to disable all the streaming or just the encode_for_ide part ?
2) I don't see where disable_fb gets reset, if you do a soft-reset it will remain disabled.
3) Can you use fb_ prefix ? It's like new module now.

kwagyeman:
1. Probably should disable all streaming just to make it consistent - only need to disable the encode for IDE part however.
2. Good catch.
3. Prefix on what?

iabdalkader:
> 3. Prefix on what?

To the new functions you just added to framebuffer.c... ex fb_disable_streaming since they're all in framebuffer.c and we may add more later.

kwagyeman:
Updated

PR #773, opened 2020-04-29
Modify compression methods to allow jpeg images

kwagyeman:
No longer will calling compress() on a jpeg image blow up on you. Now handled intelligently.

(Need this for for the jpeg streamer I just made. So you can pass sensor.PIXFORMAT_JPEG and it works fine).


iabdalkader:
@kwagyeman Please fix the conflicts, I renamed some functions in #775

PR #774, opened 2020-04-30
Close sockets that have errors.

kwagyeman:
Tested and works.

Yes, you can execute close during the callback. It works fine.

### DIFF
--- a/src/winc1500/src/winc.c
+++ b/src/winc1500/src/winc.c
@@ -82,6 +82,14 @@ static void resolve_callback(uint8_t *host, uint32_t ip)
  */
 static void socket_callback(SOCKET sock, uint8_t msg_type, void *msg)
 {
+    // Close Socket on SEND/SENDTO Error.
+    // The next SEND/SENDTO operation on the socket will return SOCK_ERR_INVALID_ARG and fail.
+
+    if (((msg_type == SOCKET_MSG_SEND) || (msg_type == SOCKET_MSG_SENDTO))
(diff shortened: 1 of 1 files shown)


src/winc1500/src/winc.c
```diff
@@ -82,6 +82,14 @@ static void resolve_callback(uint8_t *host, uint32_t ip)
  */
 static void socket_callback(SOCKET sock, uint8_t msg_type, void *msg)
 {
+    // Close Socket on SEND/SENDTO Error.
```
iabdalkader, line 85: Why isn't this in the switch below in line 160 ?

iabdalkader: Also why is it needed ? The socket will be closed when send()/sendto() returns an error.

kwagyeman: I want it to run whenever anything happens so that the socket gets closed. This is why I put it there. If you put it in the switch then it would not be run for any socket callback.

I had to add this because the winc module accepts data for any invalid sockets without errors. I.e. when you send. It will accept the data always even if the socket was closed from the other side. The winc module only communicates the socket is bad via the callback.

kwagyeman: Note, it still blocks you sometimes until there's buffer space... but, it will never just block you indefinitely like you'd expect on a broken socket. The module accepts the data, throws it away, and then returns that it couldn't send the data via the callback.

PR #779, opened 2020-05-05
Remove udev

kwagyeman:
OpenMV IDE installs udev for folks currently. This isn't needed anymore.


iabdalkader:
Might be needed if you don't use the IDE or don't want to install/use it ? @kwagyeman

kwagyeman:
The IDE installs a bunch of other stuff. If they don't use the IDE they still should run the installer as all related setup is done there.

Just trying to cleanup things.

PR #786, opened 2020-05-08
Make cambus functions accept I2C handle.

iabdalkader:
Tested with the following cams/sensors:
* OPENMV3/OV7725
* OPENMV4 OV2640, OV5640, OV7725, MT9V034 and all FLIR sensors.
* OPENMV4P/OV5640


iabdalkader:
> Looks good expect for how timing is exported. You should add an argument to init to set the timing to some freq and pull the right timing based on that.

Okay will do that in another PR. Note the freq can be (100KHz, 400KHz or 1000KHz) standard, fast and fast plus mode.

PR #787, opened 2020-05-09
New readout control code

kwagyeman:
Makes the OV5640 go at 120 FPS.


kwagyeman:
https://youtu.be/sGTJFywzuow

PR #789, opened 2020-05-09
remove unnecessary computation in template matching

georgiosfloros:
According to the paper of J. P. Lewis "**Fast normalized cross-correlation**", the computation of the numerator can be simplified, avoiding the computation of the `f_mean` and its subtraction from `f`.

### DIFF
--- a/src/omv/img/template.c
+++ b/src/omv/img/template.c
@@ -246,12 +246,11 @@ float imlib_template_match_ex(image_t *f, image_t *t, rectangle_t *roi, int step
         // The mean of the current patch
         uint32_t f_sum = imlib_integral_lookup(&sum, u, v, t->w, t->h);
         uint32_t f_sumsq = imlib_integral_lookup(&sumsq, u, v, t->w, t->h);
-        uint32_t f_mean = f_sum / (float) (t->w*t->h);
 
         // Normalized sum of squares of the image
         for (int y=v; y<(v+t->h); y++) {
             for (int x=u; x<(u+t->w); x++) {
-                int a = (int)f->data[y*f->w+x]-f_mean;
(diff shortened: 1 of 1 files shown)


kwagyeman:
Have you tested this and it doesn't change anything?

georgiosfloros:
@kwagyeman, I tested that in a small example that I created on my own. It shouldn't change anything, since it just removes a sum computation that would accumulate to zero.

What would be the appropriate way to test it? Through [this](https://github.com/openmv/openmv/blob/4f209ea9b1937499e8d3e54c89c04f5c2107f275/scripts/examples/09-Feature-Detection/template_matching.py) example? Or should I write an additional unit test similar to [this](https://github.com/openmv/openmv/blob/4f209ea9b1937499e8d3e54c89c04f5c2107f275/scripts/unittest/script/18-find_template.py) one but for the `SEARCH_EX` case?

Thanks in advance!

iabdalkader:
Hi, I don't have time now to go through the paper again but I ran this test with SEARCH_EX (on the camera with SDRAM):

```def unittest(data_path, temp_path):
    import image
    from image import SEARCH_EX, SEARCH_DS
    img = image.Image("unittest/data/graffiti.pgm", copy_to_fb=True)
    temp = image.Image("unittest/data/template.pgm", copy_to_fb=False)
    r = img.find_template(temp, 0.70, search=SEARCH_EX)
    print(r)


print(unittest("/unittest/data", "unittest/temp"))
````

After removing f_mean it doesn't find the template at all. Here's the location before in red (which is the correct template location) and after removing the f_mean in blue:

image

georgiosfloros:
Thanks for testing that @iabdalkader!
Since it doesn't work for your example, I withdraw the pull request.

PR #791, opened 2020-05-10
Add correction to imlib lens corr + optimise

matsondawson:
Most cheap lenses are not aligned to their sensors, this creates issues when overlaying two images as they don't quite align correctly.  We can fix this using a rotation_corr with a translate, or creating a new image and drawing it with offset, but this is slow.
I've added x_corr and y_corr to imlib_lens_corr.
The values represent a percentage/100 (0->1) x/y offset to apply before performing the lens correction.
As of example my H7's camera offset is over 5%.

### DIFF
--- a/src/omv/img/imlib.c
+++ b/src/omv/img/imlib.c
@@ -1244,40 +1244,41 @@ void imlib_lens_corr(image_t *img, float strength, float zoom, float x_corr, flo
                 for (int x = 0; x < halfWidth; x++) {
                     int newX = x - halfWidth;
                     int newX2 = newX * newX;
-
                     float precalculated = precalculated_table[(int)fast_sqrtf(newX2 + newY2)];
-                    int sourceX = fast_roundf(precalculated * newX); // rounding is necessary
                     int sourceY = fast_roundf(precalculated * newY); // rounding is necessary
-
-                    // plot the 4 symmetrical pixels
-                    uint32_t *ptr, pixel;
+                    int sourceX = fast_roundf(precalculated * newX); // rounding is necessary
                     int sourceY_down = down_adj + sourceY;
                     int sourceY_up = up_adj - sourceY;
                     int sourceX_right = right_adj + sourceX;
                     int sourceX_left = left_adj - sourceX;
 
+                    // plot the 4 symmetrical pixels
+                    // top 2 pixels
                     if (sourceY_down >= 0 && sourceY_down < h) {
-                        ptr = tmp + (((w + UINT32_T_MASK) >> UINT32_T_SHIFT) * sourceY_down);
+                        uint32_t *ptr = tmp + (((w + UINT32_T_MASK) >> UINT32_T_SHIFT) * sourceY_down);
+
                         if (sourceX_right >= 0 && sourceX_right < w) {
-                            pixel = IMAGE_GET_BINARY_PIXEL_FAST(ptr, sourceX_right);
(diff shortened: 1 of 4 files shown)


kwagyeman:
Hi, I paid Layer already to optimize this code. What's the performance change? If it's reduced then that's a problem.

kwagyeman:
I don't think you need to do all the changes you made except to change the lens center calculation.

matsondawson:
It's 6-7% slower, but there's big gain for me not double handling the image.
I could make it the about same speed but there'd be an increase in code complexity if you want me to look into it?  Essentially drawing the ranges that won't overflow first, then separately draw the ranges that might with checks for overflow.

> Re: I don't think you need to do all the changes you made except to change the lens center calculation.
Yeh, I'm not a smart man, I've changed that.  I also don't need to do the global "if ((0 <= sourceX) && (0 <= sourceY)) {"

There's some performance gain to be had from changing this to just swap the image pointer somehow when it's finished to point to data, but I realise there's no simple way to know what memory range the original image was allocated in.  e.g. frame_buffer, vs general memory
``` imlib.c:1218 
    void *data = fb_alloc(size, FB_ALLOC_NO_HINT);
    memcpy(data, img->data, size);
    memset(img->data, 0, size);
```

matsondawson:
I've added a look up table to this now and the performance is 50% to 80% faster depending on the pix format.

matsondawson:
My gut feeling is you're going to get little out of packing the destination word.
e.g. in grayscale you still have to do all the same calculations for every source pixel, and pack the destination pixel.  The source pixels don't come out in packed order so there's nothing you can do for that.  The store is 1/5th of memory accesses, and even then your only looking at performance gain if pipeline is stalled on a memory access.  Also the more code you put in the loop the more the registers get thrashed.  I'm happy to be proven wrong though.

matsondawson:
Interestingly on the H7 plus the memcpy in this causes a 20% performance hit, but only for RGB565 in VGA mode.  I wonder if it causes the DRAM bandwidth to be saturated somehow?  RGB565 is already half the speed of grayscale, so I'm assuming in VGA it has bandwidth issues.

kwagyeman:
It's because the image goes off chip then. Fballoc silently moves itself into onchip SRAM if it fits versuses using dram. Basically, there are two fballoc stacks. One in dram and one in SRAM. Whenever an alloc can fit in SRAM I do the same alloc I'm SRAM along with dram and then return the SRAM pointer instead of the dram pointer. When you pop I pop both regions.

kwagyeman:
Everything would be much slower on the plus if I didn't do this. The SDRAM bandwidth is just not amazing.

matsondawson:
@320x240 the table uses sqrt(160**2+120**2)*4(float)=800bytes
@640x480 the table uses sqrt(160**2+120**2)*4(float)=1600bytes

If that's an issue we can write a table and non table version with fall back if memory allocation fails.
But if it can allocate the secondary buffer anyway which @320x240 grayscale is 75k, the extra table is unlikely to be an issue.

kwagyeman:
That byte usage amount is not an issue.

matsondawson:
Also you don't have to accept my changes, I'm cool with that.  I think I'm obligated under the license to submit them back to you though?

kwagyeman:
You're not obliged. However, buy doing so you make OpenMV a better project and help it become the OpenCV library for embedded systems which was out goal. So, you're helping making that future a reality. Will check your code today.

PR #801, opened 2020-05-18
Fixed OV5640 Imaging Modes

kwagyeman:
* Grayscale/RGB565/Bayer/JPEG works perfectly on all reses.
* Increased maximum JPEG image transfer size.
* Improved line buffer transfer rate.

### DIFF
--- a/src/omv/sensor.c
+++ b/src/omv/sensor.c
@@ -59,7 +59,7 @@ const int resolution[][2] = {
     {64,   32  },    /* 64x32     */
     {64,   64  },    /* 64x64     */
     {128,  64  },    /* 128x64    */
-    {128,  128 },    /* 128x64    */
+    {128,  128 },    /* 128x128   */
     // Other
     {128,  160 },    /* LCD       */
     {128,  160 },    /* QQVGA2    */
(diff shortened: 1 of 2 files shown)


src/omv/sensor.c
```diff
@@ -525,6 +525,8 @@ int sensor_set_pixformat(pixformat_t pixformat)
         return -1;
     }
 
+    systick_sleep(100); // wait for the camera to settle
```
iabdalkader, line 528: This is usually followed by set_framesize (or the other way around), you'll be adding 200ms delay, and then skip frames is called anyway (or at least 100ms will pass by the time snapshot is called). Is it necessary ?

src/omv/sensor.c
```diff
@@ -888,10 +892,57 @@ static void sensor_check_buffsize()
     }
 }
 
+// ARM Cortex-M4/M7 Processors can access memory using unaligned 32-bit reads/writes.
```
iabdalkader, line 895: It does, but it's actually slower.

src/omv/sensor.c
kwagyeman: It just hides a corrupt frame that appears when you change modes. I didn't want to put in within the driver as that can be changed via set frame size.

src/omv/sensor.c
kwagyeman: Yes, I found an algorithm that does 4 byte aligned writes and does barrel shifting to align things also. Maybe I'll have Larry implement that faster. This was less code and I didn't want to bring in that giant algorithm.

PR #803, opened 2020-05-18
Fixed right edge bayer conversion error

bitbank2:
The bayer_to_ycbcr method was reading 1 pixel past the right edge when converting the rightmost pixels since the bayer to RGB algorithm averages the values of neighboring pixels. The new code re-uses the current pixel values to not read past the right edge.


iabdalkader:
Does this need to be done anywhere else ? Can you check all the conversion functions that may have this issue, and if so please commit to this PR.

bitbank2:
I'll double check the rest and let you know.

bitbank2:
All code appears to manage the edges properly. The bayer_to_y and bayer_to_rgb565 are not currently used, but appear to be correct.

Issue #802, opened 2020-05-18
Adding custom AprilTag


KBacsa:
I would like to add the detection of custom april tags as mentioned in #356 , mainly Tag36H8. 
### Steps to reproduce

1. Generated tags and codes using [apriltag-generation](https://github.com/AprilRobotics/apriltag-generation)

2. Create a new april tag family in apriltag.c such as :
```
////////////////////////////////////////////////////////////////////////////////////////////////////
//////// "tag36h8"
////////////////////////////////////////////////////////////////////////////////////////////////////

const apriltag_family_t tag36h8 = {
    .ncodes = 36042,
    .black_border = 1,
    .d = 6,
    .h = 8,
    .codes = {
       ...
      }
}
```

3. Add codes from Tag36h8.java file.

4. In imlib.h I added the mask constant such as :

```
typedef enum apriltag_families {
    TAG16H5 = 1,
    TAG25H7 = 2,
    TAG25H9 = 4,
    TAG36H10 = 8,
    TAG36H11 = 16,
    ARTOOLKIT = 64,
    TAG36H8 = 32
} __attribute__((aligned(8))) apriltag_families_t;
```

5. Enable family in python code.

### Expected behaviour
The AprilTag should be detectable.

### Actual behaviour
The AprilTag is not detected. Interestingly, if ```.h = 12```, the tags can be detected but with the wrong ID, I'm not sure I understand how the hamming distance influences the detection.

### Describe your setup

**IDE version:**
Qt Creator 4.5.2

**Firmware version:**

**Operating system:**
Ubuntu 18.04.3

### For developers

**GCC version:**
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

**Build target:**
[Kendryte GNU toolchain for RISC-V](https://github.com/kendryte/kendryte-gnu-toolchain) (using MaixPy project)

kwagyeman:
Please send a PR for integrating the new tag type. Given this is a highly technical feature addition just for you it shouldn't be to hard for you to add this to the code base.

PR #804, opened 2020-05-19
Add support for April Tag TAG36H8 (#802)

KBacsa:
As far as I know, this fails on the Sipeed MAix BiT for RISC-V.


iabdalkader:
There's not enough flash to merge this.

kwagyeman:
What is in AprilTag.c? How many tags did you add? 36,000+?

The algorithm will search that table linearly about 16 times in a row trying to match a tag. It's probably too large. You should reduced the number of tags in that family.

KBacsa:
I would like to add a family that would have at least 10,000+ tags. The choices are either adding a 7x7 tag or reducing the hamming distance for 6x6 tags. Is there a smarter way of doing this, or are AprilTags simply not designed for a higher number of detections ?

kwagyeman:
The original code allocates a O(1) lookup table by mallocing 30 MBs+ of memory and then computing each possible tag variant with 1 and 2 bit hamming code errors.

I removed that and used a linear table search instead as 30MB would not fit within the 384 KB of RAM on the M7 when I ported the code.

If you want to do a large number of tags it's possible. However, the algorithm will need to be modified to allow a tag library to be loaded into memory.

The original AprilTag code worked this way so it wouldn't be that difficult to handle. We could add a parameter that lets you pass a tag family. Then, it could internally malloc that tag family with a large lookup table.

However, the O(1) thing burns a ton of ram. A 10K tag family will require probably hundreds of Megabytes of RAM. So, the best you could do would be a linear search... But, that will be very slow.

Just to give you a RAM usage amount... The O(1) table has to make a code per possible bit error per tag for the number of but errors you want. So, with 36-bit tags that's 36 possible variants per tag. With 2 bit errors it's 36 * 36 possible tags per tag.

PR #809, opened 2020-05-24
fixed software JPEG encode of 1-bpp images

bitbank2:
For the hardware JPEG encoder, it wants YUV input and is happy with Y values of 0x00 and 0xff to be min and max. The software JPEG encoder wants YCbCr input which wants min/max values of Y to be 0x80 and 0x7F. I fixed the output for that.

### DIFF
--- a/src/omv/img/jpeg.c
+++ b/src/omv/img/jpeg.c
@@ -895,8 +895,8 @@ void jpeg_get_mcu(image_t *img, int mcu_w, int mcu_h, int x_offs, int y_offs, in
                     int index = (y * iPitch) + (x_offs>>3); // get byte offset
                     uint8_t *s = &img->data[index];
                     u8Pixels = s[0]; // get 8 binary pixels (1 byte)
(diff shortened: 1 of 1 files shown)


iabdalkader:
@bitbank2 Did you test this ?

bitbank2:
yes

iabdalkader:
It doesn't seem to have any effect.
image

kwagyeman:
There are two bugs. The second one is with grayscale images that have been thresholded.

kwagyeman:
https://github.com/bitbank2/openmv/blob/61739d6c40b3d83cbc00e3ae9a297d2c3bf6e67b/src/omv/img/jpeg.c#L923

Should be xor and not subtract. Subtract overlfows into the next pixels.

bitbank2:
ok, I'll fix that
I ran the binary filter sketch before submitting this PR and it worked. Let me look again.

bitbank2:
try it now

PR #830, opened 2020-06-14
Fixed buffer size on the 2/3 and reduced the frame buffer to 20M on Plus.

kwagyeman:
Fixed buffer size on the 2/3 and reduced the frame buffer to 20M on Plus.

OMV_RAW_BUF_SIZE doesn't prevent fb alloc from running into it's area currently. With the coming double buffer commit fb alloc will not easily walk into the frame buffer area anymore and will respect the buffer size except for cases where it known it can (i.e. in single buffer mode - in double buffer mode it will not re-use frame buffer memory).

### DIFF
--- a/src/omv/boards/OPENMV2/omv_boardconfig.h
+++ b/src/omv/boards/OPENMV2/omv_boardconfig.h
@@ -77,8 +77,8 @@
 #define OMV_MAIN_MEMORY     CCM     // data, bss, stack and heap
 #define OMV_DMA_MEMORY      SRAM2   // Misc DMA buffers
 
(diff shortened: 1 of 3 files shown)


iabdalkader:
Note: `OMV_FB_SIZE` and `OMV_FB_ALLOC_SIZE` are only used in the linker script script, to make sure FB + minimum FB alloc memory fit the used memory region. Changing them has No effect as long as `OMV_FB_SIZE + OMV_FB_ALLOC_SIZE <= OMV_FB_MEMORY size`. Wasn't really a bug, but thanks for making things more consistent.

PR #831, opened 2020-06-14
Simplified frame buffer access code

kwagyeman:
For the double buffer commit we shouldn't have code all over the place directly accessing the frame buffer. It needs to go through a function which returns the frame buffer you are working on out an array of frame buffers.

The wrapper methods in py_helper are applicable to the code state before the frame double buffer commit.


iabdalkader:
This is not needed anymore

kwagyeman:
This PR did a bunch of other stuff... I will redo it.

kwagyeman:
Will redo this.

PR #832, opened 2020-06-14
Expose fb_alloc_pointer for snapshot

kwagyeman:
Renames the pointer used for fb_alloc and exposes it for use with snapshot(). 

fb_alloc_pointer_snapshot is a copy of the pointer which snapshot will update when it's called to adjust how double buffers are setup given user buffer allocation. It's convenient to store here.

### DIFF
--- a/src/omv/fb_alloc.c
+++ b/src/omv/fb_alloc.c
@@ -12,7 +12,7 @@
 #include "omv_boardconfig.h"
 
 extern char _fballoc;
-static char *pointer = &_fballoc;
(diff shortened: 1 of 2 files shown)


iabdalkader:
>> fb_alloc_pointer_snapshot is a copy of the pointer which snapshot will update when it's called to adjust how double buffers are setup given user buffer allocation. It's convenient to store here.

Please keep this change out of this PR until it's actually needed so there's a context where it makes sense.

iabdalkader:
Actually this whole PR is not needed right now.

PR #835, opened 2020-06-15
Update fb access

iabdalkader:
This is a redo of PR #831 using the new framebuffer some more. Now nothing in omv/py/* uses MAIN_FB anymore, hopefully all the other code will be updated to use the new functions.

Additionally this code decouples framebuffer from MicroPython code (no exceptions in framebuffer.c) mostly with the exception of mpprint, and from imlib so it makes the new functions available to all the code that don't need these dependencies.

### DIFF
--- a/src/omv/framebuffer.c
+++ b/src/omv/framebuffer.c
@@ -156,6 +156,45 @@ void fb_update_jpeg_buffer()
     }
 }
 
+int32_t framebuffer_get_x()
+{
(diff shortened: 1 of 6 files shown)


kwagyeman:
I will replay my simplification code onto of this. I still don't like the fact that there's so much frame buffer code in the imglib.

iabdalkader:
@kwagyeman updated.

PR #837, opened 2020-06-16
fixes ulab compilation error

v923z:
This is a partial patch for https://github.com/v923z/micropython-ulab/issues/124


iabdalkader:
I'll fix the Makefiles in the same PR that updates the submodule so it doesn't break.

PR #845, opened 2020-07-13
update support for portenta and its camera carrier

pnndra:


src/omv/sensor.c
```diff
@@ -1303,6 +1303,8 @@ int sensor_snapshot(sensor_t *sensor, image_t *image, streaming_cb_t streaming_c
     // If two frames fit in ram, use double buffering in streaming mode.
     doublebuf = ((length*2) <= OMV_RAW_BUF_SIZE);
 
+    HAL_DCMI_EnableCrop(&DCMIHandle);
+    HAL_DCMI_ConfigCrop(&DCMIHandle,0,0,w-1,h-1);
```
iabdalkader, line 1307: I suggest adding this to avoid breaking something else, you could also add a flag to `sensor_t` like `use_hw_cropping`, and set it to `true` in hm01b0.c.
```suggestion
    #if OMV_ENABLE_HM01B0
        HAL_DCMI_EnableCrop(&DCMIHandle);
        HAL_DCMI_ConfigCrop(&DCMIHandle,0,0,w-1,h-1);
    #endif
```

PR #865, opened 2020-07-31
portenta specific fixes to share I2C with camera

pnndra:


src/omv/sensor.c
```diff
@@ -237,8 +237,45 @@ void sensor_init0()
...
+            #if defined(I2C_FASTMODEPLUS_I2C3)
+            HAL_I2CEx_EnableFastModePlus(I2C_FASTMODEPLUS_I2C3);
+            #endif
+        } else {
+            #if defined(I2C_FASTMODEPLUS_I2C4)
+            HAL_I2CEx_EnableFastModePlus(I2C_FASTMODEPLUS_I2C4);
+            #endif
+        }
+    }
+#endif
```
iabdalkader, line 275: That's the same as calling:
```suggestion
cambus_init(&sensor.i2c, SCCB_I2C, SCCB_TIMING);
#endif
```

iabdalkader: Also note the I2C handle in the sensor struct will Not be re-initialized, and that handle is declared on the stack it will be gone after init0.

pnndra: In my understanding before init0 there's no data structure initialized that can be passed to campus init. Since I just need to reinitialise i2c peripheral it's fine to set up a temporary struct in stack just to turn on the peripheral.

iabdalkader: > In my understanding before init0 there's no data structure initialized that can be passed to campus init

No you should be able to use the I2C handle in `sensor_t` at this point, it's statically allocated. On the first run `sensor_init0()` is called from `main()` which will initialize `sensor->i2c`, and then `sensor_init()` is called it will re-initialize sensor->i2c (so on first run `sensor->i2c` initialized twice, but that's okay). On later runs, only `sensor_init0()` is called and will reinitialize sensor->i2c. So unless I'm missing something, this should be enough to fix the issue, can you give it a try ?
```
    // Set fb_enabled
    JPEG_FB()->enabled = fb_enabled; // controlled by the IDE.
    #ifdef PORTENTA
    cambus_init(&sensor.i2c, SCCB_I2C, SCCB_TIMING);
    #endif
}
```

PR #866, opened 2020-07-31
added system reset to bootloader command for portenta

pnndra:


iabdalkader:
It's best to keep it in a new command, don't want any side effects on the other boards.

iabdalkader:
Okay, @pnndra can you please move that code to the reset command in a #ifdef ? This should make everyone happy.

iabdalkader:
I reverted the PR, please send it again.

pnndra:
Hi. You told me you use the reset command also for other purposes than fw upgrade. If this is the case we need to have two separate commands, one that resets the board and one that forces reboot to bootloader. The reboot to bootloader command won't have any difference other than writing a flag in a backup register to communicate BL it should not execute application.on your boards these two commands will be perfectly equal and if we have one conditionally defined then the ide as well would have to issue them based on vid/pid. I would still suggest to keep this as is but let me know and I'll resend the pr accordingly

iabdalkader:
That's right, sorry I forgot about that. @kwagyeman if we put the RTC code in the old reset command, it will always jump to the bootloader... The IDE must send a specific command to the Portenta to jump to the bootloader.

kwagyeman:
I see, so, with the Portenta when I am trying to get to the bootloader I will need to use this different command versus resetting the board correct?

iabdalkader:
That's right.

Issue #743, opened 2020-03-18
btree Error


skylin008:
### Steps to reproduce
1. make
2. show as follow error:
Warning: FROZEN_MPY_DIR is deprecated in favour of FROZEN_MANIFEST
CC ../../extmod/modbtree.c
In file included from ../../extmod/modbtree.c:37:0:
../../lib/berkeley-db-1.xx/PORT/include/db.h:1:1: error: expected identifier or '(' before '.' token
 ../../include/db.h
 ^
In file included from ../../extmod/modbtree.c:38:0:
../../lib/berkeley-db-1.xx/PORT/include/../../btree/btree.h:89:3: error: data definition has no type or storage class [-Werror]
 } PAGE;
   ^~~~
../../lib/berkeley-db-1.xx/PORT/include/../../btree/btree.h:89:3: error: type defaults to 'int' in declaration of 'PAGE' [-Werror=implicit-int]
../../lib/berkeley-db-1.xx/PORT/include/../../btree/btree.h:127:2: error: unknown type name 'pgno_t'
  pgno_t pgno;   /* page number stored on */
  ^~~~~~
../../lib/berkeley-db-1.xx/PORT/include/../../btree/btree.h:157:2: error: unknown type name 'recno_t'
  recno_t nrecs;   /* number of records */
  ^~~~~~~
../../lib/berkeley-db-1.xx/PORT/include/../../btree/btree.h:158:2: error: unknown type name 'pgno_t'
  pgno_t pgno;   /* page number stored below */
  ^~~~~~
../../lib/berkeley-db-1.xx/PORT/include/../../btree/btree.h:247:2: error: unknown type name 'pgno_t'
  pgno_t pgno;   /* the page number */
  ^~~~~~
../../lib/berkeley-db-1.xx/PORT/include/../../btree/btree.h:248:2: error: unknown type name 'indx_t'
  indx_t index;   /* the index on the page */
  ^~~~~~
../../lib/berkeley-db-1.xx/PORT/include/../../btree/btree.h:252:2: error: expected specifier-qualifier-list before 'PAGE'
  PAGE *page;   /* the (pinned) page */
  ^~~~
../../lib/berkeley-db-1.xx/PORT/include/../../btree/btree.h:282:2: error: unknown type name 'DBT'
  DBT  key;   /* B: Saved key, or key.data == NULL. */
  ^~~
[...]

PR #877, opened 2020-08-11
added non-blocking snapshot functions

frank26080115:
Added a non-blocking way of taking snapshots. This makes it easier to write responsive applications that require long exposure captures.

Please add the following to the `omv.sensor.rst` file

    .. function:: sensor.snapshot_start()

       Initiates a snapshot, but returns immediately. Use `sensor.snapshot_check()` to see if the image is ready. Use `sensor.snapshot_finish()` to retrieve the image.

    .. function:: sensor.snapshot_check()

       Returns whether or not the sensor has finished the snapshot started with `sensor.snapshot_start()`. True if the image is ready. Use `sensor.snapshot_finish()` to retrieve the image.

    .. function:: sensor.snapshot_finish()

       Returns the image captured after using `sensor.snapshot_start()`. If the image is not ready, it will wait until it is (or timeout), then return the image.

I have tested this on my OpenMV H7 Plus


kwagyeman:
Thanks for the PR. I have plans to make snapshot to support double buffering and N deep image buffering soon. What you've done is possible to support. However, I'd probably want the API to be worded in a particular way. I'll take a look more closely and what you've done and ask for some changes to what the double buffering API will look like.

src/omv/py/py_sensor.c
```diff
@@ -101,6 +101,45 @@ static mp_obj_t py_sensor_snapshot(uint n_args, const mp_obj_t *args, mp_map_t *
...
+        nlr_raise(mp_obj_new_exception_msg_varg(&mp_type_RuntimeError, "Capture Start Failed: %d", ret));
+    }
+
+    return mp_const_none;
+}
+
+static mp_obj_t py_sensor_snapshot_check()
+{
+    bool status;
+    sensor.snapshot_check(&sensor, &status);
```
kwagyeman, line 126: Make this just return the bool. Then it's one line.

frank26080115:
All I really did is split the original snapshot function in two, and remove the streaming callback from my variation. No change was done to the original snapshot function, so this is drop in friendly with no risk to existing code.

I know this opens up a big can of worms... like what happens if the user screws around with the camera and buffers while a capture is on-going.

Looking forward to what you decide to implement. I will continue my own project with my changes for now.

By the way, I managed to build this while on Windows, the trick is to use Ubuntu for Windows and fix the problem in https://github.com/openmv/openmv/issues/743 , I posted a solution to the problem there

kwagyeman:
Send the doc as a PR to: https://github.com/openmv/micropython/blob/c2938c32ef4c78d5c92e861db7367a3312ef6454/docs/library/omv.sensor.rst

Thanks for the good idea for a non-blocking idea!

Double buffering is my next thing to work on after getting the Arduino Interface Library out. Max FPS and video recording will then be possible.

src/omv/py/py_sensor.c
frank26080115: I made commit 3f1d58d that makes these changes for you but I can't really test the stream callback

kwagyeman:
Okay, thanks, going to sit on this for a while given the double buffering code. But, I'll see what we can do to make a non-blocking method.

iabdalkader:
> I know this opens up a big can of worms... like what happens if the user screws around with the camera and buffers while a capture is on-going.

It does, if we merge this we have to maintain it and handle all the possible ways these functions can be used (or misused), to add to what you mentioned, we need to handle `snapshot_start` being called more than once, or `snapshot_finish` called multiple times, or before/without `snapshot_start`, or buffers allocated mid-snapshot. In its current state, it just doesn't handle any of that. The same concerns apply to modifying snapshot to be non-blocking, so I'm just saying we have to be very careful with this feature.

kwagyeman:
This is not going to be needed in the future for an updated version of our driver. I will implement triple buffering as an option you can just turn on. When you do so the driver will generate 3 frame buffers with the camera driver delivering the latest image to 1 of the two buffers you are not working on. This will make the camera interface non-blocking and deliver you the latest image always when you want it (you'll also be technically able to run at frame rates faster than the camera can deliver images).

Not sure exactly how I'm make an image fifo, but, otherwise, it will be fine.

PR #905, opened 2020-09-21
Interpolate pixels for lens correction and rotation_corr

andrewleech:
The algorithm for lens correction calculates a table to map source pixels to corrected destination.
The map is calculated in float, but then it rounds each position to an int to select the source pixel to use.

With this change I've instead kept the source coordinates as float and take four neighboring source pixels, using the decimal place to select a percentage of each.

This is done in both x and y axis so the destination pixel is basically a weighed average of 4 source pixels.

eg. if the source coordinate is x=5.2, y=90.7 for dest[3, 80] I use:
```
dest[x=3, y=80] = ((source[x=5, y=90] * 0.8 + source[x=6, y=90] * 0.2) * 0.3) + 
                   (source[x=5, y=91] * 0.8 + source[x=6, y=91] * 0.2) * 0.7))
```

Note: this does incur a significant speed hit, in my initial testing it practically reverses the speed improvement recently made with https://github.com/openmv/openmv/commit/de98dfd3730293f02f6b73d6abccfad3ef56318b 
However, the quality improvement with some stronger curvature lenses is significant, which also improves the reliability of barcode reading dramatically.

Eg. `lens_corr`
lens_corr

The `rotation_corr` function works very similarly and benefits significantly from the same change.
rotation_corr

### DIFF
--- a/src/omv/img/apriltag.c
+++ b/src/omv/img/apriltag.c
@@ -12446,13 +12446,14 @@ void imlib_rotation_corr(image_t *img, float x_rotation, float y_rotation, float
 
                     for (int y = 0, yy = h; y < yy; y++) {
                         uint32_t *row_ptr = IMAGE_COMPUTE_BINARY_PIXEL_ROW_PTR(img, y);
+                        int bw = ((w + UINT32_T_MASK) >> UINT32_T_SHIFT);
                         for (int x = 0, xx = w; x < xx; x++) {
-                            int sourceX = fast_roundf(T4_00*x + T4_01*y + T4_02);
-                            int sourceY = fast_roundf(T4_10*x + T4_11*y + T4_12);
+                            float sourceX = T4_00*x + T4_01*y + T4_02;
+                            float sourceY = T4_10*x + T4_11*y + T4_12;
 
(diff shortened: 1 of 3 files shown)


kwagyeman:
Hey, we have bicubic and bilinear scaling done with integers only written by our performance optimizer coming soon. Thanks for the changes, I’m going to hold off accepting this PR. However, whatever ideas you have that we haven’t already thought of will be integrated.

andrewleech:
That sounds even better :-) 
I'll be glad if there's anything here that can help, feel free to take anything/nothing as suits best.

Looking forward to the new stuff when ready!

kwagyeman:
To note, we will have a new rendering pipeline being merged into our firmware soon that pushes speed to the max for all this stuff (e.g. a different tight optimized loop for all operations under one function call). We will then be using that for all image conversion operations in the future as once source code function for everything. This will improve speed and reduce code size.

kwagyeman:
To note, we will have a new rendering pipeline being merged into our firmware soon that pushes speed to the max for all this stuff (e.g. a different tight optimized loop for all operations under one function call). We will then be using that for all image conversion operations in the future as once source code function for everything. This will improve speed and reduce code size.

kwagyeman:
Finally, bicubic, bilinear, and area scaling will be available for all functions.

PR #932, opened 2020-10-23
Kwabena/fix rgb565 inversion 2

kwagyeman:
Removes the legacy RGB565 inversion in our code base. This is now finally fixed.

This speeds up a lot of RGB565 methods since undoing the inversion doesn't have to be done anymore.
This means we can finally use STM32 hardware like DMA2D and the LCD driver.
This PR is made of 4 commits.

1st commit -> Fixes RGB565 byte reversed data coming from the cameras.
2nd commit -> Removes the old rgb2rgb table as its not the best idea...
3rd commit -> Fixes RGB565 inversion in our lookup tables.
4th commit ->Cleans up the imglib by removing all the old RGB565 methods and then fixes functions that used them.


src/omv/img/jpeg.c
```diff
@@ -1303,7 +1303,7 @@ void jpeg_read_geometry(FIL *fp, image_t *img, const char *path, jpg_read_settin
         {
             uint16_t size;
             read_word(fp, &size);
-            size = IM_SWAP16(size);
+            size = __REV16(size);
```
iabdalkader, line 1306: Please keep the macros the IM_SWAPxx macros.

src/omv/py/py_lcd.c
```diff
@@ -327,7 +327,7 @@ static mp_obj_t py_lcd_display(uint n_args, const mp_obj_t *args, mp_map_t *kw_a
                 if (IM_IS_GS(arg_img)) {
                     for (int j=0; j<rect.w; j++) {
                         uint8_t pixel = IM_GET_GS_PIXEL(arg_img, (rect.x + j), (rect.y + i));
-                        line[j] = IM_RGB565(IM_R825(pixel),IM_G826(pixel),IM_B825(pixel));
+                        line[j] = COLOR_Y_TO_RGB565(pixel);
```
iabdalkader, line 330: The RGB case is probably broken now.

iabdalkader (changes requested):

src/omv/img/jpeg.c
iabdalkader: If it's not needed then please remove IM_SWAP_XX from imlib.h

src/omv/py/py_lcd.c
kwagyeman: Fixed

PR #938, opened 2020-10-28
Remove shadow removal code

kwagyeman:
Removing useless code. Shadow removal only works well if you have a shadow free image to start with... which is not easy to produce in practice in a real application.


iabdalkader:
Build fails on all cams, please fix and reopen.

PR #942, opened 2020-10-30
Added pause parameter to Image_Reader

StuartBox:
In order to speed up access to the SD card when reading back video for other non vision purposes like network playback.


kwagyeman (approved):

kwagyeman:
Cool

iabdalkader:
@StuartBox For what it's worth, you inverted the pause arg  (False will pause), I fixed it in #1272. This revealed a deeper issue though that causes a lockup when big frames are read too fast.

PR #948, opened 2020-11-08
Remove external libraries ulab replaces

kwagyeman:
ULAB replaces all this stuff.

Need to save space internally.


iabdalkader (changes requested):
Need to remove this as well:

unittest/script/22-frozen_modules.py

Or replace with another test

PR #1001, opened 2020-11-25
Remove the LCD code when there's no output defined

kwagyeman:
Probably should disable the module via MicroPython versus here.


iabdalkader:
Will add a flag to enable/disable py_lcd

PR #1017, opened 2020-12-01
Fix FIR I2C bus arbitration lost error.

iabdalkader:
* If the FIR I2C bus is interrupted during a read/write the I2C device misbehaves and gets stuck pulling SDA low and the next time the device is re-initialized the I2C controller detects an arbitration lost condition.
* The most affected sensor seems to be the AMG8833 on the F4, F7 and H7 micros and the MLX90640 on the F4 mirco.
* Add a function to switch SCL to GPIO and pulse SCL 10000 times to attempt to unstuck the I2C device, it will attempt this 1 time before giving up.
* This seems to work very well on all the tested micro-controllers.

### DIFF
--- a/src/omv/py/py_fir.c
+++ b/src/omv/py/py_fir.c
@@ -98,7 +98,7 @@ static uint8_t height = 0;
 static uint8_t IR_refresh_rate = 0;
 static uint8_t ADC_resolution = 0;
 
-static I2C_HandleTypeDef fir_i2c = {0};      // SCCB/I2C bus.
+static I2C_HandleTypeDef fir_i2c = {0};      // FIR/I2C bus.
 
 static enum {
     FIR_NONE,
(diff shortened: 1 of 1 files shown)


src/omv/py/py_fir.c
```diff
@@ -115,6 +115,29 @@ static void test_ack(int ret)
...
+    GPIO_InitStructure.Pull  = GPIO_NOPULL;
+    GPIO_InitStructure.Speed = GPIO_SPEED_HIGH;
+    GPIO_InitStructure.Mode  = GPIO_MODE_OUTPUT_PP;
+    GPIO_InitStructure.Pin   = FIR_I2C_SCL_PIN;
+    HAL_GPIO_Init(FIR_I2C_PORT, &GPIO_InitStructure);
+
+    // Pulse SCL to recover stuck device.
+    for (int i=0; i<10000; i++) {
+        HAL_GPIO_WritePin(FIR_I2C_PORT, FIR_I2C_SCL_PIN, GPIO_PIN_SET);
+        for(volatile int x=0; x<100; x++);
```
kwagyeman, line 131: Maybe use MP delay us...

src/omv/py/py_fir.c
```diff
@@ -115,6 +115,29 @@ static void test_ack(int ret)
...
+        HAL_GPIO_WritePin(FIR_I2C_PORT, FIR_I2C_SCL_PIN, GPIO_PIN_RESET);
+        for(volatile int x=0; x<100; x++);
```
kwagyeman, line 133: Same, please avoid busy loops.

kwagyeman (approved):

PR #1026, opened 2020-12-09
Changed RGB565 to GRAYSCALE for the Arduino examples

hpssjellis:


iabdalkader:
Fixed in the IDE.

kwagyeman:
Hi, the new IDE automatically forces the pixformat to grayscale for the an boards that have grayscale only examples. We're working on a color camera for the Portenta which will give you much better features (the ov5640).

PR #1068, opened 2020-12-30
faster hough x3

kidswong999:
1, change `sqrt(a*a + b*b)` to `(abs(a)+abs(b))/2`
2, only calculate edge (drop small mag)

# performance:
before:
QVGA RGB565: 34fps
QVGA GRAYSCALE: 26fps
VGA GRAYSCALE: 7fps

after:
QVGA RGB565: 41fps
QVGA GRAYSCALE:75fps
VGA GRAYSCALE: 21fps


kwagyeman:
@kidswong999 How did you think of these changes and why are they correct?

kidswong999:
The final result is the counting of accumulators, which does not require high precision, we just need to know which are higher. 

The non-edge point's mag is small and has little effect on the overall accumulators, but requires a lot of calculation, so I skiped it.Intuitively, the edge points are not calculated.

As for `sqrt(a*a + b*b)` to `(abs(a) + abs(b))/2`, this is a rough approximation that generally does not affect the relative number of accumulators. The reason is the same, we just need to know which are relatively high.

kwagyeman (approved):

PR #1079, opened 2021-01-01
Use SPI Callbacks

kwagyeman:
* Enabled USE_HAL_SPI_REGISTER_CALLBACKS for the F4/F7/H7.
* Switch to using callbacks with the LCD module.
* Remove stm32xx code for the callback.


src/omv/boards/OPENMV4/imlib_config.h
```diff
@@ -141,7 +141,7 @@
 #define IMLIB_ENABLE_HOG
 
 // Enable selective_search()
-#define IMLIB_ENABLE_SELECTIVE_SEARCH
+// #define IMLIB_ENABLE_SELECTIVE_SEARCH
```
iabdalkader, line 144: Send in an other PR if this is intended.

src/hal/stm32/f4/include/stm32f4xx_hal_conf.h
```diff
@@ -2,7 +2,7 @@
   ******************************************************************************
   * @file    stm32f4xx_hal_conf_template.h
   * @author  MCD Application Team
-  * @brief   HAL configuration template file. 
+  * @brief   HAL configuration template file.
```
iabdalkader, line 5: Do Not change HAL or libraries formatting.

iabdalkader (changes requested):

PR #1084, opened 2021-01-02
Add new ImageIO Type.

iabdalkader:
* This new module replaces the old ImageWriter and ImageReader and adds support for memory stream I/O.
* The API is more or less the same, with some minor changes.

Usage with file I/O:
```
stream = image.ImageIO("/stream.bin", "w") # Specify stream mode r or w
stream.read/write(image)   # checks for stream mode
stream.close()
```

Usage with memory I/O:
```
stream = image.ImageIO((width, height, bpp), n_frames) # Must specify the number of frames to pre-alloc.
stream.write(image)
stream.seek(0)
img = stream.read()
stream.close() # does nothing for now.
```

### DIFF
--- a/src/omv/modules/py_imageio.c
+++ b/src/omv/modules/py_imageio.c
@@ -276,7 +276,7 @@ mp_obj_t py_imageio_close(mp_obj_t self)
         file_close(&stream->fp);
     #endif
     } else if (stream->type == IMAGE_IO_MEMORY_STREAM) {
-        fb_alloc_free_till_mark();
+        fb_alloc_free_till_mark_past_mark_permanent();
     } else {
         mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Invalid image stream"));
     }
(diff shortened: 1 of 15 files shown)


kwagyeman (changes requested):
Update the example scripts. Also, you cannot just delete the old code. You need to make sure there's a class that keeps the old stuff working.

kwagyeman (changes requested):
Please update the examples and add a flush() method that does everything in close but keeps the file open.

src/omv/modules/py_imageio.c
```diff
@@ -0,0 +1,359 @@
...
+        stream->f_count = mp_obj_get_int(args[1]);
+        stream->w       = mp_obj_get_int(image_info[0]);
+        stream->h       = mp_obj_get_int(image_info[1]);
+        stream->bpp     = mp_obj_get_int(image_info[2]);
+        image_t image = {.w = stream->w, .h = stream->h, .bpp = stream->bpp, .pixels = NULL};
+        stream->f_size  = image_size(&image);
+        stream->offset  = 0;
+
+        fb_alloc_mark();
+        stream->buffer = fb_alloc(stream->f_size * stream->f_count, FB_ALLOC_NO_HINT);
```
kwagyeman, line 334: fb_alloc_mark_permanent();

PR #1090, opened 2021-01-03
Formatting cleanup

kwagyeman:
Removed FLT_MAX stuff. It's clean with no side affects.

I'd prefer if you can merge this as I have other edits to this file that I made which I removed from this commit that result in merge conflicts without all these changes.


iabdalkader:
image

PR #1100, opened 2021-01-04
Move vars into static namespace

kwagyeman:


src/omv/modules/py_fir.c
```diff
@@ -69,10 +69,10 @@ static enum {
     #endif
 } fir_sensor = FIR_NONE;
 
-int fir_width = 0;
-int fir_height = 0;
-int fir_ir_fresh_rate = 0;
-int fir_adc_resolution = 0;
+static int fir_width = 0;
```
iabdalkader, line 72: Why isn't the bus static too ?

```
static cambus_t fir_bus = {};
```

kwagyeman: Yes, it should be. Let me fix.

PR #1115, opened 2021-01-09
Kwabena/get openmvpt lepton working

kwagyeman:
FLIR Lepton support in our firmware uses SPI2, the same as the LCD module. I choose to build support based off of the PYB SPI module so that DMA would continue to work and to reuse all their handlers.

However, for the OpenMV PT the FLIR Lepton is on SPI3... which is used by the OpenMV4/OpenMV4P for the camera lepton module. I spent quite a bit of time trying to figure out how I could integrate these two things together without having a bunch of code changes. Then I realized that I could just disable OMV_ENABLE_LEPTON to disable the camera module which turns off the custom SPI bus interrupt and DMA handler. This then frees me to simply use SPI3 from the PYB module and have no code changes. SPI3 is enabled here: https://github.com/openmv/micropython/pull/87

Thus the SPI bus resource sharing issue is solved.

...

As for the OpenMV PT, the FIR onboard shares the I2C bus with the main camera sensor. However, the LEPTON module is always in reset on turn on so it won't respond to I2C scanning. However, the I2C bus must be reinited on soft reset. The other code changes handle this.

...

I examined creating a single interrupt handler location in stm32fxxx_hal_msp.c per Ibrahim's request. However, as I wrote more code for this it became clear I was just duplicating things and using more precious .data space without any benefit. Just disabling OMV_ENABLE_LEPTON on the OPENMVPT is the best solution.

### DIFF
--- a/src/omv/ports/stm32/main.c
+++ b/src/omv/ports/stm32/main.c
@@ -489,6 +489,9 @@ soft_reset:
 
     // Initialise low-level sub-systems. Here we need to do the very basic
     // things like zeroing out memory and resetting any of the sub-systems.
+    py_lcd_init0();
+    py_fir_init0();
+    py_tv_init0();
     readline_init0();
     pin_init0();
     extint_init0();
@@ -505,9 +508,6 @@ soft_reset:
     #ifdef IMLIB_ENABLE_IMAGE_FILE_IO
     file_buffer_init0();
     #endif
-    py_lcd_init0();
-    py_fir_init0();
-    py_tv_init0();
     servo_init();
     usbdbg_init();
     #if MICROPY_HW_ENABLE_SDCARD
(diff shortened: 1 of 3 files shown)


src/omv/ports/stm32/main.c
```diff
@@ -489,6 +489,9 @@ int main(void)
 
     // Initialise low-level sub-systems. Here we need to do the very basic
     // things like zeroing out memory and resetting any of the sub-systems.
+    py_lcd_init0();
+    py_fir_init0();
+    py_tv_init0();
```
iabdalkader, line 494: Why is this important ?

iabdalkader (changes requested):
Please update micropython submodule.

src/omv/ports/stm32/main.c
kwagyeman: These modules deinit all resources when they run. Which include the cambus. So, this has to happen before sensor_init0().

PR #1120, opened 2021-01-13
ov5640 pclk calculation now uses current register values

frank26080115:
I realized why setting long exposure settings was only sort-of working, and had weird timings, and also failed frequently.

The PCLK calculations were simply not taking into account the changes I wrote into all the PLL registers via the hack you've suggested.

I added some additional register retrievals and calculated PCLK properly and now it's working soooo much better


kwagyeman (approved):

kwagyeman:
Thanks for doing this. I was just hard coding it before.

PR #1133, opened 2021-01-21
Use interpolated LAB table.

iabdalkader:
* Saves about 100KBs.


kwagyeman (changes requested):
Interpolation is very wrong. Doesn't work for any other colors.

What you are doing to the RGB565 pixel is just dropping one bit of blue. Given this, just add a step of 2 in the table gen script and mark the table as half the size. No other changes are needed.

kwagyeman (approved):
Retested again. It all works.

PR #1137, opened 2021-01-24
Fix issues with F7 LCD triple buffering

kwagyeman:
* SPI CS was triggering too close to data movement.
* Transmit IT is not safe to use on data in a byte struct that is not statically allocated.

The LCD driver with triple buffering is now stable on the F7.

### DIFF
--- a/src/omv/ports/stm32/modules/py_lcd.c
+++ b/src/omv/ports/stm32/modules/py_lcd.c
@@ -67,7 +67,7 @@ static int lcd_refresh = 0;
 static int lcd_intensity = 0;
 
 #ifdef OMV_SPI_LCD_CONTROLLER
-DMA_HandleTypeDef spi_tx_dma = {};
+static DMA_HandleTypeDef spi_tx_dma = {};
 
 static volatile enum {
     SPI_TX_CB_IDLE,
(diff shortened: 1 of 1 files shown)


src/omv/ports/stm32/modules/py_lcd.c
```diff
@@ -208,27 +212,27 @@ void spi_lcd_callback(SPI_HandleTypeDef *hspi)
...
                 if (!spi_tx_cb_state_on[framebuffer_head]) {
-                    spi_tx_cb_state = SPI_TX_CB_DISPLAY_OFF;
-                    framebuffer_tail = framebuffer_head;
                     OMV_SPI_LCD_CS_HIGH();
                     OMV_SPI_LCD_RS_ON();
+                    spi_tx_cb_state = SPI_TX_CB_DISPLAY_OFF;
+                    framebuffer_tail = framebuffer_head;
                     OMV_SPI_LCD_CS_LOW();
-                    HAL_SPI_Transmit_IT(OMV_SPI_LCD_CONTROLLER->spi, (uint8_t []) {0x28}, 1); // display off
+                    HAL_SPI_Transmit_IT(OMV_SPI_LCD_CONTROLLER->spi, (uint8_t *) display_off, sizeof(display_off));
```
iabdalkader, line 220: @kwagyeman Why not just use transmit, it's 1 byte ?

kwagyeman: It's an interrupt FSM. Transmit IT when finished generated a tx done interrupt moving the FSM to the next state. Also, that would block during the interrupt.

kwagyeman: Moving the CS line high code around adds a few nanoseconds of high time which fixes issues with the CS toggling too quickly.

PR #1154, opened 2021-02-05
Fix JPEG Binary Image Compression

kwagyeman:
Fix JPEG Binary Image Compression, handles issues when the image width is not a multiple of 8.


iabdalkader:
Is this a bug fix ? Which one @kwagyeman

PR #1173, opened 2021-02-14
Kwabena/cache alignment

kwagyeman:
Add cache alignment support for fb_alloc(). This will let us use fb_alloc for DMA buffers without issues. Also fixed the alignment of the frame buffer and jpeg buffer to be aligned to 32-byte cache lines.

### DIFF
--- a/src/omv/alloc/fb_alloc.c
+++ b/src/omv/alloc/fb_alloc.c
@@ -132,7 +132,15 @@ void *fb_alloc(uint32_t size, int hints)
         return NULL;
     }
 
-    size=((size+sizeof(uint32_t)-1)/sizeof(uint32_t))*sizeof(uint32_t);// Round Up
+    size = ((size + sizeof(uint32_t) - 1) / sizeof(uint32_t)) * sizeof(uint32_t); // Round Up
+
+    #if __DCACHE_PRESENT
+    if (hints & FB_ALLOC_CACHE_ALIGN) {
+        size = ((size + __SCB_DCACHE_LINE_SIZE - 1) / __SCB_DCACHE_LINE_SIZE) * __SCB_DCACHE_LINE_SIZE;
+        size += __SCB_DCACHE_LINE_SIZE;
+    }
+    #endif
+
     char *result = pointer - size;
     char *new_pointer = result - sizeof(uint32_t);
 
@@ -144,6 +152,7 @@ void *fb_alloc(uint32_t size, int hints)
     // size is always 4/8/12/etc. so the value below must be 8 or more.
     *((uint32_t *) new_pointer) = size + sizeof(uint32_t); // Save size.
     pointer = new_pointer;
+
     #if defined(FB_ALLOC_STATS)
     alloc_bytes += size;
     if (alloc_bytes > alloc_bytes_peak) {
@@ -151,6 +160,7 @@ void *fb_alloc(uint32_t size, int hints)
     }
     printf("fb_alloc %lu bytes\n", size);
     #endif
+
     #if defined(OMV_FB_OVERLAY_MEMORY)
     if ((!(hints & FB_ALLOC_PREFER_SIZE))
     && (((uint32_t) (pointer_overlay - OMV_FB_OVERLAY_MEMORY_ORIGIN)) >= size)) {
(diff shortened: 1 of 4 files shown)


src/omv/imlib/framebuffer.h
```diff
@@ -33,7 +33,9 @@ typedef struct jpegbuffer {
     int32_t enabled;
     int32_t quality;
     mutex_t lock;
-    uint8_t pixels[];
+    int64_t __padding;
```
iabdalkader, line 36: Why is this padding required ?

kwagyeman: Just to ensure the 32-byte alignment. The align macro also does this too. I wanted to show that 32-bytes of the structure were in use. But, it's unnecessary.

src/omv/alloc/fb_alloc.c
```diff
@@ -132,7 +132,15 @@ void *fb_alloc(uint32_t size, int hints)
         return NULL;
     }
 
-    size=((size+sizeof(uint32_t)-1)/sizeof(uint32_t))*sizeof(uint32_t);// Round Up
+    size = ((size + sizeof(uint32_t) - 1) / sizeof(uint32_t)) * sizeof(uint32_t); // Round Up
+
+    #if __DCACHE_PRESENT
+    if (hints & FB_ALLOC_CACHE_ALIGN) {
+        size = ((size + __SCB_DCACHE_LINE_SIZE - 1) / __SCB_DCACHE_LINE_SIZE) * __SCB_DCACHE_LINE_SIZE;
+        size += __SCB_DCACHE_LINE_SIZE;
```
iabdalkader, line 140:
```suggestion
```
Why do you add another 32 bytes it's already a multiple of 32. Any example why this is needed ?

src/omv/imlib/framebuffer.h
iabdalkader, line 36:
```suggestion
```

src/omv/alloc/fb_alloc.c
```diff
@@ -160,6 +170,16 @@ void *fb_alloc(uint32_t size, int hints)
         *new_pointer |= FB_OVERLAY_MEMORY_FLAG; // Add flag.
     }
     #endif
+
+    #if __DCACHE_PRESENT
+    if (hints & FB_ALLOC_CACHE_ALIGN) {
+        int offset = ((uint32_t) result) % __SCB_DCACHE_LINE_SIZE;
```
iabdalkader, line 176: How can offset != 0 here after aligning ?

iabdalkader (changes requested):

iabdalkader:
I think you also need to make sure _fballoc and _fballoc_overlay are aligned to 32 bytes in the linker script.

src/omv/alloc/fb_alloc.c
kwagyeman: This is what does the alignment. The before just makes the size larger so that I ensure after this step is done that enough memory was granted.

kwagyeman:
@iabdalkader  You don't need to do that. The alignment in the code takes care of it. Remember that there are like 4 byte pointers and etc on the stack. So, each alloced region has to be made bigger than necessary and then a pointer returned to an aligned area inside of it.

src/omv/alloc/fb_alloc.c
kwagyeman: This is needed for the offset code below. Here's what's happening:

size = ((size + __SCB_DCACHE_LINE_SIZE - 1) / __SCB_DCACHE_LINE_SIZE) * __SCB_DCACHE_LINE_SIZE;

Ensures that you are using up some number of cache lines. This makes sure that no data at the end of the buffer will be inside another cache line. However, we don't know if the data will start on a cache line.

size += __SCB_DCACHE_LINE_SIZE;

Gives us extra memory that will be eaten by the offset code below. We have no idea what alignment of RAM fb_alloc() may give us. So, there's no way to know how much of __SCB_DCACHE_LINE_SIZE will be used up. We could be aligned perfectly by default, or, 4, 8, 12, 16, 20, 24, or 28 bytes offset. If we are offset by an number then the offset code below will push the pointer forward and eat 28, 24, 20, 16, 12, 8, or 4 bytes respectively. In the case where we are perfectly aligned we use 32 bytes extra.

Back to: size = ((size + __SCB_DCACHE_LINE_SIZE - 1) / __SCB_DCACHE_LINE_SIZE) * __SCB_DCACHE_LINE_SIZE;

Now that the pointer returned is aligned to a cache line... then because we did the above we are using N number of cache lines. While this means we will have some unused data at the end of the cache line it's okay as if the processor accesses memory around the DMA buffer it won't end up caching part of any of the DMA buffer causing problems.

PR #1191, opened 2021-02-25
Add alignment for the M4

kwagyeman:
Makes fb_alloc() provide 32-byte alignment on the M4 without a cache to ensure DMA memory access alignment rules for not crossing 1KB boundaries. The number could be 16 below but I wanted it to match the F7/H7.

### DIFF
--- a/src/omv/alloc/fb_alloc.c
+++ b/src/omv/alloc/fb_alloc.c
@@ -12,6 +12,12 @@
 #include "framebuffer.h"
 #include "omv_boardconfig.h"
 
+#ifndef __DCACHE_PRESENT
+#define FB_ALLOC_ALIGNMENT 32 // Use 32-byte alignment on MCUs with no cache for DMA buffer alignment.
+#else
+#define FB_ALLOC_ALIGNMENT __SCB_DCACHE_LINE_SIZE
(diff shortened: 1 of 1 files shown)


src/omv/alloc/fb_alloc.c
```diff
@@ -12,6 +12,12 @@
 #include "framebuffer.h"
 #include "omv_boardconfig.h"
 
+// Use 32-byte alignment on MCUs with no cache (this allows the Cortex-M4 to benefit from DMA buffer alignment).
+#ifndef __DCACHE_PRESENT
+#define __DCACHE_PRESENT 1
+#define __SCB_DCACHE_LINE_SIZE 32
+#endif
+
```
iabdalkader, line 20: If you need to do something later with this flag, ex cache maintenance, you can't force define it like that. I suggest this and just use this value for alignment.
```suggestion
#ifndef __DCACHE_PRESENT
#define FB_ALLOC_ALIGNMENT 32
#else
#define FB_ALLOC_ALIGNMENT  __SCB_DCACHE_LINE_SIZE
#endif

```

iabdalkader (changes requested):

PR #1194, opened 2021-02-26
TV driver SIMD update

kwagyeman:
I discovered a faster way of doing this with JPEG and just want to make sure it's used in this code too.


iabdalkader:
Can you test this first, make sure it's still working ?

kwagyeman:
Yes, I already did that.

PR #1202, opened 2021-03-03
Subtle max limit improvement for SPI LCD DMA.

kwagyeman:
Since the DMA system is talking in halfwords and not bytes this is 16/2=8.


kwagyeman:
It works. Tested.

PR #1206, opened 2021-03-05
Move all DMA buffers to SRAM3

kwagyeman:
This PR moves all the DMA buffers for H7 based OpenMV Cam's to SRAM3. This must be done to get the camera DMA off the D2-to-D1 AHB bus which can get busy and bottlenecked when another DMA stream is reading data from SDRAM/AXI_RAM over the D2-to-D1 AHB bus.

Capture

From this diagram above you can see that if any DMA master in the D2 domain wants to access SDRAM/AXI_RAM there's only one bus to the D1 domain to do so. This means that if the camera DMA needs to write to D1 domain and another DMA master is reading/writing SDRAM/AXI_RAM then the camera DMA has to wait. Note that there's no QoS for the AHB matrix in domain D2. So, there's no priority fix you can do here if multiple DMA masters need to use the same resource.

Because of the shallow depth of the camera DMA fifos you can't really delay that long without the loss of data. So, by moving the DMA buffer much closer to the DMA engine we remove it being blocked for a long time when trying to write data.

...

I tested the changes on the H7 Plus. Using the TV shield with the OV5640 with triple buffering used to crash. It does not anymore.

I tested the changes on the OpenMV Pure Thermal. Using the LCD Output, plus reading the FLIR, plus running the OV5640 used to crash. It does not anymore.

I tested the changes on the Portenta. It continues to work as normal.

### DIFF
--- a/src/omv/boards/OPENMV4P/omv_boardconfig.h
+++ b/src/omv/boards/OPENMV4P/omv_boardconfig.h
@@ -142,13 +142,12 @@
 #define OMV_FFS_MEMORY          DTCM        // Flash filesystem cache memory
 #define OMV_MAIN_MEMORY         SRAM1       // data, bss and heap
 #define OMV_STACK_MEMORY        ITCM        // stack memory
-#define OMV_DMA_MEMORY          AXI_SRAM    // DMA buffers memory.
+#define OMV_DMA_MEMORY          SRAM3       // DMA buffers memory.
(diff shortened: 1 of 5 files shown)


iabdalkader:
Keep this open, will send a PR to add a DMA region in each domain, then you can rebase and change the misc DMA buffers memory, and be able to reserve DMA buffers in a specific domain.

PR #1218, opened 2021-03-09
Add DMA buffer allocator

iabdalkader:


kwagyeman:
I'd make the min block size like 128 bytes. This allows the size to more tightly fit dma regions.

PR #1221, opened 2021-03-14
Sped up jpeg bayer_to_ycbcr for encode

bitbank2:
I changed the bayer_to_ycbcr code for jpeg encode to separate the boundary checking part into a separate path and made use of unaligned reads to speed up the data gathering.


kwagyeman:
Nice, this reduced the latency from 40ms to 28/29ms

PR #1225, opened 2021-03-15
Filled out PWM TIMER init for FLIR/LCD

kwagyeman:
TIM_ETRPRESCALER_DIV1 is not a valid prescaler value. It happens to be 0 which is what we want but it's not the right define.

Also, filled in other fields with defaults.


iabdalkader:
Please edit commit message/title to something meaningful for the change log.

kwagyeman:
Done

PR #1232, opened 2021-03-17
Portenta example fixes

sebromero:
This PR removes the incompatible function calls to `sensor.set_auto_whitebal`. 
It also replaces all occurrences of `sensor.RGB565` with `sensor.GRAYSCALE`
There is already some magic replacement happening for the pixel format in the IDE but it doesn't work for all occurrences. `alloc_extra_fb` calls contain a reference to the pixel format too which doesn't get replaced.


iabdalkader:
@sebromero want me to merge this and fix the deepsleep example or are you going to fix it ?

sebromero:
@iabdalkader Done

iabdalkader:
@sebromero  I just noticed that a few examples actually assume RGB more than just setting the format, for example in a few scripts there's drawing with color triplets.. The code should be able to handle that, meaning it will probably just read the first value, but the colors will be wrong. Would you like to fix this now or merge as is ?

sebromero:
@iabdalkader Oh, sorry I didn't see that. Would probably be good to fix it before we merge. Or if you'd like to help with this, you could merge and fix it in a separate branch. Up to you.

iabdalkader:
@sebromero Let me take a look and see what needs to be fixed.

iabdalkader:
Let's do another round of updates later.

PR #1235, opened 2021-03-19
Move software jpeg encoder front end to new SIMD MCU code

kwagyeman:
This PR updates the JPEG code to use the new and tested get MCU front end developed for the H7 hardware JPEG compressor. The code is now shared between both software and hardware JPEG.

To ensure speed... I have ifdefs that switch things around in the output generated depending on if we are in software versus hardware mode.

I have tested this code on the H7/H7P/M7/M4.

I tested the hardware code for BINARY/GRAYSCALE/RGB565/BAYER.

I tested the software code for BINRARY/GRAYSCALE/RGB565/BAYER with 1x1, 2x1, and 2x2 sampling for each.

To support the NANO33 (Cortex-M3) I also added in intrinsics for non-m4/m7 processors. The code was tested and continued to work when using these. 

Additionally, I spent time trying to optimize the DCT methods in the software JPEG code. However, this did not yield any good results. It's not possible to improve the speed of the column-wise DCT with cortex-simd instructions. I could only get the inner loop down to 51 instructions from 59. As for the row wise DCT I was able to get that down to 40 instructions from 59. However, neither of these will produce much speedup. For example, debayering is noticeably faster because I reduced the instruction count to like 24 from 50.

I also tried to optimize the software UV 2x1/2x2 averaging. However, the cortex-m4/m7 SIMD instruction set can't help easily in this case given the layout of MCUs in RAM. The U/V MCUs would need to be interleaved to enable the use the __SHADD8() instruction. __USAD8 is not usable given the absolute value op it does.

### DIFF
--- a/src/hal/cmsis/include/cmsis_gcc.h
+++ b/src/hal/cmsis/include/cmsis_gcc.h
@@ -2174,9 +2174,25 @@ __STATIC_FORCEINLINE int32_t __SMMLA (int32_t op1, int32_t op2, int32_t op3)
  return(result);
 }
 
+#else
+
+__STATIC_FORCEINLINE uint32_t __UXTB(uint32_t op1)
+{
+  return op1 & 0xFF;
+}
+
+__STATIC_FORCEINLINE uint32_t __UXTB_RORn(uint32_t op1, uint32_t rotate)
+{
+  return (op1 >> rotate) & 0xFF;
+}
+
+__STATIC_FORCEINLINE uint32_t __SSUB16(uint32_t op1, uint32_t op2)
+{
+  return ((op1 & 0xFFFF0000) - (op2 & 0xFFFF0000)) | ((op1 - op2) & 0xFFFF);
+}
+
 #endif /* (__ARM_FEATURE_DSP == 1) */
 /*@} end of group CMSIS_SIMD_intrinsics */
-
 
 #pragma GCC diagnostic pop
 
--- a/src/omv/imlib/jpeg.c
+++ b/src/omv/imlib/jpeg.c
@@ -21,19 +21,524 @@
 #include "py/mphal.h"
 #endif
 
+#define MCU_W                       (8)
+#define MCU_H                       (8)
+#define JPEG_444_GS_MCU_SIZE        ((MCU_W) * (MCU_H))
+#define JPEG_444_YCBCR_MCU_SIZE     ((JPEG_444_GS_MCU_SIZE) * 3)
+
 // Expand 4 bits to 32 for binary to grayscale - process 4 pixels at a time
-static const uint32_t u32Expand[16] = {0x0, 0xff, 0xff00, 0xffff, 0xff0000,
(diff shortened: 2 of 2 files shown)


kwagyeman:
This is the code I came up with the the DCT row loop. It doesn't work though because of the __QADD8 which clamps too early causing bad image quality. But, the rest of the SIMD logic is sensible. However, there are too many packing and unpacking instructions in the code which don't actually do work.

        `for (int i = 8, *p = DU; i > 0; i--, p += 8, CDU += 8) {
                int t30 = *((uint32_t *) CDU);
                int t47 = __REV(*((uint32_t *) (CDU + 4)));
                int tadd = __QADD8(t30, t47);
                int tsub = __QSUB8(t30, t47);
        
                int t2_t0 = __SXTB16(tadd);
                int t3_t1 = __SXTB16_RORn(tadd, 8);
                int t10_t12 = __SASX(t3_t1, t2_t0);
                int t11_t13 = __SASX(t2_t0, t3_t1);
                int t11_t10 = __PKHTB(t11_t13, t10_t12, 16);
                int t12_t13 = __PKHBT(t11_t13, t10_t12, 16);
        
                p[0] = __SMUAD(t11_t10, 0x10001);
                p[4] = __SMUSD(t11_t10, 0x10001);
                p[2] = ((int) __SMUAD(t12_t13, (FIX_0_707106781 << 16) | (256 + FIX_0_707106781))) >> 8;
                p[6] = ((int) __SMUSD(t12_t13, (FIX_0_707106781 << 16) | (256 - FIX_0_707106781))) >> 8;
        
                int t5_t7 = __SXTB16(tsub);
                int t4_t6 = __SXTB16_RORn(tsub, 8);
                t10_t12 = __SADD16(t4_t6, t5_t7);
                int t11_t7 = __SASX(t5_t7, t4_t6 & 0xffff);
                int z2 = __SMUSDX((FIX_0_382683433 << 16) | (FIX_0_541196100 + FIX_0_382683433), t10_t12);
                int z4 = __SMUADX(((FIX_1_306562965 - FIX_0_382683433) << 16) | FIX_0_382683433, t10_t12);
                int z11 = __SMUAD(t11_t7, (FIX_0_707106781 << 16) | 256);
                int z13 = __SMUSD(t11_t7, (FIX_0_707106781 << 16) | 256);
        
                p[5] = (z13 + z2) >> 8;
                p[3] = (z13 - z2) >> 8;
                p[1] = (z11 + z4) >> 8;
                p[7] = (z11 - z4) >> 8;
            }`

kwagyeman:
Anyway, software JPEG is about as fast as it will get on the M4/M7. Will need the M55 to optimize the DCT loops.

kwagyeman:
U/V Averaging for JPEG 2x1/2x2 cannot be speed up with SIMD. But, the loop for that is tight.

iabdalkader:
Is this ready ?

kwagyeman:
Yes, it's all done and tested.

PR #1242, opened 2021-03-26
Ensure lepton is off

kwagyeman:
For any boards with a FLIR Lepton onboard this ensures the device is forced into reset and power down on init. Hardware pull-ups are not enough.

### DIFF
--- a/src/omv/ports/stm32/stm32fxxx_hal_msp.c
+++ b/src/omv/ports/stm32/stm32fxxx_hal_msp.c
@@ -188,11 +188,11 @@ void HAL_MspInit(void)
     OMV_AXI_QOS_LTDC_W_SET(OMV_AXI_QOS_LTDC_W_PRI);
     #endif
 
-    #if defined(DCMI_RESET_PIN) || defined(DCMI_PWDN_PIN) || defined(DCMI_FSYNC_PIN)
-    /* Configure DCMI GPIO */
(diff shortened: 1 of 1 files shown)


src/omv/ports/stm32/stm32fxxx_hal_msp.c
```diff
@@ -188,12 +188,12 @@ void HAL_MspInit(void)
     OMV_AXI_QOS_LTDC_W_SET(OMV_AXI_QOS_LTDC_W_PRI);
     #endif
 
-    #if defined(DCMI_RESET_PIN) || defined(DCMI_PWDN_PIN) || defined(DCMI_FSYNC_PIN)
```
iabdalkader, line 191: This makes sure that if none of the pins are defined, the unused struct doesn't cause a build error

iabdalkader (changes requested):

src/omv/ports/stm32/stm32fxxx_hal_msp.c
kwagyeman: Ah, yes, true.

iabdalkader: Maybe best to make a struct per pin, ex: gpio_reset_init, gpio_fsync_init etc...

PR #1245, opened 2021-03-29
Make sure dedicated analog pads are isolated

kwagyeman:
This makes sure the dedicated ADC channels that go to pads are not connected to their respective GPIO pin.

### DIFF
--- a/src/omv/ports/stm32/stm32fxxx_hal_msp.c
+++ b/src/omv/ports/stm32/stm32fxxx_hal_msp.c
@@ -220,6 +220,15 @@ void HAL_MspInit(void)
     #endif
 
     #endif // DCMI_RESET_PIN || DCMI_PWDN_PIN || DCMI_FSYNC_PIN
+
+    // This disconnects PA0/PA1/PC2/PC3 from PA0_C/PA1_C/PC2_C/PC3_C.
+    // PA0_C/PA1_C connect to ADC1/2 Channels P0/P1. PC2_C/PC3_C connect to ADC3 Channels P0/P1.
+    #if defined(MCU_SERIES_H7)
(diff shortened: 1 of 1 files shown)


kwagyeman:
You can read PA0_C/PA1_C via ADC.read(0/1)

src/omv/ports/stm32/stm32fxxx_hal_msp.c
```diff
@@ -220,6 +220,13 @@ void HAL_MspInit(void)
     #endif
 
     #endif // DCMI_RESET_PIN || DCMI_PWDN_PIN || DCMI_FSYNC_PIN
+
```
iabdalkader, line 223: @kwagyeman Can you please add a 1/2 lines comment here on what it's doing and why ? Also, please make sure this doesn't affect the DAC/ADC pin.

iabdalkader (changes requested):

src/omv/ports/stm32/stm32fxxx_hal_msp.c
kwagyeman: It doesn't affect the DAC/ADC pin. Tested that.

kwagyeman: Documented

kwagyeman:
I checked this. It appears PC2 on the 100-LFQP is actually PC2_C. Same for PC3.

So, we don't actually have the proper PC2/PC3 pins on the LFQP package but instead just the analog connected switch pin. So, these should be disabled. 

The Pure Thermal doesn't need them too. Only PA0_C and PA1_C.

So, I'll remove the extra lines.

PR #1254, opened 2021-04-02
Remove streaming mode.

iabdalkader:
* Will be replaced by double buffering soon.


kwagyeman (approved):
Woot!

Streaming mode made all of this so complex.

iabdalkader:
@kwagyeman It doubled the FPS for UVC streaming, but it's not needed anymore since double buffering can achieve the same.

PR #1256, opened 2021-04-04
Initial double buffering support.

iabdalkader:


src/omv/imlib/framebuffer.c
```diff
@@ -250,3 +253,8 @@ void framebuffer_set(int32_t w, int32_t h, int32_t bpp)
     framebuffer->h = h;
     framebuffer->bpp = bpp;
 }
+
```
kwagyeman, line 256: You need to make sure the get_buffer() method returns the right frame buffer. Right now, this will break a lot of code elsewhere.

PR #1258, opened 2021-04-05
Initial double buffering support (Take 2)

iabdalkader:


src/omv/imlib/framebuffer.c
```diff
@@ -237,14 +282,46 @@ uint32_t framebuffer_get_buffer_size()
...
+    }
+
+    // Initialize from the first buffer, copying x,y,w,h,u,v, etc...
+    framebuffer->x = old_fb->x;
+    framebuffer->y = old_fb->y;
+    framebuffer->w = old_fb->w;
+    framebuffer->h = old_fb->h;
+    framebuffer->u = old_fb->u;
+    framebuffer->v = old_fb->v;
+    framebuffer->x = old_fb->x;
```
kwagyeman, line 314: Repeat

src/omv/imlib/framebuffer.c
```diff
@@ -237,14 +282,46 @@ uint32_t framebuffer_get_buffer_size()
...
+    framebuffer->bpp  = -1;
+
+    // The second buffer needs to be initialized from fb0
+    if (framebuffer != &buffers[0] && framebuffer->pixels == &_fb_base) {
+        // If this is first time this buffer is used, adjust its pixels pointer
+        // offset within the raw buffer to point just after the first frame buffer.
+        framebuffer->pixels = &_fb_base + FB_ALIGN_SIZE(offset);
```
kwagyeman, line 321: Users are allowed to change the pixformat/framesize while a script is running. I don't see this recomputed ever again. You need a way to reset everything when a user changes any major setting.

iabdalkader: Yes, we need to reset buffers just like MAIN_FB()->bpp = -1;

src/omv/ports/stm32/sensor.c
```diff
@@ -1526,16 +1454,21 @@ int sensor_snapshot(sensor_t *sensor, image_t *image, uint32_t flags)
...
-        image->h = MAIN_FB()->h;
-        image->bpp = MAIN_FB()->bpp;
+        image->w      = MAIN_FB()->w;
+        image->h      = MAIN_FB()->h;
+        image->bpp    = MAIN_FB()->bpp;
         image->pixels = MAIN_FB()->pixels;
     }
+
+    // If double buffering is enabled, check if there's enough memory for two frames.
+    if ((flags & SNAPSHOT_DBLBUF) && ((length * 2) < framebuffer_get_buffer_size())) {
```
kwagyeman, line 1466: I see what you are trying to do here. However, there's a problem. While this works fine for a fast frame rate the snapshot of the next image is generated immediately after the first image capture. If I do something that takes 500ms then and call snapshot you're going to end up with a frame that's 500ms old. Not the latest frame.

This issue applies to pretty much any use case. If I'm running apriltags and call this... And I'm running at 20FPS I'll be running the algorithm always on a frame that's 50ms old since the snapshot is always queued by the last time you call this method.

We have quite a few users who like our product specifically because we have a low-latency system that gets you the latest image when you call snapshot. This will break their applications. Particularly given the camera is generating images at like 150 FPS... so, we're adding quite a bit of latency variably based on their main loop freq. Right now we have bounded latency at 1/FPS.

To get around this, double buffering has to be an interrupt driven architecture with a producer thread that runs in interrupt context mode capturing frames continuously and then the main thread just grabs and locks whatever frame is already available. This ensures that the main thread is always grabbing the latest frame from the camera.

Next, you have a choice on the number of buffers. With one buffer the interrupt thread is only ever allowed to work while the main thread is waiting for data. With two buffers the main thread and the interrupt thread ping-pong back and forth. The interrupt thread is always storing frames in this case and has priority for buffer access. The main main thread just grabs the latest buffer from the interrupt thread or waits until the interrupt thread is done capturing when it wants a frame. Because both threads do not run in lockstep the main thread will generally have to wait for frame because the interrupt thread will be busy capturing one. At the worst case the main thread will wait 1/FPS of the camera. But generally it will be about 0.5/FPS.

Finally, the best performance is achieve at 3 buffers and above with triple buffering where images are always captured by the camera by the interrupt thread and the main thread reads from one of the two buffers not in use by the interrupt thread. When the main thread wants to switch buffers it just moves to whatever was the latest buffer generated by the interrupt thread. The main thread is then never blocked. The same for the interrupt thread.

For more than 3 buffers you then keep the N or more previous frames. This is useful if we need to handle larger delays in the main thread. Since the two threads run independently you now get an elastic buffer effect.

I've already demonstrated this architecture for the LCD driver SPI/LTDC image output. The code to do triple buffering pointer management is really simple. I'm not sure what more than 3 looks like but it's likely a simple extension.

kwagyeman (changes requested):

src/omv/ports/stm32/sensor.c
iabdalkader: It's doable, I could experiment with that, alternatively we could use a timestamp or count VSYNCs to discard old frames.

PR #1263, opened 2021-04-14
Rename head/tail per convention

kwagyeman:
Tail is the front of the queue. Head is back.


src/omv/ports/stm32/modules/py_fir_lepton.c
```diff
@@ -175,13 +175,13 @@ void fir_lepton_spi_callback(const uint16_t *base)
 
         if (frame_ready) {
             // Update head which means a new image is ready.
```
iabdalkader, line 177:
```suggestion
            // Update tail which means a new image is ready.
```
missed one

iabdalkader (changes requested):
Missed one comment.

kwagyeman:
Commited the suggestion.

PR #1266, opened 2021-04-16
Fix mp_const_none comparison

kwagyeman:
Type checking was wrong.


kwagyeman:
@iabdalkader - Fixed.

PR #1285, opened 2021-04-28
Fix build errors encountered with GNU Arm Embedded Toolchain 10-2020-q4-major

jepler:
This series of fixes allows the firmware to be built using gcc10, rather than gcc7 as used in your CI process.

I'm investigating using gcc10, as this is what CircuitPython uses (it appears that micropython uses 9-2019-q4-0ubuntu1 at the moment).

I have **NOT** tested this on hardware, and one change does modify behavior.

Please let me know if you'd prefer this be broken into multiple PRs, as it touches multiple parts of the project.

### DIFF
--- a/src/drivers/winc1500/src/winc.c
+++ b/src/drivers/winc1500/src/winc.c
@@ -353,8 +353,8 @@ static void wifi_callback_sta(uint8_t msg_type, void *msg)
             memcpy(netinfo->mac_addr, con_info->au8MACAddress, WINC_MAC_ADDR_LEN);
 
             // Copy SSID.
+            con_info->acSSID[WINC_MAX_SSID_LEN-1] = 0;
             strncpy(netinfo->ssid, con_info->acSSID, WINC_MAX_SSID_LEN);
-            netinfo->ssid[WINC_MAX_SSID_LEN-1] = 0;
 
             async_request_done = true;
 			break;
--- a/src/drivers/winc1500/src/winc.c
+++ b/src/drivers/winc1500/src/winc.c
@@ -390,8 +390,8 @@ static void wifi_callback_sta(uint8_t msg_type, void *msg)
 
             // Copy BSSID and SSID.
             memcpy(wscan_result.bssid, scan_result->au8BSSID, WINC_MAC_ADDR_LEN);
+            scan_result->au8SSID[WINC_MAX_SSID_LEN-1] = 0;
             strncpy((char*) wscan_result.ssid, (const char *) scan_result->au8SSID, WINC_MAX_SSID_LEN);
-            wscan_result.ssid[WINC_MAX_SSID_LEN-1] = 0;
 
             // Call scan result callback
             scan_arg->cb(&wscan_result, scan_arg->arg);
--- a/src/drivers/winc1500/src/winc.c
+++ b/src/drivers/winc1500/src/winc.c
@@ -353,7 +353,8 @@ static void wifi_callback_sta(uint8_t msg_type, void *msg)
             memcpy(netinfo->mac_addr, con_info->au8MACAddress, WINC_MAC_ADDR_LEN);
 
             // Copy SSID.
-            strncpy(netinfo->ssid, con_info->acSSID, WINC_MAX_SSID_LEN-1);
+            strncpy(netinfo->ssid, con_info->acSSID, WINC_MAX_SSID_LEN);
+            netinfo->ssid[WINC_MAX_SSID_LEN-1] = 0;
 
             async_request_done = true;
 			break;
@@ -389,7 +390,8 @@ static void wifi_callback_sta(uint8_t msg_type, void *msg)
 
             // Copy BSSID and SSID.
             memcpy(wscan_result.bssid, scan_result->au8BSSID, WINC_MAC_ADDR_LEN);
-            strncpy((char*) wscan_result.ssid, (const char *) scan_result->au8SSID, WINC_MAX_SSID_LEN-1);
+            strncpy((char*) wscan_result.ssid, (const char *) scan_result->au8SSID, WINC_MAX_SSID_LEN);
+            wscan_result.ssid[WINC_MAX_SSID_LEN-1] = 0;
 
             // Call scan result callback
             scan_arg->cb(&wscan_result, scan_arg->arg);
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ script:
   # install ARM GCC
   - pushd .
   - cd ~ && mkdir gcc && cd gcc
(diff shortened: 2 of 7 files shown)


iabdalkader:
Thank you that's something I've been wanting to do for a while... Will review this and get back to you.

iabdalkader:
I think declaring the linker scripts variables as arrays should fix most of these warnings. For example: `extern char _vfs_buf[];`. I prefer not to hard-code the sizes as they differ from board to board and may change in the future, but if we must the sizes are probably defined in omv_boardconfig.h files.

jepler:
Thanks for the advice. I've rebased the branch to implement the `...[]` notation as suggested.  This passes locally with gcc10.  This also resolves a merge conflict in `py_winc.c`.

jepler:
Looks like there are problems in the portenta build with gcc10.  I don't plan to pursue them to completion at this time (fixes in submodules will be needed, in fact apparently way down in https://github.com/apache/mynewt-nimble.git).  Shall I revert (or rebase out) the commit that changes the toolchain so that it's possible to merge this, or what would you like?

iabdalkader:
Hi,

>  This also resolves a merge conflict in `py_winc.c`.

Yes I forgot to mention, I pushed that change to fix the build if WINC is disabled it failed with gcc10.

> Looks like there are problems in the portenta build with gcc10. I don't plan to pursue them to completion at this time (fixes in submodules will be needed, in fact apparently way down in https://github.com/apache/mynewt-nimble.git).

I fixed that issue by adding a flag for the nimble submodule, if you rebase now it should be fixed. That said, the Portenta build now adds an extra 20K for some reason, which overflows the flash... I will need some more time to figure out why this is happening. So you can either remove the Travis change commit, or wait for me to debug this issue (may end up reducing the fw size).

Finally, can you please revert all the WINC string related changes ? I found much easier ways to fix this, a) you can just increase WINC_MAX_SSID_LEN to 34 which means (WINC_MAX_SSID_LEN-1) is bigger than the src and will be null terminated. Or use WINC_MAX_SSID_LEN (not WINC_MAX_SSID_LEN-1) and then null terminate it, like so:
```diff
diff --git a/src/drivers/winc1500/src/winc.c b/src/drivers/winc1500/src/winc.c
index f643b5c1..1ffd323c 100644
--- a/src/drivers/winc1500/src/winc.c
+++ b/src/drivers/winc1500/src/winc.c
@@ -353,7 +353,8 @@ static void wifi_callback_sta(uint8_t msg_type, void *msg)
             memcpy(netinfo->mac_addr, con_info->au8MACAddress, WINC_MAC_ADDR_LEN);
 
             // Copy SSID.
-            strncpy(netinfo->ssid, con_info->acSSID, WINC_MAX_SSID_LEN-1);
+            strncpy(netinfo->ssid, con_info->acSSID, WINC_MAX_SSID_LEN);
+            netinfo->ssid[WINC_MAX_SSID_LEN-1] = 0;
 
             async_request_done = true;
                        break;
@@ -389,7 +390,8 @@ static void wifi_callback_sta(uint8_t msg_type, void *msg)
 
             // Copy BSSID and SSID.
             memcpy(wscan_result.bssid, scan_result->au8BSSID, WINC_MAC_ADDR_LEN);
-            strncpy((char*) wscan_result.ssid, (const char *) scan_result->au8SSID, WINC_MAX_SSID_LEN-1);
+            strncpy((char*) wscan_result.ssid, (const char *) scan_result->au8SSID, WINC_MAX_SSID_LEN);
+            wscan_result.ssid[WINC_MAX_SSID_LEN-1] = 0;
 
             // Call scan result callback
             scan_arg->cb(&wscan_result, scan_arg->arg);
```

jepler:
I have rebased the branch and dropped the two commits
```
ff26ab7d travis.yml: Use 10-2020q4, presently the newest arm toolchain
f40e66e2 winc1500: Fix possible string truncation of 32-byte SSIDs.
```
The remaining fixes feel safe, while the string manipulation changes feel tricky and I don't have HW to test them on.

In CircuitPython, we see firmware sizes go up and down as we change toolchain versions; it seems to be just one of those things.

iabdalkader:
> while the string manipulation changes feel tricky and I don't have HW to test them on.

Hi, the patch above is perfectly fine I tested it with OMV+WINC, it will copy the whole 33 bytes from `src` and make sure `dst` is null terminated, so please include it in this PR too.

Also I fixed all code size issues, if you rebase on master now everything should build fine with gcc10, so it would be great to also add the Travis change while we're at it, just to make sure that it doesn't get broken again soon by new code.

jepler:
Thanks for doing that testing!  Let's see what CI says.

src/drivers/winc1500/src/winc.c
```diff
@@ -353,7 +353,8 @@ static void wifi_callback_sta(uint8_t msg_type, void *msg)
             memcpy(netinfo->mac_addr, con_info->au8MACAddress, WINC_MAC_ADDR_LEN);
 
             // Copy SSID.
-            strncpy(netinfo->ssid, con_info->acSSID, WINC_MAX_SSID_LEN-1);
+            strncpy(netinfo->ssid, con_info->acSSID, WINC_MAX_SSID_LEN);
+            netinfo->ssid[WINC_MAX_SSID_LEN-1] = 0;
```
iabdalkader, line 357:
```suggestion
            con_info->acSSID[WINC_MAX_SSID_LEN-1] = 0;
            strncpy(netinfo->ssid, con_info->acSSID, WINC_MAX_SSID_LEN);
```

src/drivers/winc1500/src/winc.c
```diff
@@ -389,7 +390,8 @@ static void wifi_callback_sta(uint8_t msg_type, void *msg)
 
             // Copy BSSID and SSID.
             memcpy(wscan_result.bssid, scan_result->au8BSSID, WINC_MAC_ADDR_LEN);
-            strncpy((char*) wscan_result.ssid, (const char *) scan_result->au8SSID, WINC_MAX_SSID_LEN-1);
+            strncpy((char*) wscan_result.ssid, (const char *) scan_result->au8SSID, WINC_MAX_SSID_LEN);
+            wscan_result.ssid[WINC_MAX_SSID_LEN-1] = 0;
```
iabdalkader, line 394:
```suggestion
            scan_result->au8SSID[WINC_MAX_SSID_LEN-1] = 0;
            strncpy((char*) wscan_result.ssid, (const char *) scan_result->au8SSID, WINC_MAX_SSID_LEN);
```

iabdalkader (changes requested):

iabdalkader:
> Thanks for doing that testing! Let's see what CI says.

It's all good as expected, I made some changes to winc.c I realized that it won't help to null terminate after the strncpy, now I terminate the src string coming from the driver before the strncpy, will merge after the CI is done.

iabdalkader (approved):

iabdalkader:
Merged, thanks!

PR #1298, opened 2021-05-04
Improved WiFi debugging.

iabdalkader:


kwagyeman:
Please lower the UDP broadcast rate to 1 Hz.

kwagyeman:
Bug: sometimes a command goes missing from the IDE. Needs to be fixed before it's ready.

PR #1345, opened 2021-05-29
Kwabena/special sensor stuff

kwagyeman:
This code allows users to drop frames in the sensor driver and use the image fifo for long term storage of images. Please see this thread: https://forums.openmv.io/t/image-capture-interrupt-handler-registration/5924/16?u=kwagyeman

Basically, if you want frame capture running in the background now always capturing frames while your main code goes and does something else for a very long time this is now possible... and you will not drop any frames and have unstable video.

### DIFF
--- a/src/omv/imlib/framebuffer.h
+++ b/src/omv/imlib/framebuffer.h
@@ -34,6 +34,8 @@ typedef struct framebuffer {
     volatile int32_t tail;
     bool check_head;
     int32_t sampled_head;
+    // Disable autoflush policy on full to prevent stale frames.
+    bool disable_full_flush;
     OMV_ATTR_ALIGNED(uint8_t data[], FRAMEBUFFER_ALIGNMENT);
 } framebuffer_t;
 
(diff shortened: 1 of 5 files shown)


src/omv/imlib/framebuffer.h
```diff
@@ -34,6 +34,8 @@ typedef struct framebuffer {
     volatile int32_t tail;
     bool check_head;
     int32_t sampled_head;
+    // Disable autoflush policy on full to prevent stale frames.
+    bool disable_full_flush;
```
iabdalkader, line 38:
```suggestion
```
This should be in sensor_t, it's not really a framebuffer thing.

iabdalkader (changes requested):

src/omv/imlib/framebuffer.h
iabdalkader: Or check for this flag in framebuffer_flush_buffers() ? but make sure it's initialized to zero.

kwagyeman: It gets zeroed in the main memset of the frame buffer header.

Issue #1350, opened 2021-06-07
set_windowing not work at some parameter


kidswong999:
When it is a certain parameter, set_windowing will fail, causing an error:

RuntimeError: Capture Failed: -2
I used OpenMV4 Plus, firmware 4.0.1.

code:
```
import sensor, image, time

sensor.reset()
sensor.set_pixformat(sensor.GRAYSCALE)
sensor.set_framesize(sensor.VGA)
sensor.set_windowing((230, 45, 94, 85))
sensor.skip_frames(time = 1000)

while(True):
    img = sensor.snapshot()
```

kwagyeman:
This will happen if you ask for an image that’s not a multiple of 16 bytes. This is not an error. However, I think I should add better error messages. Anyway, the resolution you are asking for is not possible to capture.

PR #1352, opened 2021-06-07
Improve snapshot error messages

kwagyeman:
Make error messages for snapshot more human readable.

### DIFF
--- a/src/omv/modules/py_sensor.c
+++ b/src/omv/modules/py_sensor.c
@@ -122,7 +122,7 @@ static mp_obj_t py_sensor_snapshot(uint n_args, const mp_obj_t *args, mp_map_t *
                      MP_ERROR_TEXT("PIXFORMAT not set!"));
     } else if (ret == -2) {
         mp_raise_msg(&mp_type_RuntimeError,
-                     MP_ERROR_TEXT("FRAMESIZE not set, unsupported FRAMESIZE, or unsupported WINDOW!"));
+                     MP_ERROR_TEXT("Invalid Frame Size or Window!"));
     } else if (ret == -3) {
         mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Framebuffer Error!"));
     } else if (ret == -4) {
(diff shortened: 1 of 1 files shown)


iabdalkader:
Okay but might as well use full words (like Pixel format and Frame size etc..)... Also this will not stop questions like #1350 if it just says "Framebuffer Error" still doesn't explain what's wrong, best to add some more sanity checks in py_sensor_set_windowing().

kwagyeman:
Error number -3 shouldn't happen ever. I don't know how to make the message more useful given that.

iabdalkader:
Sorry was looking at another error number, the message is okay, if it's not possible to detect the error in py_set_windowing then I guess it's fine.

kwagyeman:
I don't have all the information until snapshot is called. User could set the pixelformat after setwindowing.

src/omv/modules/py_sensor.c
```diff
@@ -115,10 +115,21 @@ static mp_obj_t py_sensor_snapshot(uint n_args, const mp_obj_t *args, mp_map_t *
...
-    // Note: OV2640 JPEG mode can __fatal_error().
     int ret = sensor.snapshot(&sensor, (image_t *) py_image_cobj(image), 0);
 
-    if (ret < 0) {
+    if (ret == -1) {
+        mp_raise_msg(&mp_type_RuntimeError,
+                     MP_ERROR_TEXT("PIXFORMAT not set!"));
+    } else if (ret == -2) {
+        mp_raise_msg(&mp_type_RuntimeError,
+                     MP_ERROR_TEXT("FRAMESIZE not set, unsupported FRAMESIZE, or unsupported WINDOW!"));
```
iabdalkader, line 125:
```suggestion
                     MP_ERROR_TEXT("Invalid Frame Size or Window!"));
```

src/omv/modules/py_sensor.c
```diff
@@ -115,10 +115,21 @@ static mp_obj_t py_sensor_snapshot(uint n_args, const mp_obj_t *args, mp_map_t *
 #endif // MICROPY_PY_IMU
 
     mp_obj_t image = py_image(0, 0, 0, 0);
-    // Note: OV2640 JPEG mode can __fatal_error().
     int ret = sensor.snapshot(&sensor, (image_t *) py_image_cobj(image), 0);
 
-    if (ret < 0) {
+    if (ret == -1) {
+        mp_raise_msg(&mp_type_RuntimeError,
+                     MP_ERROR_TEXT("PIXFORMAT not set!"));
```
iabdalkader, line 122: @kwagyeman I forgot to fix this, if you want to send another PR that's fine.

PR #1353, opened 2021-06-07
Add disable full flush logic

kwagyeman:
Adds the ability for the user to turn off automatic flushing of the frame buffer when it overflows. This allows the user to use the frame buffer as a long term storage fifo.

### DIFF
--- a/src/omv/common/sensor.h
+++ b/src/omv/common/sensor.h
@@ -164,6 +164,7 @@ typedef struct _sensor {
     uint16_t gs_bpp;            // Grayscale bytes per pixel.
     uint32_t hw_flags;          // Hardware flags (clock polarities/hw capabilities)
     const uint16_t *color_palette;    // Color palette used for color lookup.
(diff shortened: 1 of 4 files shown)


src/omv/imlib/framebuffer.h
```diff
@@ -34,6 +34,8 @@ typedef struct framebuffer {
     volatile int32_t tail;
     bool check_head;
     int32_t sampled_head;
+    // Disable autoflush policy on full to prevent stale frames.
+    bool disable_full_flush;
```
iabdalkader, line 38:
```suggestion
    bool disable_flush;
```
This should be in sensor_t.

iabdalkader (changes requested):

src/omv/imlib/framebuffer.h
kwagyeman: It's really a frame buffer policy option.

But okay. Note, the whole frame buffer struct and sensor struct could be one.

kwagyeman:
Updated

PR #1356, opened 2021-06-08
Add working frame rate control for all sensors to STM32

kwagyeman:
You just need to do set_framerate(fps) and the camera will run at a frame rate that's below or equal to the number passed. If you pass 0 then the camera goes back to the maximum fps. The code does a fairly good job of going exactly at the fps you want.

Only an implementation is available for the STM32. The code is smart and doesn't do work after it decides to drop a frame. Processor time/MDMA time is not wasted moving an image that will be tossed. Power is still wasted however transferring the image from the DCMI hardware through DMA2 to SRAM. But, this should be minimal.

Finally, this code works in conjunction with lowering the sensor frame rate. So, for each sensor you can run it slower too and you will get the correct result. Tested this on the Portenta.

### DIFF
--- a/src/omv/common/sensor.h
+++ b/src/omv/common/sensor.h
@@ -175,6 +175,8 @@ typedef struct _sensor {
     pixformat_t pixformat;      // Pixel format
     framesize_t framesize;      // Frame size
     int framerate;              // Frame rate
+    uint32_t last_frame_ms;     // Last sampled frame timestamp in milliseconds.
+    bool last_frame_ms_valid;   // Last sampled frame timestamp in milliseconds valid.
     gainceiling_t gainceiling;  // AGC gainceiling
     bool hmirror;               // Horizontal Mirror
     bool vflip;                 // Vertical Flip
--- a/src/omv/ports/stm32/sensor.c
+++ b/src/omv/ports/stm32/sensor.c
@@ -50,6 +50,8 @@ static DCMI_HandleTypeDef DCMIHandle = {.Instance = DCMI};
 static MDMA_HandleTypeDef DCMI_MDMA_Handle0 = {.Instance = MDMA_Channel0};
 static MDMA_HandleTypeDef DCMI_MDMA_Handle1 = {.Instance = MDMA_Channel1};
 #endif
+static bool first_line = false;
+static bool drop_frame = false;
 
 extern uint8_t _line_buf;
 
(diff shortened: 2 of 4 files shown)


src/omv/ports/stm32/sensor.c
```diff
@@ -50,6 +50,8 @@ static DCMI_HandleTypeDef DCMIHandle = {.Instance = DCMI};
 static MDMA_HandleTypeDef DCMI_MDMA_Handle0 = {.Instance = MDMA_Channel0};
 static MDMA_HandleTypeDef DCMI_MDMA_Handle1 = {.Instance = MDMA_Channel1};
 #endif
+static bool DCMI_DMAConvCpltUser_first_line = false;
+static bool DCMI_DMAConvCpltUser_drop_frame = false;
```
iabdalkader, line 54:
```suggestion
static bool first_line = false;
static bool drop_frame = false;
```

src/omv/modules/py_sensor.c
```diff
@@ -336,6 +336,10 @@ static mp_obj_t py_sensor_set_windowing(uint n_args, const mp_obj_t *args)
 
 static mp_obj_t py_sensor_get_windowing()
 {
+    if (sensor.framesize == FRAMESIZE_INVALID) {
+        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Frame size not set yet!"));
+    }
+
```
iabdalkader, line 342:
```suggestion
```
Doesn't look related, if not please send in another PR.

src/omv/common/sensor.h
```diff
@@ -175,6 +175,8 @@ typedef struct _sensor {
     pixformat_t pixformat;      // Pixel format
     framesize_t framesize;      // Frame size
     int framerate;              // Frame rate
+    uint32_t last_frame_ms;     // Last sampled frame timestamp in milliseconds.
+    bool last_frame_ms_valid;   // Last sampled frame timestamp in milliseconds valid.
```
iabdalkader, line 179:
```suggestion
```
Why not check if last_frame_ms is 0 ?

iabdalkader (changes requested):

src/omv/common/sensor.h
kwagyeman: Wanted to make it as robust as possible. 0 is a valid time.

src/omv/ports/stm32/sensor.c
kwagyeman: Done

src/omv/modules/py_sensor.c
kwagyeman: Moved to other PR.

kwagyeman:
Updated

PR #1360, opened 2021-06-11
Reorganize the ISC SPI related handles/macros.

Lak4CYUT:
The SPI bus on image sensor connector is not design for Lepton sensor only.
Thereforce, change the related functions and macros name from LEPTON_SPI_*
to ISC_SPI_* and move interrupt callback to sensor.c.


src/omv/ports/stm32/sensor.c
```diff
@@ -50,6 +50,13 @@ static DCMI_HandleTypeDef DCMIHandle = {.Instance = DCMI};
 static MDMA_HandleTypeDef DCMI_MDMA_Handle0 = {.Instance = MDMA_Channel0};
 static MDMA_HandleTypeDef DCMI_MDMA_Handle1 = {.Instance = MDMA_Channel1};
 #endif
+// SPI on image sensor connector.
+SPI_HandleTypeDef ISC_SPIHandle = {
+    #ifdef ISC_SPI
+    .Instance = ISC_SPI
+    #endif // ISC_SPI
+};
+
```
iabdalkader, line 59:
```suggestion
#ifdef ISC_SPI
SPI_HandleTypeDef ISC_SPIHandle = {.Instance = ISC_SPI};
#endif // ISC_SPI

```

iabdalkader (changes requested):

iabdalkader:
@Lak4CYUT Looks good! Just need the above small change, the handle shouldn't be declared if nothing is using it.

Lak4CYUT:
> @Lak4CYUT Looks good! Just need the above small change, the handle shouldn't be declared if nothing is using it.

Ya, it is. But if we do so, every driver which rely on ISC_SPI must be check both #ifdef ISC_SPI and OMV_ENABLE_XXX.
Because it may cause a compiler time error.

iabdalkader:
> Ya, it is. But if we do so, every driver which rely on ISC_SPI must be check both #ifdef ISC_SPI and OMV_ENABLE_XXX.
> Because it may cause a compiler time error, are you sure u want to do it?

Yes it's okay, and you don't need to for ISC_SPI anywhere else, if you enable an SPI sensor without defining ISC_SPI it should cause an error so you know that something is wrong.

iabdalkader:
@Lak4CYUT merged, thanks! You should send the PR that adds the driver next (all the Makefiles changes, all omv/drivers/pixart/* and omv/sensors/paj6100*), followed by a PR with changes to sensor.h and sensor.c. Finally a PR to enable it in omv_boardconfig.h

Lak4CYUT:
> Yes it's okay, and you don't need to check for ISC_SPI anywhere else, if you enable an SPI sensor without defining ISC_SPI it should cause an error so you know that something is wrong. Currently it will define an empty handle (with a NULL instance) and no IRQ/DMA handlers which will cause faults if you run it.

Wow, agree that. It is a good point.
I will send the next PR later, see you soon.

PR #1370, opened 2021-06-19
Finish MT9M114 driver

kwagyeman:
Everything works except:
* Need to update bayer image processing code to handle all possible pattern options versus 1.
* Set_contrast appears to do nothing. Verified it's setting the right register however per the datasheet.
* Set_windowing doesn't really work. You can get cropping and a frame rate boost if you shrink the vertical res only. Need to work on the code more.
* Bigger issue with the above... changing the resolution/windowing wipes exposure settings making the 100 FPS example not work. Not really sure if this can be fixed given the way the sensor works.


src/omv/common/sensor.h
```diff
@@ -38,12 +38,12 @@
 #define OV7725_ID           (0x77)
 #define OV9650_ID           (0x96)
 #define MT9V034_ID          (0x13)
-#define MT9M114_ID          (0x81)
```
iabdalkader, line 41: Let's keep it here to match the order of init.

iabdalkader:
>     * Bigger issue with the above... changing the resolution/windowing wipes exposure settings making the 100 FPS example not work. Not really sure if this can be fixed given the way the sensor works.

That's easy we can maintain the reg values in some structure, and write again after res changes or whenever it's needed.

src/omv/common/sensor.h
kwagyeman: Fixed

kwagyeman:
It's ready

PR #1383, opened 2021-06-29
Move common image sensor driver code to a common file.

iabdalkader:
* Provide image sensor driver utility functions that can be used by all ports.
* Provide default (weak) functions that can be replaced by port-specific drivers.


kwagyeman (approved):
I think you can move the set xclk stuff into the sensor common folder too.

Anyway, this is great!

iabdalkader:
> I think you can move the set xclk stuff into the sensor common folder too.

No this one is too hw/port-specific.

iabdalkader:
Tested on OMV4, PICO, NANO33

PR #1401, opened 2021-07-19
RP2: Optimize bus scanning.

iabdalkader:
* For sensor-less boards, lower the scan address space.

### DIFF
--- a/src/omv/ports/rp2/cambus.c
+++ b/src/omv/ports/rp2/cambus.c
@@ -9,7 +9,7 @@
 #include "cambus.h"
 
 #define I2C_TIMEOUT             (100*1000)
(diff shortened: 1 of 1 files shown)


src/omv/ports/rp2/cambus.c
```diff
@@ -66,7 +66,7 @@ int cambus_deinit(cambus_t *bus)
 
 int cambus_scan(cambus_t *bus)
 {
-    for (uint8_t addr=0x09, rxdata; addr<=0x77; addr++) {
+    for (uint8_t addr=0x20, rxdata; addr<=0x48; addr++) {
```
iabdalkader, line 69: @kwagyeman Should probably do the same for STM32. There's no need to scan lower and higher addresses.

PR #1422, opened 2021-08-04
Add IMU example

sebromero:
This PR adds an example to use the IMU on the RP2040


iabdalkader:
Thanks but this one is already included in  #1423

PR #1432, opened 2021-08-16
speed up software JPEG encoding by about 10% with improved bit handling

bitbank2:
The software JPEG encoder now has more efficient code to write the variable length codes to the output file.

### DIFF
--- a/src/omv/imlib/jpeg.c
+++ b/src/omv/imlib/jpeg.c
@@ -1010,6 +1010,51 @@ static const uint16_t UVAC_HT[256][2] = {
     {0xFFFC, 0x0010},{0xFFFD, 0x0010},{0xFFFE, 0x0010},{0x0000, 0x0000},{0x0000, 0x0000},{0x0000, 0x0000},{0x0000, 0x0000},{0x0000, 0x0000},
 };
 
+// Macro to write variable length codes to the output stream more efficiently
+#define STORECODE(pOut, iLen, ulCode, ulAcc, iNewLen) \
+        if (iLen+iNewLen > 32) { while (iLen >= 8) \
+                        {unsigned char c = (unsigned char)(ulAcc >> 24); *pOut++ = c; \
+        if (c == 0xff) { *pOut++ = 0;} ulAcc <<= 8; iLen -= 8; }} \
+        iLen += iNewLen; ulAcc |= (ulCode << (32-iLen));
+
+//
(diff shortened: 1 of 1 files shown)


iabdalkader:
@bitbank2 I'm sorry I completely missed this PR, will review and merge soon.

iabdalkader:
> How would you like it to handle the error case of running out of memory? Just return without finishing?

Hi, it should set the flag and stop writing to the buffer, just like the current code, when it returns the overflow will be detected.

bitbank2:
ok, I just pushed a fix which will allow the encode to finish successfully.

src/omv/imlib/jpeg.c
```diff
@@ -1173,20 +1223,30 @@ static int jpeg_processDU(jpeg_buf_t *jpeg_buf, int8_t *CDU, float *fdtbl, int D
         }
     }
 
+    jpeg_check_highwater(jpeg_buf); // check if we're getting close to the end of the buffer
+    // Use local vars to speed up buffer access
+    // and a macro (STORECODE) to manipulate the local vars
+    uint8_t *pOut, iBitCount; // output pointer and bit count
+    uint32_t ulBits; // accumulated bits
+    pOut = &jpeg_buf->buf[jpeg_buf->idx];
```
iabdalkader, line 1231: @bitbank2 Assuming the above call in line 1226 to `jpeg_check_highwater` overflows and sets the `jpeg_buf->length` to 0,  pOut is set using the idx not length, and then the following code will overflow right ?

Perhaps just instead of resetting the length or idx, simply set the overflow flag and return true ?  For example:

```C
static bool jpeg_check_highwater(jpeg_buf_t *jpeg_buf)
{   
    if ((jpeg_buf->idx+1) >= jpeg_buf->length - 256) {
        if (jpeg_buf->realloc == false) {
            // Can't realloc buffer
            jpeg_buf->overflow = true;
        } else {
            jpeg_buf->length += 1024;
           jpeg_buf->buf = xrealloc(jpeg_buf->buf, jpeg_buf->length);
       }
    }
    return jpeg_buf->overflow;
} /* jpeg_check_highwater() */
```

Then:

```C
if (jpeg_check_highwater(jpeg_buf) == true) {
    return;
}
```

bitbank2: Thanks for catching that mistake. I cleaned up the function to behave correctly.

PR #1436, opened 2021-08-24
Fix set_framerate.

iabdalkader:


kwagyeman (changes requested):
No, this is not right. If the sensor control fails the software layer can continue to drop frames to set the frame rate.

kwagyeman:
Wait, I see. Um, just don't call set frame rate part if null.

kwagyeman:
`if (sensor.set_framerate != NULL) {` -> if (sensor.set_framerate == NULL) {

kwagyeman (approved):

kwagyeman:
It's correct.

PR #1445, opened 2021-09-01
Add full debayering support

kwagyeman:
* Adds support to debayer all possible bayer patterns (with SIMD support).
* Bayer pattern output by cameras can now be specified. Defaults to BGGR.
* Sets the MT9M114 bayer pattern default to match the camera sensor output.
* Crop will automatically adjust the bayer pattern as it moves.

Notes:

It's now possible to cleanup some of the older image sensor set_framesize code which adjusts the bayer pattern readout pixel start to align to bggr. This is no longer needed. These fixes can be done in another PR.

In the future if we want to support sensors with clear bayer patterns this can be done easily by just extending the possible bayer pattern enums.

### DIFF
--- a/src/omv/imlib/bayer.c
+++ b/src/omv/imlib/bayer.c
@@ -10,7 +10,7 @@
  */
 #include "imlib.h"
 
-void imlib_debayer_line_to_binary(int x_start, int x_end, int y_row, uint32_t *dst_row_ptr, image_t *src)
+void imlib_debayer_line(int x_start, int x_end, int y_row, void *dst_row_ptr, pixformat_t pixfmt, image_t *src)
 {
     int src_w = src->w, w_limit = src_w - 1, w_limit_m_1 = w_limit - 1;
     int src_h = src->h, h_limit = src_h - 1, h_limit_m_1 = h_limit - 1;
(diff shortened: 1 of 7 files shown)


src/omv/common/sensor.h
```diff
@@ -193,6 +193,7 @@ typedef struct _sensor {
     uint16_t chip_id_w;         // Sensor ID 16 bits.
     };
     uint8_t  slv_addr;          // Sensor I2C slave address.
+    int bayer;                  // CFA bayer pattern.
```
iabdalkader, line 196:
```suggestion
    uint32_t bayer_pattern;                  // CFA bayer pattern.
```

src/omv/imlib/bayer.c
```diff
@@ -10,7 +10,7 @@
  */
 #include "imlib.h"
 
-void imlib_debayer_line_to_binary(int x_start, int x_end, int y_row, uint32_t *dst_row_ptr, image_t *src)
```
iabdalkader, line 13: IMHO the multiple small functions were much more readable and easier to add to than 1 monolithic function. At least keep the smaller functions internally (inline'd static) and call them from debayer_line/image, and add a function/macro to clip the row pointers as needed.

src/omv/imlib/bayer.c
```diff
@@ -94,860 +94,388 @@ void imlib_debayer_line_to_binary(int x_start, int x_end, int y_row, uint32_t *d
...
-            int row_1g = __UHADD8(row_bgbg_1, __PKHTB(row_bgbg_1, row_bgbg_1, 16));
+            int r_pixels_0, g_pixels_0, b_pixels_0;
 
-            int r_pixels_0 = __UXTB16(__UHADD8(row_01, __PKHTB(row_01, row_01, 16)));
-            int g_pixels_0 = __UXTB16(__UHADD8(row_1g, __PKHTB(row_1g, row_01, 8)));
-            int b_pixels_0 = __UXTB16_RORn(__UHADD8(row_bgbg_1, __PKHBT(row_bgbg_1, row_bgbg_1, 16)), 8);
-            #else
+            switch (src->bpp) {
+                case IMAGE_BPP_BAYER_BGGR: {
+                    #if defined(MCU_SERIES_F4) || defined(MCU_SERIES_F7) || defined(MCU_SERIES_H7)
```
iabdalkader, line 101:
```suggestion
#if defined(ARM_MATH_DSP)
```

You should double check this, but the right macro to test for before using these functions is `#if (__ARM_FEATURE_DSP == 1) see cmsis_gcc.h. The MCU_SERIES_xx macros are for specific to stm32, ex: if NXP is supported this won't work, especially in common/shared code.

iabdalkader (changes requested):

iabdalkader:
Actually I think maybe it's time to start fixing the bpp/format issues. We should add a format field to image_t and use bpp exclusively for bytes per pixel, and fix any code that uses BPP to check the image format.

kwagyeman:
I'm fine with that. But, it should be in another PR.

src/omv/imlib/bayer.c
kwagyeman: You are right. Should replace this in other places too.

src/omv/imlib/bayer.c
kwagyeman: The amount of code generated would be enormous. I merged it all into 1 method so we wouldn't have 4K lines of code here.

iabdalkader:
> I'm fine with that. But, it should be in another PR.

I agree, but I think it should be done first then this PR rebased accordingly...I'm looking at it and I realized we already have the PIXFORMAT_X enum it's used for sensors but we could reuse it for image formats too. Let's discuss... could reuse it like this:

```diff
diff --git a/src/omv/imlib/imlib.h b/src/omv/imlib/imlib.h
index 52f14501..bef2cccc 100644
--- a/src/omv/imlib/imlib.h
+++ b/src/omv/imlib/imlib.h
@@ -351,15 +351,24 @@ typedef enum image_bpp
     IMAGE_BPP_BINARY,       // BPP = 0
     IMAGE_BPP_GRAYSCALE,    // BPP = 1
     IMAGE_BPP_RGB565,       // BPP = 2
-    IMAGE_BPP_BAYER,        // BPP = 3
+    IMAGE_BPP_BAYER,        // BPP = 3 // TODO: define as 1 bpp later.
     IMAGE_BPP_JPEG          // BPP > 3
 }
 image_bpp_t;
 
+// Use image format high byte for format flags:
+ // Could expand in the future, for example PIXFORMAT_GRAYSCALE_16BITS.
+#define PIXFORMAT_MASK          (0x00FFFFFF)
+#define PIXFORMAT_BAYER_BGGR    ((1<<0) << 24)
+#define PIXFORMAT_BAYER_GBRG    ((1<<2) << 24)
+#define PIXFORMAT_BAYER_GRBG    ((1<<3) << 24)
+#define PIXFORMAT_BAYER_RGGB    ((1<<4) <<24)
+
 typedef struct image {
     int w;
     int h;
     int bpp;
+    uint32_t fmt;
     union {
         uint8_t *pixels;
         uint8_t *data;
@@ -593,7 +602,7 @@ extern const int kernel_high_pass_3[9];
 
 #define IM_IS_BAYER(img) \
     ({ __typeof__ (img) _img = (img); \
-       _img->bpp == 3; })
+       (_img->fmt & PIXFORMAT_MASK) == PIXFORMAT_BAYER; })
 
 #define IM_IS_JPEG(img) \
     ({ __typeof__ (img) _img = (img); \
```


In snapshot:
```c
    image->bpp    = MAIN_FB()->bpp;
    image->fmt    = sensor->pixformat | sensor->pixformat_flags;
```

In switch statements:
```C
switch (img->fmt) {
    case PIXFORMAT_GRAYSCALE:
        break;
    case (PIXFORMAT_GRAYSCALE | PIXFORMAT_GRAYSCALE_16BITS ):
        break;
    case (PIXFORMAT_BAYER | PIXFORMAT_BAYER_BGGR):
        break;
    case (PIXFORMAT_BAYER | PIXFORMAT_BAYER_GBRG):
        break;
    ....
}
```

Thoughts ?

src/omv/imlib/bayer.c
iabdalkader: Not looking for speed ups, but it will be harder to add to/fix in the future, your choice anyway.

kwagyeman:
I kind like how bpp does multiple things though. What if bpp was split into 24-bits for the jpeg image and 8-bits for the format?

iabdalkader:
> I kind like how bpp does multiple things though. What if bpp was split into 24-bits for the jpeg image and 8-bits for the format?

I thought about this initially, but I was trying to avoid mixing things again, plus what's the point ? We'll need the same amount of code changes, plus macros to set/extract the format and bpp, for example:

```C
switch (IMLIB_GET_PIXFORMAT(img)) {
....
img->bpp = IMLIB_GET_BPP(src);
```

Vs just using fmt and bpp which is much cleaner:
```C
switch(img->fmt) {
...
img->bpp = src->bpp;
```

It's possible though, but wither way we still need BAYER flags for "subformats" because sensors expect PIXFORMAT_BAYER so image->bpp can't be set to sensor.pixformat directly...

Alternatively to save some space (which is insignificant but still) we could use bitfields:
```C
typedef struct image {
    uint32_t w;
    uint32_t h;
    uint32_t fmt:8;
    uint32_t bpp:24;
    union {
        uint8_t *pixels;
        uint8_t *data;
    };
} image_t;
```

kwagyeman:
Mainly I just want to keep the ImageWriter format samish. So, I'd like to avoid too many changes. 

Note, I have plans for:

YUYV (16-bits per pixel)
YVYU (16-bits per pixel - you need this additional one to handle crops of 1 pixel)
ARGB (32-bits) - for PNG support for loading graphics like logo overlays

iabdalkader:
> Mainly I just want to keep the ImageWriter format samish.

This shouldn't change at all (the imageio format is w, h, bpp, size followed by image data) it should be the same regardless of how bpp is stored internally. 

> Mainly I just want to keep the ImageWriter format samish. So, I'd like to avoid too many changes.

Either way will require a lot of changes, but if bpp is shared (fmt + bpp) it will need even more changes because wherever bpp is used will need to be updated to extract the bpp/size 24-bit part, but if fmt is introduced, bpp will be the same.

> ARGB (32-bits) - for PNG support for loading graphics like logo overlays

All the more reason to fix this, because JPEG is tested with bpp >= 4

iabdalkader:
How about this, we can encode everything in one field (fmt), but we need to fix all the code that uses the current bpp for bytes per pixel and pixel format ID:

```C
#define PIXFORMAT_FLAGS_MF      (1 << 31) // Mutable format.
#define PIXFORMAT_FLAGS_CF      (1 << 30) // Color format.
#define PIXFORMAT_FLAGS_CM      (1 << 29) // Compressed image.
#define PIXFORMAT_FLAGS_RF      (1 << 28) // Raw format.
// Each pixel format encodes flags pixel format, and bpp/size in 32-bits.
// 31 30 29 28  27......24 23..........0
// MF CF CM RF <PIXFORMAT> <BPP or size>
typedef enum {
    PIXFORMAT_INVALID       = (                                                                         0),
    PIXFORMAT_BINARY        = (PIXFORMAT_FLAGS_MF                                           | (1 << 24) | 1),   // 1BPP/BINARY
    PIXFORMAT_GRAYSCALE     = (PIXFORMAT_FLAGS_MF                                           | (2 << 24) | 1),   // 1BPP/GRAYSCALE
    PIXFORMAT_RGB565        = (PIXFORMAT_FLAGS_MF | PIXFORMAT_FLAGS_CF                      | (3 << 24) | 2),   // 2BPP/RGB565
    PIXFORMAT_YUV422,       = (PIXFORMAT_FLAGS_MF | PIXFORMAT_FLAGS_CF                      | (4 << 24) | 2),   // 2BPP/YUV422
    PIXFORMAT_BAYER_BGGR,   = (PIXFORMAT_FLAGS_MF | PIXFORMAT_FLAGS_CF | PIXFORMAT_FLAGS_RF | (5 << 24) | 1),   // 1BPP/RAW
    PIXFORMAT_BAYER_GBRG,   = (PIXFORMAT_FLAGS_MF | PIXFORMAT_FLAGS_CF | PIXFORMAT_FLAGS_RF | (6 << 24) | 1),   // 1BPP/RAW
    PIXFORMAT_BAYER_GRBG,   = (PIXFORMAT_FLAGS_MF | PIXFORMAT_FLAGS_CF | PIXFORMAT_FLAGS_RF | (7 << 24) | 1),   // 1BPP/RAW
    PIXFORMAT_BAYER_RGGB,   = (PIXFORMAT_FLAGS_MF | PIXFORMAT_FLAGS_CF | PIXFORMAT_FLAGS_RF | (8 << 24) | 1),   // 1BPP/RAW
    PIXFORMAT_JPEG          = (PIXFORMAT_FLAGS_CM                                           | (9 << 24) | 0),   // JPEG/COMPRESSED
} pixformat_t;

```
From left to right: (color, mutable, compressed, raw flags in 4 bits), unique ID for the format (4 or 5 bits) followed by BPP (or JPEG size) in 24 bits.

Note the flags are there to replace these ever expanding macros and to make tests simpler:
```C
#define IMAGE_IS_MUTABLE(image) \
({ \
    __typeof__ (image) _image = (image); \
    IMAGE_PIXFORMAT_EQUAL(_image->fmt, PIXFORMAT_BINARY) || \
    IMAGE_PIXFORMAT_EQUAL(_image->fmt, PIXFORMAT_GRAYSCALE) || \
    IMAGE_PIXFORMAT_EQUAL(_image->fmt, PIXFORMAT_RGB565); \
})
```

With 
```C
#define IMAGE_IS_MUTABLE(image) (image->fmt & PIXFORMAT_FLAGS_MF)
#define IMAGE_IS_BAYER(image) (image->fmt & PIXFORMAT_FLAGS_RF)
#define IMAGE_IS_MUTABLE_BAYER(image)  (image->fmt & (PIXFORMAT_FLAGS_MF | PIXFORMAT_FLAGS_RF) )
```

kwagyeman:
That's pretty sensible actually. If you think about it these macros take up a lot of flash. An & with a bitmask is a very cheap operation on ARM.

Anyway, what fixes do you want to unblock merging this? We can do the pixformat update after this PR is merged.

...

Regarding the RAW/Compressed format. Not sure if that's exactly needed. But, okay.

iabdalkader:
> Anyway, what fixes do you want to unblock merging this? We can do the pixformat update after this PR is merged. The ones above where you asked me to change the code? Can do very soon.

Actually I'd like to send a couple of PRs first then get this PR rebased after, because a lot of the changes won't be needed.

> Regarding the RAW/Compressed format. Not sure if that's exactly needed. But, okay.

Not likely, it's still a work in progress I just sent a draft still thinking about some details. The Compressed flag was meant to be a quick check to see if the image is JPEG (or PNG if it's ever supported), but the Mutable flag can be used for the same purpose, so will I probably remove it.

kwagyeman:
Larry will be adding PNG support soon. So, that's coming (PNG files are what makes the ARGB output)

iabdalkader:
@kwagyeman one more PR to go, but I noticed that bpp is used sometimes for things like:

```C
        uint32_t avail = fb_avail();
        uint32_t space = src_w_mcus_bytes_2 + FB_ALLOC_PADDING;
        dst->bpp = avail - space;
        dst->data = fb_alloc(dst->bpp, FB_ALLOC_PREFER_SIZE | FB_ALLOC_CACHE_ALIGN);
```
24 bits bpp might cause issues for code like that, I will still encode bpp in the pixformat, but I think there should a separate `uint32_t size` field for JPEG.

kwagyeman:
Ah, you are right...

Mmm, well, the jpeg is not going to be larger than 16MB. But, that would break instantly on this PR as it would 24-bit mod the value and roll over causing problems. So, we'd need to fix all places like that to limit the max available to the 2^24-1.

iabdalkader:
>  So, we'd need to fix all places like that to limit the max available to the 2^24-1.

We could just add a 32 bit size field to image_t for JPEG/PNG size in general,  and for places like the example above to be safe. If I'm going through the all the code that does that to fix it, might as well replace bpp/size.... Plus I have a really cool idea for those extra 24-bits in the format which works with everything else.

kwagyeman:
Okay, sounds good.

iabdalkader:
This is going to be a big change, so I figured I should share the basic idea first, so it's clear: https://github.com/openmv/openmv/commit/577fced2b481eb99b2a3829e622a09bc5f568711

Note the change in image_t, the union of bit-field and int in image_t allow setting all the encoded info using the PIXFORMAT_ constants like this:
```C
// All format bits, bpp etc.. are encoded in the constant PIXFORMAT_JPEG
image->pixfmt = PIXFORMAT_JPEG;

// Now all the bits are set, test compression/mutable:
if (image->is_compressed) {
} else if (image->is_mutable) {
}
printf("bpp %u\n", image->bpp);
//etc...
```

The following macros now look like this (they won't be needed but anyway):
```C
#define IMAGE_IS_MUTABLE(image)             (image->is_mutable && !image->is_bayer)
#define IMAGE_IS_COLOR(image)               (image->is_color)
#define IMAGE_IS_MUTABLE_BAYER(image)       (image->is_mutable)
#define IMAGE_IS_MUTABLE_BAYER_JPEG(image)  (image->is_mutable || image->is_compressed)
```

You can also simply just set a sub-format when needed, like for example in the bayer case in snapshot:
```C
image->pixfmt = sensor->pixformat; // assume it's bayer any sub format
image->subfmt_id = sensor->bayer; // sets the bayer sub-format that was set in sensor drivers.
```

Now you can just switch on /test for that format like this:
```C
switch (image->pixfmt):
       case PIXFORMAT_BAYER_GBRG:
       case PIXFORMAT_BAYER_RGB565:
       ....
```

Or if you only care that it's a bayer format (or any RGB format) you can do this:
```C
switch (image->pixfmt_id):
     case PIXFORMAT_ID_BAYER:
     case PIXFORMAT_ID_RGB:
     case PIXFORMAT_ID_YUV:
}
```

iabdalkader:
On second thought I'll leave the 24-bits bpp for now to reduce the change set and size can be added later... The main issue that will be fixed is making sure bpp is not used to test for pixel formats, and I'll also make sure it doesn't overflow when used for jpeg size.

kwagyeman:
Cool

src/omv/imlib/imlib.h
```diff
@@ -385,7 +417,10 @@ bool image_get_mask_pixel(image_t *ptr, int x, int y);
...
     (_image->bpp == IMAGE_BPP_GRAYSCALE) || \
     (_image->bpp == IMAGE_BPP_RGB565) || \
-    (_image->bpp == IMAGE_BPP_BAYER); \
+    (_image->bpp == IMAGE_BPP_BAYER_BGGR) || \
+    (_image->bpp == IMAGE_BPP_BAYER_GBRG) || \
+    (_image->bpp == IMAGE_BPP_BAYER_GRBG) || \
+    (_image->bpp == IMAGE_BPP_BAYER_RGGB); \
 })
 
 #define IMAGE_IS_MUTABLE_BAYER_JPEG(image) \
```
iabdalkader, line 426: @kwagyeman What's this macro for ? it's basically always true.

kwagyeman: Not for jpeg. But, it depends on where it's used.

iabdalkader: https://github.com/openmv/openmv/blob/master/src/omv/imlib/imlib.h#L391

It literally tests all the formats.

iabdalkader (approved):

PR #1455, opened 2021-09-23
Add basic frogeye2020 driver

kwagyeman:
Adds an event sensor driver for the frogeye2020 with examples.

### DIFF
--- a/scripts/examples/OpenMV/37-Event-Cameras/frogeye2020.py
+++ b/scripts/examples/OpenMV/37-Event-Cameras/frogeye2020.py
@@ -1,0 +1,39 @@
+# This example shows off using the frogeye2020 event camera.
+#
+# The frogeye2020 is a 320x240 event camera. There are two bits per pixel which show no motion,
+# motion in one direction, or motion in another direction. The sensor runs at 50 FPS.
+
+import sensor, image, time
+
+sensor.reset()
+sensor.set_pixformat(sensor.GRAYSCALE)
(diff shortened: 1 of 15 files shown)


kwagyeman:
`# This example shows off using the frogeye2020 event camera with tracking.
#
# The frogeye2020 is a 320x240 event camera. There are two bits per pixel which show no motion,
# motion in one direction, or motion in another direction. The sensor runs at 50 FPS.

import sensor, image, time

sensor.reset()

palette = image.Image(1, 256, sensor.RGB565)

for i in range(64):
    palette.set_pixel(0, i, (0, 0, 0))

for i in range(64, 128):
    palette.set_pixel(0, i, (255, 0, 0))

for i in range(128, 192):
    palette.set_pixel(0, i, (0, 0, 255))

for i in range(192, 256):
    palette.set_pixel(0, i, (0, 255, 0))

clock = time.clock()

while(True):
    clock.tick()

    img = sensor.snapshot()
    # Handle sensor rotation.
    img.assign(hmirror=True, vflip=True)
    # Make pretty.
    img.to_rainbow(color_palette=palette)
    # Cleanup noise.
    img.erode(1)

    blobs = img.find_blobs([(0, 0)], invert=True,
        pixels_threshold=10, area_threshold=10, merge=False)

    for blob in blobs:
        img.draw_rectangle(blob.rect(), color=(0, 255, 0))

    print(clock.fps())`

PR #1460, opened 2021-10-17
Kwabena/fix network usage in scripts

kwagyeman:
* Updates all the streaming scripts to use any network interface instead of just WINC.
* Renames interfaces too.
* Fixes a bug with IS_JPEG using the wrong value to compare with.


iabdalkader:
> Fixes a bug with IS_JPEG using the wrong value to compare with

Please don't mix fixes in one PR again.

PR #1462, opened 2021-10-25
Add Objection Detection Support

kwagyeman:
Support for segmentation networks has been added. You can now segment images using tensorflow into grayscale images that show a heatmap per class you are looking for. The segment() method will return a list of images of these grayscale heat maps.

detect() will then do all the above but internally run find_blobs() on the heat maps to return instead a list of lists, where each sub list is the blobs detected per class.

EdgeImpulse will have support for running segmentation networks thus enabling object detection and localization on Cortex-M processors.

* PYTF now uses the optimal amount of memory for buffers versus all - buffers are placed in SRAM if they fit producing a massive speed boost.
* Custom scaled/offset outputs now work.
* Updated to the latest tensorflow library.
* You have access to all input/output model parameters.
* Person detection is now int8 and blazing fast - 20 FPS on the Arduino Portena.
* Added m55 libs (m0plus libs coming soon once EdgeImpulse adds support for them in the tensorflow make file)
* Classify/Segment/Detect work on all image types directly (JPG/BAYER/YUV/RGB565/GRAYSCALE/BINARY)


kwagyeman:
Ibrahim, please feel free to merge squash all these commits into one.

PR #1472, opened 2021-12-12
Unsigned var in count down loop fixed

kwagyeman:
GCC doesn't feature a warning for this... maybe we should just change all loops to like this blog?

https://nachtimwald.com/2019/06/02/unsigned-count-down/

Anyway, this is not the issue that's crashing TensorFlow in the later release. I found this issue already.


iabdalkader:
> maybe we should just change all loops to like this blog?

Most loops already use `int` and count up to < INT_MAX.

iabdalkader:
I just noticed something else just doesn't look right will send a PR.

PR #1474, opened 2021-12-12
Fix TF U8 to S8 conversion.

iabdalkader:
* Both fast (words) and slow (bytes) paths are executed, effectively undoing the conversion.


kwagyeman:
```
for (; size >= 0; size -= 1) {
    model_input_8[size] ^= 0x80;
}
```

At the end cleans up the tail. It's not run on all the bytes.

PR #1481, opened 2021-12-31
Make default disk files different per camera

kwagyeman:
Files created by main.c on the first disk boot are not specific per board and are actually editable as files now.

I've wanted to add this for a while so that the OpenMV Pure Thermal's main.py boots the board with all it's glory bey default. You'll notice the main.py for it is quite a bit more complex.

### DIFF
--- a/src/omv/Makefile
+++ b/src/omv/Makefile
@@ -15,6 +15,12 @@ SRCS += $(addprefix alloc/,     \
 	unaligned_memcpy.c          \
    )
 
+SRCS += $(addprefix boards/$(TARGET)/,  \
+	main.py                             \
+	readme.txt                          \
+	selftest.py                         \
+   )
+
 SRCS += $(addprefix common/,    \
 	array.c                     \
 	ff_wrapper.c                \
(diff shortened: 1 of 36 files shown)


iabdalkader:
Hi, this won't work because if there's an SD card, the internal filesystem will Not be mounted and `main.py` will Not executed so for PureThermal the boot code will not run (besides, it should be in `boot.py`).

There's already an infrastructure for what you're trying to do: we can freeze `main.py`, `boot.py` per board with board-specific files, however this wasn't supported in stm32 port, now I fixed in #1482

Now you can override `main.py` or freeze a `boot.py` in `omv/boards/OPENMVPT/boot.py`, enable it in the manifest like this: `freeze ("./", "boot.py")` and it will run from there regardless of the filesystem state.

iabdalkader:
Another issue with this, it generates the following:
```python
import sensor, time, pyb

def test_int_adc():
    adc  = pyb.ADCAll(12)
    # Test VBAT
    vbat = adc.read_core_vbat()
    vbat_diff = abs(vbat-"OMV_CORE_VBAT")   # <------
```

`OMV_CORE_VBAT` is defined in board config files.

PR #1482, opened 2021-12-31
Execute frozen boot scripts first before filesystem.

iabdalkader:
* Run frozen _boot.py if it exists (for early boot stuff).
* Allow freezing main.py and boot.py boot scripts.
* Give frozen boot scripts priority over filesystem boot scripts.
* Add frozen early boot script for OpenMV-PT.


iabdalkader:
@kwagyeman updated.

PR #1484, opened 2022-01-01
Fix person detector model labels

kwagyeman:
Removes the unsure label.


iabdalkader:
Too bad it didn't make it in time for the release.

kwagyeman:
I just locally fixed in in the IDE resources.

PR #1491, opened 2022-01-05
Kwabena/image io

kwagyeman:
This is massive re-write of the ImageIO module with a lot of new features:

- In memory more or less works the same, however, you can now use different image types per frame as long as they fit within the byte size of the frame first allocated.
- On disk works the same, but, now seeking is supported, additionally, you are in read/write mode now always so you can move seek around and read/write frames (that said, this only makes sense with fixed buffer sizes).
- Add methods to get a lot more information about the module.
- Previous versions of ImageIO are now handled correctly. If you open an older file version you can append to it without issue. You can also read from older file versions without issue.

### DIFF
--- a/src/omv/modules/py_imageio.c
+++ b/src/omv/modules/py_imageio.c
@@ -11,7 +11,6 @@
 #include "imlib_config.h"
 #if defined(IMLIB_ENABLE_IMAGE_IO)
 
-#include <stdio.h>
 #include "py/obj.h"
 #include "py/nlr.h"
 #include "py/mphal.h"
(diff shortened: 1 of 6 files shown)


src/omv/modules/py_imageio.c
```diff
@@ -21,131 +20,319 @@
...
+        stream->closed ? "\"true\"" : "\"false\"",
+        stream->count,
+        stream->offset,
+        (stream->type == IMAGE_IO_FILE_STREAM)
+#if defined(IMLIB_ENABLE_IMAGE_FILE_IO)
+            ? stream->version
+#else
+            ? 0
+#endif
+            : 0,
```
iabdalkader, line 110:
```suggestion
#if defined(IMLIB_ENABLE_IMAGE_FILE_IO)
        (stream->type == IMAGE_IO_FILE_STREAM) ? stream->version  : 0,
#else
         0,
#endif
```
More readable.

kwagyeman: Done

PR #1493, opened 2022-01-07
Fix flag spec on timer

kwagyeman:
Processor crashes without this. It tries to call the python object as if it were a C method.

And expiry is set in the insert call. No need to set it.


iabdalkader:
Thanks, the API changed and I can't test this.

PR #1497, opened 2022-01-09
Low resolution linked list mode working

kwagyeman:
This uses linked list mode for MDMA for lower resolutions. It works in all modes, etc.

However, it doesn't actually appear to fix the issue we were having. So, not sure if this makes sense to merge. But, definitely nice to have the PR lying around.

### DIFF
--- a/src/omv/ports/stm32/sensor.c
+++ b/src/omv/ports/stm32/sensor.c
@@ -29,6 +29,7 @@
 // Higher performance complete MDMA offload.
 #if (OMV_ENABLE_SENSOR_MDMA == 1)
 #define OMV_ENABLE_SENSOR_MDMA_TOTAL_OFFLOAD 1
+#define OMV_MDMA_LL_LINES 144
 #endif
 
 sensor_t sensor = {};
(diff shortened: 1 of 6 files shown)


iabdalkader:
Maybe turn it into a semi-general purpose MDMA driver in mdma.c ? Move mdma_config/mdma_memcpy  and this to mdma.c.

src/omv/ports/stm32/sensor.c
```diff
@@ -29,6 +29,7 @@
 // Higher performance complete MDMA offload.
 #if (OMV_ENABLE_SENSOR_MDMA == 1)
 #define OMV_ENABLE_SENSOR_MDMA_TOTAL_OFFLOAD 1
+#define OMV_MDMA_LL_LINES 144
```
iabdalkader, line 32: should be in board config.

src/omv/ports/stm32/sensor.c
```diff
@@ -421,6 +426,18 @@ static void mdma_memcpy(vbuffer_t *buffer, void *dst, void *src, int bpp, bool t
 }
 #endif
 
+#if (OMV_ENABLE_SENSOR_MDMA_TOTAL_OFFLOAD == 1)
```
iabdalkader, line 429: I don't think this is needed anymore, it should by default fully offload to MDMA.

PR #1498, opened 2022-01-09
Make sure cached CPU writes are flushed before returning the buffer

### LINKED ISSUE #1407: The morph function causes many rectangles in the image
The morph function causes many rectangles in the image. It may be an algorithm error, or it may be a transmission error.
Seems to be related to resolution

error

code:
```
import sensor, image, time
kernel_size = 1
kernel = [-1, -1, -1,\
          -1, +8, -1,\
          -1, -1, -1]
sensor.reset()
sensor.set_framesize(sensor.QQQVGA)
sensor.skip_frames(10)
clock = time.clock()
while(True):
    clock.tick()
    img = sensor.snapshot()
    img.morph(kernel_size, kernel)
    print(clock.fps())
```

 - IDE version: 2.6.9
 - Firmware version: 4.0.2
 - Board: OPENMV-H7 Plus, OV5640 sensor
 - OS: Windows

iabdalkader:
@kwagyeman 

* Should fix #1407

### DIFF
--- a/src/omv/imlib/framebuffer.c
+++ b/src/omv/imlib/framebuffer.c
@@ -364,12 +364,18 @@ void framebuffer_auto_adjust_buffers()
 
 void framebuffer_free_current_buffer()
 {
+    vbuffer_t *buffer = framebuffer_get_buffer(framebuffer->head);
+    #ifdef __DCACHE_PRESENT
+    // Make sure all cached CPU writes are flushed before returning the buffer.
+    SCB_InvalidateDCache_by_Addr(buffer->data, framebuffer_get_buffer_size());
+    #endif
+
     // Invalidate frame.
     framebuffer->pixfmt = PIXFORMAT_INVALID;
 
     // Allow frame to be updated in single buffer mode...
     if (framebuffer->n_buffers == 1) {
-        framebuffer_get_buffer(framebuffer->head)->waiting_for_data = true;
+        buffer->waiting_for_data = true;
     }
 }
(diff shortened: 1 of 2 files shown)


kwagyeman:
This fix is fine but could be faster. The invalidate call could take a very long amount of time. We should put this in sensor.c and make it a little smarter so it does the least amount of work

iabdalkader:
I didn't notice any performance drop, note this is invalidate by addr it doesn't invalidate the whole cache. You can move it to sensor.c and optimize, but if in the future a sensor driver for a port with cache uses the framebuffer it will have to do the same.

kwagyeman:
Yes, but, if you read that function it invalidates 32 bytes at a time in a loop. So, it's going to run for hundreds of thousands of iterations for a larger image.

iabdalkader:
> Yes, but, if you read that function it invalidates 32 bytes at a time in a loop. So, it's going to run for hundreds of thousands of iterations for a larger image

It's unavoidable, but maybe it should be moved to `framebuffer_free_current_buffer()`, get_tail is called multiple times.

kwagyeman (approved):
The cache invalidations in sensor.c are no longer needed.

The flushing here is in the exact right place. Great job!

I tested all the cameras and a bunch of reses and modes. Everything works.

iabdalkader:
@kwagyeman Updated.

kwagyeman:
Fee free to merge

PR #1504, opened 2022-01-15
Fix CAN keyword arg.

iabdalkader:
* sampling_point was renamed to sample_point in upstream.

### DIFF
--- a/scripts/libraries/rpc.py
+++ b/scripts/libraries/rpc.py
@@ -282,20 +282,20 @@ class rpc_slave(rpc):
                 self.__schedule_cb = None
             if self.__loop_cb is not None: self.__loop_cb()
 
-def __get_can_settings(bit_rate, sampling_point):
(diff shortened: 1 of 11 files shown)


scripts/libraries/rpc.py
```diff
@@ -282,20 +282,20 @@ def loop(self, recv_timeout=1000, send_timeout=1000): # public
                 self.__schedule_cb = None
             if self.__loop_cb is not None: self.__loop_cb()
 
-def __get_can_settings(bit_rate, sampling_point):
+def __get_can_settings(bit_rate, sample_point):
```
kwagyeman, line 285: Is this needed anymore now that you put it inside the CAN init method?

iabdalkader: It can calculate bit timing from bitrate.

kwagyeman (approved):

PR #1506, opened 2022-01-19
Fix fb alloc underflow

kwagyeman:


iabdalkader:
Will test it later today, but looks promising.

iabdalkader:
There's no bug, I forgot to update `OMV_RAW_BUF_SIZE`. This macro is defined at the top of the file, I always forget to update it.. Would be good to define it right after linker script FB buffer size, example:

```
#define OMV_FB_SIZE              (100K) // FB memory
#define OMV_FB_SIZE_BYTES (100*1024)
```

iabdalkader:
Or maybe use linker script symbols, `_fballoc - _fb_base`

PR #1509, opened 2022-01-24
Add PNG Support

kwagyeman:
Decoder working, encoder still broken.


iabdalkader:
Fix and resend if you want to compare the performance.

kwagyeman:
I can own doing more upgrades when I have time.

PR #1512, opened 2022-01-26
Fix strncpy issue with new GCC

kwagyeman:
Fixes warning preventing compile on GCC 9.2.1


iabdalkader:
FYI you should be using `gcc version 10.2.1`

PR #1524, opened 2022-02-06
Add support for building customs nets into the firmware

kwagyeman:
(From Updated README)

The build system automatically supports changing the built-in TensorFlow model from the person detection
TensorFlow Lite Model to another model of your choice. Just drop your `.tflite` file into the `src/omv/modules`
folder and the build system will automatically detect the file and build your model into the firmware. Note
that you need to `make clean` between switching from the built-in model to your custom model. Modifying the
file does not require a rebuild. Call `tf.load_builtin_model()` to load your built-in model.

You can also build your model labels into the firmware. Just drop a `.txt` file into the `src/omv/modules`
folder. The text file should have a single line of labels in a C define like below:

    #define IMLIB_TF_BUILT_IN_NETWORK_LABELS "a", "b", "c"

Call `tf.load_builtin_labels()` then to return the list of strings. Note you need to `make clean` between
adding/removing the custom labels file to the firmware. Modifying the file does not require a rebuild.

Finally, your network/labels may not fit inside the firmware. If you encounter a flash overflow you either
need to shrink your network/labels or turn off features in `src/omv/boards/<TARGET>/imlib_config.h`


iabdalkader:
Hi...
1. `tf.load_builtin_model()` should return a tuple (labels, net):

```Python
labels, net = tf.load_builtin_model()
```
It's less typing in scripts, saves some flash, and fewer functions to document. Note labels can be an empty list if there isn't any labels.

2. Please make the person detection model itself a builtin model and process it unconditionally. i.e. remove the model data from the lib, replace it with `builtin_model.tflite` and `builtin_labels.txt`, process with xxd when building unconditionally. To replace the model and labels, you can just overwrite it and rebuild the firmware.
Note: `omv/modules/` is for Python external user C modules.

3. Please don't disable image features unless the flash overflows.

iabdalkader:
Superseded by #1526

PR #1526, opened 2022-02-07
Libtf: Add support for replacing/adding built-in models.

iabdalkader:


src/omv/modules/py_tf.c
```diff
@@ -207,12 +214,39 @@ STATIC mp_obj_t int_py_tf_load(mp_obj_t path_obj, bool alloc_mode, bool helper_m
 
 STATIC mp_obj_t py_tf_load(uint n_args, const mp_obj_t *args, mp_map_t *kw_args)
 {
-    return int_py_tf_load(args[0],
-                          py_helper_keyword_int(n_args, args, 1, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_load_to_fb), false),
-                          false);
+    const char *path = mp_obj_str_get_str(args[0]);
```
kwagyeman, line 217: Seems unnecessary, the code will continue to work without much effort. No reason to print this warning.

kwagyeman (approved):

src/omv/modules/py_tf.c
iabdalkader: There's no function to load built-in labels, you'll have to provide them if you use `tf.load('builtin')` that's why I'm making it obsolete. `load_builtin_model` should be used for built-in models/labels but this could wait for one or two releases, the docs and examples should be updated now though.

kwagyeman: Yeah, so, I'd just update the example scripts and documentation. But, we don't need the warning. The way you wrote the code the old method will continue to work fine indefinitely.

iabdalkader: Okay removed, please update the examples and docs.

kwagyeman:
I'll update the examples and the tf library in a PR. Once that's done my next todo is the docs.

PR #1544, opened 2022-02-18
High temp cleanup

linknum23:
Add a high temp example and use reasonable default temps in C.


iabdalkader:
Hi, thanks for the PR. I have a small request, can you please change the commit titles to match this:
```
Examples: Add Lepton high temp example.
Lepton: Fix default temperature range.
```

PR #1552, opened 2022-02-23
Algorithm output value error

dianjixz:
When imlib_rgb565_to_l function input is zero, the output is a negative, but it should not be a negative value.,116 * 0.137931 = 15.999996.


iabdalkader:
Hi, please fix the commit subject line and pull request title, see: https://github.com/openmv/openmv#contribution-guidelines

dianjixz:
Yes, thank you for reminding me, this is my first time to submit, so it is not so perfect.

PR #1553, opened 2022-02-24
imlib: Fix imlib_rgb565_to_l output.

dianjixz:
It should not be negative when the imlib_rgb565_to_l function input is zero, e.g. y=0.137931f, fast_floorf(116 * 0.137931f) = 15.999996,  15.999996 - 16 < 0.


iabdalkader:
Merged, thank you!

PR #1601, opened 2022-03-30
imlib: Fix quad_segment_maxima memory leak.

dianjixz:
Memory forgot to free when nmaxima is least 4 maxima.


iabdalkader:
Thanks!

PR #1604, opened 2022-04-02
imlib: Enable hardware JPEG decoder.

kwagyeman:
Per @kdarras asking us to fix jpeg support constantly I got on this and finished the implementation of the hardware jpeg decoder using Ibrahim's code. All modes work. It can decode grayscale jpegs to binary/grayscale/rgb565. It can decode color 444/420/422 jpegs to binary/grayscale/rgb565. It uses DMA2D when it can. And it handles jpegs that aren't a multiple of 8x8, 16x8, 16x16 depending on the MCU size.

JPEG support issues should be closed now.


kwagyeman:
Also, cropping and scaling jpeg images now work as seamlessly as non-jpeg images.

iabdalkader:
* Why move the code from `jpegd.c` which is the sw/hw decoder file ? I prefer if encoder/decoder are kept separate. Any good reason for that ? Alternatively, the software encoder/decoder can be in a file and the hardware encoder/decoder can be in another file (ex `jpeg_sw.c` and `jpeg_hw.c`) although some code is shared between encoders and decoders.
* You dropped the progressive JPEG sanity check. The SW/HW decoder only supports baseline JPEG.
* Are you using software for GS to GS ? Because DMA2D can be used for GS->GS if set in `DMA2D_M2M` mode, in which it basically just copies 2D blocks of pixels.

kwagyeman:
> Why move the code from jpegd.c which is the sw/hw decoder file ? I prefer if encoder/decoder are kept separate. Any good reason for that ? Alternatively, the software encoder/decoder can be in a file and the hardware encoder/decoder can be in another file (ex jpeg_sw.c and jpeg_hw.c) although some code is shared between encoders and decoders.

I re-use the same handles and constants from the jpeg encoder. I'd have to extern a bunch of stuff and duplicate defines otherwise.

> You dropped the progressive JPEG sanity check. The SW/HW decoder only supports baseline JPEG.

It's still there. However, I made it more through. See the start of the decode method.

> Are you using software for GS to GS ? Because DMA2D can be used for GS->GS if set in DMA2D_M2M mode, in which it basically just copies 2D blocks of pixels.

Yes, I didn't think of this. Yes, you could improve the jpeg transfer speed here by using DMA2D_M2M.

> SW decoder still needs to work for other MCUs.

It's still enabled for other MCUs. Larry will have to fix any problems with it.

iabdalkader:
> I'd have to extern a bunch of stuff and duplicate defines otherwise.

That's fine, encoder and decoder should be in separate files. Alternatively you can move all the hardware jpeg to `jpeg_hw.c` and the software jpeg to `jpeg_sw.c` you choice, but the sw encoder here and the sw decoder there is Not okay.

> It's still there. However, I made it more through. See the start of the decode method.

I saw it now, you should move these checks to `jpeg_is_progressive` and improve it, which is also used by the software decoder, and rename it if you want to `jpeg_is_valid` or something.

iabdalkader (changes requested):
> I'd have to extern a bunch of stuff and duplicate defines otherwise.

That's fine, encoder and decoder should be in separate files. Alternatively you can move all the hardware jpeg to `jpeg_hw.c` and the software jpeg to `jpeg_sw.c` your choice, this is probably cleaner, but the sw encoder here and the sw decoder there is Not okay.

> It's still there. However, I made it more through. See the start of the decode method.

I saw it now, you should move these checks to `jpeg_is_progressive` and improve it, which is also used by the software decoder, and rename it if you want to `jpeg_is_valid` or something.

> Yes, I didn't think of this. Yes, you could improve the jpeg transfer speed here by using DMA2D_M2M.

The working decoder code that you removed did that already. Is this an improvement or a regression ? 

>   // STM32H7 BUG FIX! The jpeg decoder incorrectly computes the number of MCUs to generate by doing
    // ((w * h) / (mcu_w * mcu_h)) verus ((w + mcu_w - 1) / mcu_w) * ((w + mcu_h - 1) / mcu_h)). This
    // results the jpeg decoder stalling when you have an image that is not a multiple of the MCU
    // width/height. By patching the JPEG bytestream with the correct width and height the JPEG
    // decoder will not stall anymore. We then ignore the excess pixels in our code..

It does Not. The JPEG decoder outputs padded MCUs lines/columns. For example, a 300x240 Grayscale image will output 38x30 MCUs, see the manual and appnotes...This is probably an issue with the output chunk size you set, or something else.

image
This is how I calculated the MCUs per line in the old code:

```
mcu_per_line = (width / mcu_width) + !!(width % mcu_width);
// Example: (300 / 8) + !!(300 % 8) == 38 MCUs.
```

kwagyeman:
> That's fine, encoder and decoder should be in separate files. Alternatively you can move all the hardware jpeg to jpeg_hw.c and the software jpeg to jpeg_sw.c your choice, this is probably cleaner, but the sw encoder here and the sw decoder there is Not okay.

Maybe you should move around in another PR since you are particular about this? I didn't like the jpegd.c file int he first place. But, since the jpeg decoder was so much code it made since to put it into it's own file.

> I saw it now, you should move these checks to jpeg_is_progressive and improve it, which is also used by the software decoder, and rename it if you want to jpeg_is_valid or something.

I guess it needs to be shared between the both hardware and software decoders.

> It does Not. The JPEG decoder outputs padded MCUs lines/columns. For example, a 300x240 Grayscale image will output 38x30 MCUs, see the manual and appnotes...This is probably an issue with the output chunk size you set, or something else.

No, you don't understand, the hardware jpeg decoder is literally bugged. If you give it an image that's not a multiple of 8 in width/height it doesn't do what you expect. My comment explains the behavior exactly. E.g. 300x240 will produce exactly 1,125 MCUs out of output versus 1140.

Without this fix it's not possible to handle non-MCU widths/heights. The decoder will just stall on you and fail to produce the expected number of MCUs.

---

Anyway, what do you want fixed?

iabdalkader:
> Maybe you should move around in another PR since you are particular about this? I

It's your PR you fix it.

> the hardware jpeg decoder is literally bugged. If you give it an image that's not a multiple of 8 in width/height it doesn't do what you expect

The text literally says if the image is Not a multiple of 8 or 16 last MCUs will be padded. Also how do you explain the old decoder working on images not a multiple of 8/16 ? You can test it btw just enable it in master and then you can figure out what's wrong with yours.

iabdalkader:

https://forums.openmv.io/t/jpeg-hardware-decode/8811/2

kwagyeman:
@iabdalkader Going to fix this PR now. Will do everything you want. Let me know if you want anything else.

iabdalkader:

https://forums.openmv.io/t/image-not-rendering-correctly-on-arduino-giga-display/8852/10

kwagyeman:
Refactoring jpeg decoder and encoder into another file is done. Need to use M2M mode for grayscale now.

iabdalkader:
> Need to use M2M mode for grayscale now.

No it's not important it's just not worth complicating the code any further for this. However you do need to address all of the comments above, in particular:

> The JPEG decoder outputs padded MCUs lines/columns. For example, a 300x240 Grayscale image will output 38x30 MCUs, see the manual and appnotes...This is probably an issue with the output chunk size you set, or something else.

The disabled code that existed before handled non-multiple of 8, the docs say it does, and it's not mentioned in any errata.

> I saw it now, you should move these checks to jpeg_is_progressive and improve it, which is also used by the software decoder, and rename it if you want to jpeg_is_valid or something.

kwagyeman:
Debugging it.

kwagyeman:
Rebased on master and addressed the jpeg_is_progressive is comment. Regarding M2M mode for grayscale. It's not actually possible to do since DMA2D can't output L8, it can only input L8.

Working on debugging the mcu issue.

kwagyeman:
Okay, I debugged the issue. The width/height patching code is not needed. When I was originally writing this code I must have not properly reset the encoder config when switching between jpeg encoding and decoding. If you don't do this the system crashes in interesting ways and you get all kinds of corrupt jpegs. So, the image patching was a hack making it appear to work...

Anyway, it appears I added code to reset the encoder config later on if the decoder config differs. So... the whole image patching part can just be deleted.

metanav:
Hi @kwagyeman,

Just curious, any specific reason to not using HAL_JPEG_Decode_DMA?

kwagyeman:
I actually use that method. The code DMA accelerated for encoding and decoding.

metanav:
> I actually use that method. The code DMA accelerated for encoding and decoding. 
I could find the HAL_JPEG_Decode_DMA but not HAL_JPEG_Decode_DMA.
https://github.com/kwagyeman/openmv/blob/718044c7da0c15b34e352e4b992086dc7d51d67e/src/omv/imlib/jpegd.c#L275

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+            while (!JPEG_input_paused) {
+                __WFI();
+
+                if (JPEG_output_paused) {
+                    memset(&JPEG_Config, 0, sizeof(JPEG_ConfTypeDef));
+                    HAL_JPEG_Abort(&JPEG_Handle);
+                    fb_free(); // mcu_row_buffer (after DMA is aborted)
+                    return true; // overflow
+                }
+            }
```
iabdalkader, line 230: We should use timeout and poll events for all of those loops.
```suggestion
            for (mp_uint_t tickstart = mp_hal_ticks_ms(); !JPEG_input_paused; ) {
                if (JPEG_output_paused || (mp_hal_ticks_ms() - tickstart) > DECODE_TIMEOUT) {
                    jpeg_overflow = true;
                    goto exit_cleanup;
                }
                MICROPY_EVENT_POLL_HOOK;
            }
```

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+            JPEG_Info.ChromaSubsampling = JPEG_444_SUBSAMPLING;
+            break;
+        default:
+            break;
+    }
+
+    if (memcmp(&JPEG_Config, &JPEG_Info, sizeof(JPEG_ConfTypeDef))) {
+        HAL_JPEG_ConfigEncoding(&JPEG_Handle, &JPEG_Info);
+        memcpy(&JPEG_Config, &JPEG_Info, sizeof(JPEG_ConfTypeDef));
+    }
```
iabdalkader, line 138:
```suggestion
    if (jpeg_state.config_invalid) {
        jpeg_state.config_invalid = false;
        HAL_JPEG_ConfigEncoding(&JPEG_Handle, &JPEG_Info);
    }
```

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+                mcu_h = MCU_H * 2;
+                mcu_size = JPEG_420_YCBCR_MCU_SIZE;
+                break;
+            }
+            case JPEG_422_SUBSAMPLING: {
+                DMA2D_Handle.LayerCfg[1].ChromaSubSampling = DMA2D_CSS_422;
+                mcu_w = MCU_W * 2;
+                mcu_h = MCU_H;
+                mcu_size = JPEG_422_YCBCR_MCU_SIZE;
+                break;
```
iabdalkader, line 400:
```suggestion
            case JPEG_444_SUBSAMPLING: {
                mcu_w = MCU_W;
                mcu_h = MCU_H;
                mcu_size = JPEG_444_YCBCR_MCU_SIZE;
                DMA2D_Handle.LayerCfg[1].ChromaSubSampling = DMA2D_NO_CSS;                
                break;
            }
            case JPEG_420_SUBSAMPLING: {
                mcu_w = MCU_W * 2;
                mcu_h = MCU_H * 2;
                mcu_size = JPEG_420_YCBCR_MCU_SIZE;
                DMA2D_Handle.LayerCfg[1].ChromaSubSampling = DMA2D_CSS_420;                
                break;
            }
            case JPEG_422_SUBSAMPLING: {
                mcu_w = MCU_W * 2;
                mcu_h = MCU_H;
                mcu_size = JPEG_422_YCBCR_MCU_SIZE;
                DMA2D_Handle.LayerCfg[1].ChromaSubSampling = DMA2D_CSS_422;                
                break;
```

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+            HAL_DMA2D_Init(&DMA2D_Handle);
+            DMA2D_Handle.LayerCfg[1].InputOffset = (((src->w + mcu_w - 1) / mcu_w) * mcu_w) - src->w;
+            DMA2D_Handle.LayerCfg[1].InputColorMode = DMA2D_INPUT_YCBCR;
+            DMA2D_Handle.LayerCfg[1].AlphaMode = DMA2D_NO_MODIF_ALPHA;
+            DMA2D_Handle.LayerCfg[1].InputAlpha = 0xFF;
+            DMA2D_Handle.LayerCfg[1].AlphaInverted = DMA2D_REGULAR_ALPHA;
+            DMA2D_Handle.LayerCfg[1].RedBlueSwap = DMA2D_RB_REGULAR;
+            HAL_DMA2D_ConfigLayer(&DMA2D_Handle, 1);
+            // Invalidate the dst image for DMA2D.
+            SCB_InvalidateDCache_by_Addr((uint32_t *) dst->data, image_size(dst));
```
iabdalkader, line 425:
```suggestion
            DMA2D_Handle.Instance = DMA2D;
            
            // Configure DMA2D output.
            DMA2D_Handle.Init.Mode = DMA2D_M2M_PFC;
            DMA2D_Handle.Init.ColorMode = DMA2D_OUTPUT_RGB565;
            DMA2D_Handle.Init.OutputOffset = 0;
            DMA2D_Handle.Init.AlphaInverted = DMA2D_REGULAR_ALPHA;
            DMA2D_Handle.Init.RedBlueSwap = DMA2D_RB_REGULAR;
            DMA2D_Handle.Init.BytesSwap = DMA2D_BYTES_REGULAR;
            DMA2D_Handle.Init.LineOffsetMode = DMA2D_LOM_PIXELS;
           
            // Configure DMA2D input.
            DMA2D_Handle.LayerCfg[1].InputOffset = (((src->w + mcu_w - 1) / mcu_w) * mcu_w) - src->w;
            DMA2D_Handle.LayerCfg[1].InputColorMode = DMA2D_INPUT_YCBCR;
            DMA2D_Handle.LayerCfg[1].AlphaMode = DMA2D_NO_MODIF_ALPHA;
            DMA2D_Handle.LayerCfg[1].InputAlpha = 0xFF;
            DMA2D_Handle.LayerCfg[1].AlphaInverted = DMA2D_REGULAR_ALPHA;
            DMA2D_Handle.LayerCfg[1].RedBlueSwap = DMA2D_RB_REGULAR;
            
            // DMA2D initialization.
            HAL_DMA2D_Init(&DMA2D_Handle);
            HAL_DMA2D_ConfigLayer(&DMA2D_Handle, 1);
            
            // Invalidate the dst image for DMA2D.
            SCB_InvalidateDCache_by_Addr((uint32_t *) dst->data, image_size(dst));
```

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+                                int v = p > 128;
+                                IMAGE_PUT_BINARY_PIXEL_FAST(rp, x, v);
+                            }
+
+                            Y0 += MCU_W - dx;
+                        }
+                        break;
+                    }
+                    case PIXFORMAT_GRAYSCALE: {
+                        for (int y = y_offset, yy = y + dy; y < yy; y++) {
```
iabdalkader, line 520: Please no more of this.
```suggestion
                        for (int y = y_offset; y < dy + y_offset; y++) {
```

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+#if (TIME_JPEG == 1)
+#include <stdio.h>
+#endif
+
+#define DECODE_TIMEOUT              (1000)
+#define FB_ALLOC_PADDING            ((__SCB_DCACHE_LINE_SIZE) * 4)
+#define OUTPUT_CHUNK_SIZE           (512) // The minimum output buffer size is 2x this - so 1KB.
+#define MAX_BLOCK_SIZE              (65536UL) // Maximum bytes MDMA can transfer at once.
+#define JPEG_INPUT_FIFO_BYTES       (32)
+#define JPEG_OUTPUT_FIFO_BYTES      (32)
```
iabdalkader, line 31:
```suggestion
#define JPEG_CODEC_TIMEOUT       (1000)
#define JPEG_ALLOC_PADDING            ((__SCB_DCACHE_LINE_SIZE) * 4)
#define JPEG_OUTPUT_CHUNK_SIZE           (512) // The minimum output buffer size is 2x this - so 1KB.
#define JPEG_MAX_MDMA_BLOCK_SIZE              (65536UL) // Maximum bytes MDMA can transfer at once.
#define JPEG_INPUT_FIFO_BYTES       (32)
#define JPEG_OUTPUT_FIFO_BYTES      (32)
```

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+
+    fb_free(); // mcu_row_buffer
+
+    // Set output size.
+    dst->size = JPEG_out_data_length;
+
+    // STM32H7 BUG FIX! The JPEG Encoder will occasionally trigger the EOCF interrupt before writing
+    // a final 0x000000D9 long into the output fifo as the end of the JPEG image. When this occurs
+    // the output fifo will have a single 0 value in it after the encoding process finishes.
+    if (__HAL_JPEG_GET_FLAG(&JPEG_Handle, JPEG_FLAG_OFNEF) && (!JPEG_Handle.Instance->DOR)) {
```
iabdalkader, line 263: Are you sure about this ? Note you were also very sure about the other one.

iabdalkader (changes requested):

iabdalkader:
Once this is done, in another PR please rename `omv/imlib/jpeg.c` to `omv/imlib/jpege.c`

src/omv/ports/stm32/jpeg.c
kwagyeman: Yes, this definitely occurs.

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+ *
+ * Copyright (c) 2013-2023 Ibrahim Abdelkader <iabdalkader@openmv.io>
+ * Copyright (c) 2013-2023 Kwabena W. Agyeman <kwagyeman@openmv.io>
+ *
+ * This work is licensed under the MIT license, see the file LICENSE for details.
+ *
+ * Hardware Accelerated JPEG Encoder and Decoder
+ */
+#include "imlib.h"
+#if (OMV_HARDWARE_JPEG == 1)
```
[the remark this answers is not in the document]
kwagyeman: Sure, changing, but, imlib.h includes omv_boardconfig.h.

src/omv/imlib/jpegd.c
```diff
@@ -3064,8 +2795,8 @@ void jpeg_decompress(image_t *dst, image_t *src) {
     #endif
 
     // Supports decoding baseline JPEGs only.
-    if (jpeg_is_progressive(src)) {
-        mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Progressive JPEG is not supported."));
+    if (!jpeg_is_valid(src)) {
+        mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Non-Baseline JPEG!"));
```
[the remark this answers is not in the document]
kwagyeman: Done.

src/omv/ports/stm32/jpeg.c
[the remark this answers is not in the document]
kwagyeman: Mmm, I guess that's not necessary. Doing what you said just speeds up the compile time.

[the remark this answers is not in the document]
kwagyeman: Done

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+
+    uint8_t *mcu_row_buffer = fb_alloc(src_w_mcus_bytes_2, FB_ALLOC_PREFER_SPEED | FB_ALLOC_CACHE_ALIGN);
+
+    for (int y_offset = 0; y_offset < src->h; y_offset += MCU_H) {
+        uint8_t *mcu_row_buffer_ptr = mcu_row_buffer + (src_w_mcus_bytes * ((y_offset / MCU_H) % 2));
+
+        int dy = src->h - y_offset;
+        if (dy > MCU_H) {
+            dy = MCU_H;
+        }
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+                    return true; // overflow
+                }
+            }
+
+            // Reset the lock.
+            JPEG_input_paused = false;
+
+            // Restart the DMA process on the next row of MCUs (that were already prepared).
+            HAL_JPEG_ConfigInputBuffer(&JPEG_Handle, mcu_row_buffer_ptr, src_w_mcus_bytes);
+            HAL_JPEG_Resume(&JPEG_Handle, JPEG_PAUSE_RESUME_INPUT);
```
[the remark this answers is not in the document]
kwagyeman: Done

kwagyeman:
> Once this is done, in another PR please rename omv/imlib/jpeg.c to omv/imlib/jpege.c

Done

src/omv/ports/stm32/jpeg.c
kwagyeman: Done

src/omv/ports/stm32/jpeg.c
kwagyeman: Done

src/omv/ports/stm32/jpeg.c
kwagyeman: Done

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+
+static JPEG_HandleTypeDef JPEG_Handle = {};
+static JPEG_ConfTypeDef JPEG_Config = {};
+static MDMA_HandleTypeDef JPEG_MDMA_Handle_In = {};
+static MDMA_HandleTypeDef JPEG_MDMA_Handle_Out = {};
+
+static volatile int JPEG_out_data_length_max = 0;
+static volatile int JPEG_out_data_length = 0;
+static volatile bool JPEG_input_paused = false;
+static volatile bool JPEG_output_paused = false;
```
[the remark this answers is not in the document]
kwagyeman: Working on this... out of time for today.

kwagyeman:
Posting Performance Numbers Just for reference:
- HD (1280x720) RGB565 (Old code) 102ms, new code is 21ms (4.85x speedup)
- Grayscale (Old code) 44ms, new code is 14 ms (3.14x speedup)
- SVGA (800x600) RGB565 (Old code) 53ms, new code is 10ms (5.3x speedup)
- Grayscale (Old code) 23ms, new code is 4ms (5.75x speedup)
- VGA (640x480) RGB565 (Old code) 32ms, new code is 6ms (5.33x speedup)
- Grayscale (Old code) 15ms, new code is 2ms (7.5x speedup)

Software Decoder Performance:
- HD (1280x720) RGB565 - 209ms, Grayscale - 66ms
- SVGA (800x600) RGB565 - 109ms, Grayscale - 36ms
- VGA (640x480) RGB565 - 70ms, Grayscale - 23ms

kwagyeman:
Will move commit that renames jpeg file into another PR. Was just saving it right now before I leave.

src/omv/ports/stm32/jpeg.c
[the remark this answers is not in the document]
kwagyeman: Done, note, that the in/out-count variables in the jpeg structure get reset by the HAL code so I can't use them directly. I was able to remove the extra Config variable, however.

kwagyeman:
Addressed all issues except 4 hidden messages. Did you want me to address those? Also, leaving the commit to rename the jpeg file in this PR. I can move it to another PR if necessary.

iabdalkader:
> Addressed all issues except 4 hidden messages. Did you want me to address those?

Yes. Github displays a max number of conversations and hides the rest, all comments need to be fixed first.

src/omv/imlib/imlib.h
```diff
@@ -732,6 +732,17 @@ bool image_get_mask_pixel(image_t *ptr, int x, int y);
         ((uint16_t *) _image->data) + (_image->w * _y); \
     })
 
+////////////////
+// JPEG Stuff //
+////////////////
+
+#define MCU_W                      (8)
+#define MCU_H                      (8)
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/ports/stm32/jpeg.c
kwagyeman: Done, also, cleaned up loop exits using goto.

src/omv/ports/stm32/jpeg.c
kwagyeman: Done

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,753 @@
...
+    // Decode the JPEG Header...
+    memset(&JPEG_Info, 0, sizeof(JPEG_ConfTypeDef));
+    uint8_t *temp = fb_alloc(OUTPUT_CHUNK_SIZE, FB_ALLOC_NO_HINT);
+    HAL_JPEG_Decode(&JPEG_Handle, (uint8_t *) JPEG_in_data, JPEG_in_data_len, temp, OUTPUT_CHUNK_SIZE, DECODE_TIMEOUT);
+    fb_free(); // temp
+
+    // Reset the encoder config if the decoder config is different.
+    if (memcmp(&JPEG_Config, &JPEG_Info, sizeof(JPEG_ConfTypeDef))) {
+        memset(&JPEG_Config, 0, sizeof(JPEG_ConfTypeDef));
+    }
```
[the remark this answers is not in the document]
kwagyeman: Done

kwagyeman:
@iabdalkader All comments have been addressed.

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,707 @@
...
+            fb_alloc_fail();
+        }
+
+        dst->size = IMLIB_IMAGE_MAX_SIZE(avail - space);
+        dst->data = fb_alloc(dst->size, FB_ALLOC_PREFER_SIZE | FB_ALLOC_CACHE_ALIGN);
+    }
+
+    if (src->is_compressed) {
+        return true;
+    }
```
iabdalkader, line 164: What should happen in this case ? Should we just copy src to dst ? Note dst might be fb_alloc'd just above this.

iabdalkader: Oh I forgot that returning true meant overflow, I guess it's fine.

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,707 @@
...
+    if (diff) {
+        JPEG_state.in_data_len += JPEG_INPUT_FIFO_BYTES - diff;
+    }
+
+    // Copy to cache aligned buffer.
+    JPEG_state.jpeg_descr.pJpegInBuffPtr = fb_alloc(JPEG_state.in_data_len, FB_ALLOC_CACHE_ALIGN);
+    memcpy(JPEG_state.jpeg_descr.pJpegInBuffPtr, src->data, src->size);
+
+    // Zero remaining.
+    memset(JPEG_state.jpeg_descr.pJpegInBuffPtr + src->size, 0, JPEG_state.in_data_len - src->size);
```
iabdalkader, line 342: There's room for a minor optimization here, maybe the source buffer is already cache-aligned ?

kwagyeman: Added it. This is actually a huge speedup if aligned if the JPEG is large.

src/omv/ports/stm32/jpeg.c
kwagyeman: Yep. Top level code free()s.

kwagyeman:
Restested and it's fine. Works in all modes.

src/omv/ports/stm32/jpeg.c
```diff
@@ -0,0 +1,713 @@
...
+
+            // DMA2D initialization.
+            HAL_DMA2D_Init(&DMA2D_Handle);
+            HAL_DMA2D_ConfigLayer(&DMA2D_Handle, 1);
+
+            // Invalidate the dst image for DMA2D.
+            SCB_InvalidateDCache_by_Addr((uint32_t *) dst->data, image_size(dst));
+        }
+    } else if (JPEG_state.jpeg_descr.Conf.ColorSpace == JPEG_CMYK_COLORSPACE) {
+        fb_free(); // JPEG_state.jpeg_descr.pJpegInBuffPtr
```
iabdalkader, line 421: But now you have to check if it was allocated or aligned, maybe `JPEG_state.jpeg_descr.pJpegInBuffPtr != src->data`

iabdalkader: Or `if (((uint32_t) src->data) % __SCB_DCACHE_LINE_SIZE) {` like below.

kwagyeman: Good catch.

kwagyeman: Fixed

src/omv/ports/stm32/jpeg.c
kwagyeman: @iabdalkader Making this change destroyed JPEG performance. This is because this keeps the JPEG buffer in SDRAM when MDMA is accessing it. Sending another PR to remove this change.

kwagyeman: Mmm, on further testing, this doesn't seem to be the culprit. However, JPEG performance is generally less than what I measured before I made many of the requested changes. Will need to figure out what is costing so much speed.

Something else though is causing a vary serious slowdown when trying to play MJPEG files.

kwagyeman: Mmm, false alarm. So, when the jpeg decoder gets stuck it just partially decodes an image and you busy wait for 1 second on the tiemout.

kwagyeman: https://github.com/openmv/openmv/pull/2093 - Would have saved hours of debugging.

iabdalkader: I missed this, all good now, right ?

kwagyeman: Yeah.

PR #1605, opened 2022-04-02
imlib: Fix black lines on first frame.

### LINKED ISSUE #1580: First frame has corrupted lines.
**Describe the bug**
The first frame (technically the second, since we skip the first) shows corrupted lines when setting a window. It's very likely a cache flush/invalidate issue.

**To Reproduce**
```python
import sensor, image, time

sensor.reset()
sensor.set_framebuffers(1)
sensor.set_windowing((0, 0, 240, 240))
sensor.snapshot().flush()
time.sleep_ms(5000)
```

**Screenshots**
image

 - IDE version: [latest]
 - Firmware version: [latest]
 - Board: [OPENMV-4]
 - OS: [Linux]

kwagyeman:
All frame buffers need to be invalidated before starting MDMA frame capture. The invalidation is done as late as possible so that it does not get repeatedly done over and over again in the sensor_set_*() calls.

I added the method to all MCUs so that if we find we need to do anything else in the future it can be put inside this same method.

Fixes  #1580

### DIFF
--- a/src/omv/imlib/framebuffer.c
+++ b/src/omv/imlib/framebuffer.c
@@ -368,7 +368,7 @@ void framebuffer_free_current_buffer()
 {
     vbuffer_t *buffer = framebuffer_get_buffer(framebuffer->head);
     #ifdef __DCACHE_PRESENT
-    // Make sure all cached CPU writes are flushed before returning the buffer.
+    // Make sure all cached CPU writes are discarded before returning the buffer.
     SCB_InvalidateDCache_by_Addr(buffer->data, framebuffer_get_buffer_size());
     #endif
 
@@ -379,6 +379,19 @@ void framebuffer_free_current_buffer()
     if (framebuffer->n_buffers == 1) {
         buffer->waiting_for_data = true;
     }
+}
+
+void framebuffer_setup_buffers()
+{
+    #ifdef __DCACHE_PRESENT
+    for (int32_t i = 0; i < framebuffer->n_buffers; i++) {
+        if (i != framebuffer->head) {
+            vbuffer_t *buffer = framebuffer_get_buffer(i);
+            // Make sure all cached CPU writes are discarded before returning the buffer.
+            SCB_InvalidateDCache_by_Addr(buffer->data, framebuffer_get_buffer_size());
+        }
+    }
+    #endif
 }
 
 vbuffer_t *framebuffer_get_head(framebuffer_flags_t flags)
--- a/src/omv/imlib/framebuffer.h
+++ b/src/omv/imlib/framebuffer.h
@@ -116,6 +116,9 @@ void framebuffer_auto_adjust_buffers();
 // Call when done with the current vbuffer to mark it as free.
 void framebuffer_free_current_buffer();
 
+// Call to do any heavy setup before frame capture.
+void framebuffer_setup_buffers();
+
 // Sets the current frame buffer to the latest virtual frame buffer.
 // Returns the buffer if it is ready or NULL if not...
 // Pass FB_PEEK to get the next buffer but not take it.
--- a/src/omv/ports/nrf/sensor.c
+++ b/src/omv/ports/nrf/sensor.c
@@ -183,6 +183,7 @@ int sensor_snapshot(sensor_t *sensor, image_t *image, uint32_t flags)
     }
 
     framebuffer_free_current_buffer();
+    framebuffer_setup_buffers();
     vbuffer_t *buffer = framebuffer_get_tail(FB_NO_FLAGS);
 
     if (!buffer) {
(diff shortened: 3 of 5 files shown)


src/omv/imlib/framebuffer.c
```diff
@@ -381,6 +381,19 @@ void framebuffer_free_current_buffer()
     }
 }
 
+void framebuffer_setup_buffers()
+{
+    #ifdef __DCACHE_PRESENT
+    for (int32_t i = 0; i < framebuffer->n_buffers; i++) {
+        if (i != framebuffer->head) {
+            vbuffer_t *buffer = framebuffer_get_buffer(i);
+            // Make sure all cached CPU writes are flushed before returning the buffer.
```
iabdalkader, line 390: This comment (which is copied from `framebuffer_free_current_buffer`) is wrong, because this invalidates the cache not flush writes.

kwagyeman: It's job is to the same as the other line except for all other buffers versus the one at framebuffer->head. So, the comment is correct.

Both comments are wrong using the word flushed otherwise.

kwagyeman: Updated the wording.

iabdalkader:
Regarding the code comment I was just pointing out that both comments wording is wrong, which can be misleading, thanks for fixing. That said, I still need to test this because I don't understand why we need to invalidate all buffers before every snapshot, the CPU does Not access the buffer at this point. You did not explain what was the issue and why this fix works, it seems like it just works. So please leave this open for a while until I can test it.

I think the invalidate should happen once in `framebuffer_get_head` you can test that or wait for me to get the hw and test it.

iabdalkader:
See #1610

kwagyeman:
You have 3 buffers.

Only the first gets invalidated. The other two aren't.

So, when MDMA writes to them they aren't invalidate.

Note to that the call to set_windowing() resets the buffer count to 3 after you set it to 1.

kwagyeman:
You can't invalidated in get tail because it will take too long and it's too late in get head. The cpu might have flushed something via a write back then.

iabdalkader:
> So, when MDMA writes to them they aren't invalidated.

You seem to be confusing invalidate and clean. Invalidate will drop cached lines, clean will write back dirty lines (cleaninvalidate will do both).
When MDMA writes to the vbuffer they don't need to be invalidated first, the vbuffer needs to be invalidated just before handing it back to the CPU and the CPU is not using reading/writing the vbuffer while MDMA is writing to it.

Finally, this fix works too it's been tested.

kwagyeman:
I'm not confusing the terminology. If you don't invalidate the image buffer before starting MDMA then the processor will write back lines to any region it was working on before when it needs to free up space. This is what has been causing the issue.

Sounds good that this works. Seems cleaner.

iabdalkader:
> The cpu might have flushed something via a write back then.

Okay that's very unlikely also I'm not sure if the memory isn't write-through, but anyway in that case we just need to invalidate in both `free_buffer` and in `get_head`. In `free_buffer` any cached writes will be dropped before relinquishing the buffer to MDMA and in `get_head` any cached reads will be drop handing the buffer back to the CPU. Does this fix the issue ? Am I missing anything else ?

kwagyeman:
I think you kinda have it. The problem is just you have to invalidate the next buffer MDMA will write too.

However, remember there are N buffers. This only ensures 1 is cleared. The other two aren't.

Given MDMA automatically will switch to the next buffer... this fix isn't likely to totally fix the issue.

Like, let's say you allocate 10 buffers. MDMA is going to write to 9 of them one after another. Some cached writes may be pending them and could get flushed.

This is why I was invalidating all the buffers except the one that you freed and invalidated via another call.

kwagyeman:
If you noticed in my PR I invalidated all the buffers if the MDMA needed to be restarted. And only in this condition.

kwagyeman:
Yeah... so, my PR remains the fix for this. Maybe it could be done cleaner. But, you still essentially need to invalidate all buffers if MDMA is being restarted to ensure the CPU doesn't have any dirty lines it needs to write back.

I guess there's a performance optimization here though because the  CPU always only controls 1 buffer and it flushes it when it gives it back...

iabdalkader:
> let's say you allocate 10 buffers. MDMA is going to write to 9 of them one after another. Some cached writes may be pending them and could get flushed.

Initially on first run there won't be any cached writes, and one following snapshots MDMA will Not write a buffer unless it was previously free'd with `free_buffer` and consequently invalidated. 

> you still essentially need to invalidate all buffers if MDMA is being restarted to ensure the CPU doesn't have any dirty lines it needs to write back.

For this special case you could invalidate all the framebuffer memory or all the buffers yes.

kwagyeman:
This PR is needed along with: https://github.com/openmv/openmv/pull/1610

PR #1606, opened 2022-04-03
imlib: Fix py_lcd cache alignment and handling.

kwagyeman:
Minor fixes for the LCD output on the OpenMV PT. These fixes don't appear to help any bugs. However, the previous code was not the best.


iabdalkader:
I made similar fixes to draw.c to get the hardware decoder working.

PR #1607, opened 2022-04-03
stm32: Fix module deinit between main.py to USBDBG.

kwagyeman:
The change below looks complex but I just moved everything under "else".

When you stop main, it doesn't run all the deinit methods before USBDBG starts. This puts the camera into a weird state. We never saw any issues with this before because our main.py was simple. However, for the OpenMVPT main.py this results in the driver state being off as quite alot more stuff was going on than blinking an LED.

### DIFF
--- a/src/omv/ports/stm32/main.c
+++ b/src/omv/ports/stm32/main.c
@@ -567,62 +567,62 @@ soft_reset:
     // Run main script if it exists.
     if (first_soft_reset) {
         exec_boot_script("main.py", false, true, openmv_config.wifidbg);
-    }
+    } else {
+        do {
+            usbdbg_init();
 
-    do {
-        usbdbg_init();
+            if (openmv_config.wifidbg == true) {
+                // Need to reinit imlib in WiFi debug mode.
+                imlib_deinit_all();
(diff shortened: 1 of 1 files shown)


iabdalkader:
> When you stop main, it doesn't run all the deinit methods before USBDBG starts

Yeah it's not supposed to, this will introduce a delay (deinit, reinit, soft-reboot) for all cams when the IDE connects and stops the running script (`main.py`). This is not a good fix, I'd debug and find out what's causing the actual issue.

kwagyeman:
The issue is that hardware is not deinited.

It's straight forward. Init0 calls need to be executed and such.

iabdalkader:
> The issue is that hardware is not deinited.

Something needs to be deinitialized/reinitialized properly before being used.

kwagyeman:
Yes, the above fix is the easiest way to do this by just causing a soft reset when main.py is stopped.

iabdalkader:
> Yes, the above fix is the easiest way to do this by just causing a soft reset when main.py is stopped

I'm not slowing down every single stm32 board because it's easier to fix it this way, come on.

PR #1610, opened 2022-04-08
imlib: Invalidate vbuffers before returning them.

### LINKED ISSUE #1580: First frame has corrupted lines.

### LINKED ISSUE #2216: Corrupt Lines
When the number of frame buffers is set to 1 on the RT1062 you'll get those corrupt cache invalidation lines on small resolutions like QQVGA. That said, unlike before, it's only one line near the top left of the frame. Not all over it. So, it's a small race condition somewhere. Most likely has to do with the eDMA capture potentially falling behind. The code doesn't wait for it to finish before releasing the frame right now. Probably need to add some code to wait on it finishing.

It doesn't seem to bother multiple buffer mode, which everyone uses. But, something to eventually fix.

iabdalkader:
The CPU might fetch lines, speculatively, from a vbuffer after an invalidate and before the [M/E]DMA is done writing to it. In this case, when the DMA is done, the CPU will have some "outdated" lines in cache. Speculative access in CM7 is documented here https://developer.arm.com/documentation/ddi0489/f/memory-system/speculative-accesses

> Speculative cache linefills can be initiated to any Cacheable memory address, and in rare cases, regardless of whether there is any instruction that causes the cache linefill.

This change invalidates the buffer after the DMA is done and before returning it, to ensure the CPU drops any cached lines of that vbuffer.

* Fixes #2216

### DIFF
--- a/src/omv/imlib/framebuffer.c
+++ b/src/omv/imlib/framebuffer.c
@@ -342,6 +342,7 @@ void framebuffer_auto_adjust_buffers() {
 
 void framebuffer_free_current_buffer() {
     vbuffer_t *buffer = framebuffer_get_buffer(framebuffer->head);
+
     #ifdef __DCACHE_PRESENT
     // Make sure all cached CPU writes are discarded before returning the buffer.
     SCB_InvalidateDCache_by_Addr(buffer->data, framebuffer_get_buffer_size());
@@ -400,7 +401,14 @@ vbuffer_t *framebuffer_get_head(framebuffer_flags_t flags) {
         framebuffer->head = new_head;
     }
 
-    return framebuffer_get_buffer(new_head);
+    vbuffer_t *buffer = framebuffer_get_buffer(new_head);
+
+    #ifdef __DCACHE_PRESENT
+    // Make sure any cached CPU reads are dropped before returning the buffer.
+    SCB_InvalidateDCache_by_Addr(buffer->data, framebuffer_get_buffer_size());
+    #endif
+
+    return buffer;
 }
 
 vbuffer_t *framebuffer_get_tail(framebuffer_flags_t flags) {


iabdalkader:
Note we should probably add a flag to indicate that the vbuffer is invalidated already, if `framebuffer_get_head` is used `FB_PEEK` to avoid invalidating the buffer multiple times, that said currently nothing uses `FB_PEEK` with `framebuffer_get_head`.

iabdalkader:
@kwagyeman updated.

kwagyeman:
This doesn't fix the issue. framebuffer_get_head() returns a buffer after it's been written to by MDMA. So, any writes from the CPU could have been written back by the cache when it makes space for more things while MDMA is updating the buffer.

You have to invalidate before MDMA starts writing to the buffer.

kwagyeman (approved):
Yes, it's probably needed. You can make it more efficient though by putting the code into sensor.c and parsing the image type to only invalidate the image versus the whole buffer.

kwagyeman:
Do invalidates here: https://github.com/openmv/openmv/blob/master/src/omv/ports/stm32/sensor.c#L950

kwagyeman:
@iabdalkader - I don't think this PR is needed. We've been running without it for over a year and I haven't noticed any more frame buffer corruption. Also, since buffers are all invalidated before DMA starts and then invalidated when returning a buffer to the pool for DMA to use they are always invalidated before DMA starts writing to them.

We don't invalidate the vbuffer_t header though. But, since that's on it's own cache line and not read by DMA it doesn't need to be.

kwagyeman:
This PR is needed. The frame buffer must be invalidated before and after. This fix invalidates it after.

iabdalkader:

https://forums.openmv.io/t/openmv-cam-h7-screen-tearing/9686/9

PR #1638, opened 2022-05-15
ports/stm32: Reset CDC buffers after disabling IRQs.

### LINKED ISSUE #1637: Stopping Script in the middle of printing crashes firmware.
Run this:

```
# Hello World Example
#
# Welcome to the OpenMV IDE! Click on the green run arrow button below to run the script!

import sensor, image, time

sensor.reset()                      # Reset and initialize the sensor.
sensor.set_pixformat(sensor.RGB565) # Set pixel format to RGB565 (or GRAYSCALE)
sensor.set_framesize(sensor.QVGA)   # Set frame size to QVGA (320x240)
sensor.skip_frames(time = 2000)     # Wait for settings take effect.
clock = time.clock()                # Create a clock object to track the FPS.

while(True):
    clock.tick()                    # Update the FPS clock.
    #img = sensor.snapshot()         # Take a picture and return the image.
    print(clock.fps())              # Note: OpenMV Cam runs about half as fast when connected

```

Press stop when it starts printing a lot of stuff. On the H7 Plus you crash when this happens. The IDE never gets the USB packet completion from the camera and locks up.

iabdalkader:
* Fixes #1637 for good.
* Reset USB CDC buffers after disabling IRQs, when stopping scripts, to avoid overflows or blocking when printing IDE exception traceback, without losing the exception message.

@kwagyeman Please undo any IDE changes (unnecessary delays) related to that issue.
Also note, it's very likely that this function is useless, because USB IRQs are disabled when this is called:
```c
void usbdbg_wait_for_command(uint32_t timeout)
{
    for (mp_uint_t ticks = mp_hal_ticks_ms(); ((mp_hal_ticks_ms() - ticks) < timeout) && (cmd != USBDBG_NONE); );
}
```

### DIFF
--- a/src/omv/common/usbdbg.c
+++ b/src/omv/common/usbdbg.c
@@ -41,6 +41,10 @@ static mp_obj_t mp_const_ide_interrupt = MP_OBJ_NULL;
 // These functions must be implemented in MicroPython CDC driver.
 extern uint32_t usb_cdc_buf_len();
 extern uint32_t usb_cdc_get_buf(uint8_t *buf, uint32_t len);
+void __attribute__((weak)) usb_cdc_reset_buffers()
+{
+
+}
 
 void usbdbg_init()
 {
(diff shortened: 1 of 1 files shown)


kwagyeman:
Where is that?

Anyway, I made the delays lower already. The IDE is very strict about feeding commands around stop and start as you previously wanted.

iabdalkader:
> Where is that?

https://github.com/openmv/openmv/commit/d1b6202693997c54ed0419417bef9f219d13b691

PR #1652, opened 2022-05-30
sensors/MT9V0xx: Add support for color MT9V0xx sensors.

kwagyeman:
MT9V0XX sensors with a bayer pattern now work. RCCC sensors are not supported (we can easily support them however by adding the necessary debayring code and etc. RCCC sensors are really only used for automotive applications though).

Note that there's no scaling support with bayer pattern sensors.

pic


kwagyeman:
@iabdalkader - This PR unblocks us selling MT9V034 sensors with the bayer pattern on them.

PR #1660, opened 2022-06-06
imlib: Add Support for Stereo Disparity for the Arduino Portenta.

kwagyeman:
Adds support for this guys stereo camera: https://twitter.com/i/status/1533190596134703104

When called on an a double width image (and image with two camera outputs on either side) it will replace one of the images with the disparity:

`img.stereo_disparity(reversed=True, max_disparity=48, threshold=64)`

_reversed - False to compare left image to right, True to compare right image to left. If this is wrong the disparity map will be bonkers._

_max_disparity - Needs to be larger than the maximum pixel shift between images. Bad artifacts occur if too small. However, this value has a huge impact on the run-time of the algorithm._

_threshold - Making this larger short circuits testing all disparities per pixel. Making this large can cut the run-time down by a half or a third but at the cost of a very flattened disparity map._

example

It can process the above 900x375 image (which is made of two 450x375 images) in **1.5s** on the Arduino Portenta. This is with all the SIMD and etc. speeds ups I could think of. Speed ups like doing histogram matching instead of block matching were tried but they don't work well because of the loss of pixel positions within a patch (if histogram matching did work it provides a 5x speedup). Rearranging the shifting image to make the column add/drop cheaper was tried to but didn't work well either.

Anyway, with a resolution of 320x120 (two 160x120 images) the algorithm should get about 4-5 FPS.

(The block size is 4x4. With 8x8 you get nicer looking outputs. With 16x16 things look very smooth. However, 16x16 can take 30 seconds per frame).

Cool that stereo disparity can be done on the STM32H7, but, not usable for real-time video. Given the code below the algorithm should be ported to verilog and included in whatever FPGA is combining sterero cameras. The algorithm works on a scanline basis making this possible without external DRAM on said FPGA.

### DIFF
--- a/src/omv/Makefile
+++ b/src/omv/Makefile
@@ -94,6 +94,7 @@ SRCS += $(addprefix imlib/,     \
 	selective_search.c          \
 	sincos_tab.c                \
 	stats.c                     \
+	stereo.c                    \
 	template.c                  \
 	xyz_tab.c                   \
 	yuv.c                       \
--- a/src/omv/boards/ARDUINO_NANO_RP2040_CONNECT/imlib_config.h
+++ b/src/omv/boards/ARDUINO_NANO_RP2040_CONNECT/imlib_config.h
@@ -150,4 +150,7 @@
 // Enable STM32 DMA2D
 //#define IMLIB_ENABLE_DMA2D
 
+// Stereo Imaging
+// #define IMLIB_ENABLE_STEREO_DISPARITY
+
 #endif //__IMLIB_CONFIG_H__
--- a/src/omv/boards/NANO33/imlib_config.h
+++ b/src/omv/boards/NANO33/imlib_config.h
@@ -150,4 +150,7 @@
 // Enable STM32 DMA2D
 //#define IMLIB_ENABLE_DMA2D
 
+// Stereo Imaging
+// #define IMLIB_ENABLE_STEREO_DISPARITY
+
 #endif //__IMLIB_CONFIG_H__
--- a/src/omv/boards/NICLAV/imlib_config.h
+++ b/src/omv/boards/NICLAV/imlib_config.h
@@ -153,4 +153,7 @@
 #define IMLIB_ENABLE_PNG_ENCODER
 #define IMLIB_ENABLE_PNG_DECODER
 
+// Stereo Imaging
+// #define IMLIB_ENABLE_STEREO_DISPARITY
+
 #endif //__IMLIB_CONFIG_H__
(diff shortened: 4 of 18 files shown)


iabdalkader:
I'll approve this but I'm not sure if we should enable/disable features by default for this board, since you normally can't buy it anywhere.

> Given the code below the algorithm should be ported to verilog and included in whatever FPGA is combining sterero cameras. The algorithm works on a scanline basis making this possible without external DRAM on said FPGA.

In addition to this, can we configure the FPGA from the H7 ? It would be very cool to be able to load bitstreams to the FPGA from ROM the same way firmware blobs are loaded to some sensors. If it's on an I2C bus it can be detected as a special image sensor, add a driver for it, which loads a bitstream that configures it to do the SAD, then it should output the depth map at 60FPS :) we can also get more creative with ioctls, for example an IOCTL to load a custom bitstream.

BurenkovS:
> The algorithm works on a scanline basis making this possible without external DRAM on said FPGA.

I'm using Gowin GW2AR fpga, it contain SDRAM inside, which is opening many interesting features to implement.



> In addition to this, can we configure the FPGA from the H7 ?

It is theoretically possible. JTAG port of FPGA connected to SPI port of arduino, and arduino have JTAG library for configure Intel Cyclone 10 fpga on Arduino MKR Vidor 4000 board. I'm sure I can do it with Gowin FPGA too.

iabdalkader:
> It is theoretically possible. JTAG port of FPGA connected to SPI port of arduino, and arduino have JTAG library for configure Intel Cyclone 10 fpga on Arduino MKR Vidor 4000 board. I'm sure I can do it with Gowin FPGA too.

@BurenkovS Maybe we could add support for loading the bitstream via software (bitbanged) JTAG, but it seems these FPGAS also support SPI, Serial and other download modes, which would be much easier to support:

image

We could then add a new sensor driver for it, but it will have to detect this sensor board somehow, this is normally done via I2C address, maybe if you change the default I2C address pins. Anyway, then we can implement `ioctl`s to load custom bitstreams (that do different image processing). We could also drop the external flash (if there's any) and make the driver load the bitstream on sensor init.

That's a side feature though, mainly it would be great if you can make it send depth data (scaled from 0->255) directly instead of the pixels or maybe GS+DEPTH so 2BPP ? We can add support for new pixel formats if needed.

BurenkovS:
> We can add you to our slack, @kwagyeman

it will be great. my email is sergey.a.burenkov@gmail.com

iabdalkader:
@BurenkovS done, I removed your email for privacy.

kwagyeman:
Hi, I'll fix up this code to be ready for merging into our source tree. It will not be enabled by default. But, this will make it available for reference.

kwagyeman:
This is ready to merge.

PR #1715, opened 2022-09-05
boards/OPENMVPT: Use IR rather than color image for temperature stats.

dennisss:
Before this change, the mean temperature of a blob is calculated based on the color values of the image. When running on the Pure Thermal this has the effect of labeling every blob with the min temperature of the entire scene. This change fixes this bug by using the IR image instead of the color image to get the right value.


iabdalkader:
Hi, thank you for this PR. I will wait for @kwagyeman to take a look since he wrote that script, but in the meantime could you please amend the commit message and PR title to `boards/OPENMVPT` ? This way in the change log it will show the board name.

kwagyeman:
This is right. Wow. I guess I wasn't checking the temp so it looked right.

dennisss:
Thanks for the review. Commit/PR messages fixed.

PR #1716, opened 2022-09-06
scripts/examples: Fix MAVLINK apriltag example.

shiv-tyagi:
This PR brings few changes to the MAVLINK apriltag landing target example after testing with [ArduPilot](http://ardupilot.org/) firmware while working on a project. The changes are:
- Remove the DISTANCE_SENSOR message stuff since it is not required and the distance can be sent within LANDING_TARGET message
- Update the sensor height and width as per the latest OV5650 sensor specifications
- The translation function doesn't seem to be right. The distance values received from the existing translation function are not near to the real values. After this fix, the distance values are quite accurate.
- Add a flashing led to indicate if the target is in sight. The LED flashes red when the camera can't see a target and green when the camera can see it.

### DIFF
--- a/scripts/examples/07-Interface-Library/02-MAVLink/mavlink_apriltags_landing_target.py
+++ b/scripts/examples/07-Interface-Library/02-MAVLink/mavlink_apriltags_landing_target.py
@@ -13,7 +13,6 @@ uart_baudrate = 115200
 
 MAV_system_id = 1
 MAV_component_id = 0x54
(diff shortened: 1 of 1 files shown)


kwagyeman:
Cool, please fix the commit message to pass. See other commits for format.

iabdalkader:
I fixed the commit/PR messages, squashed and merged. Thanks!

rmackay9:
Great to see this, thanks!

PR #1726, opened 2022-09-26
imlib: Fix Draw Image Area Scaling.

kwagyeman:
Makes area scaling work for scaling from 0.99 -> 0.01. Tried all 100 scale factors.

I noticed this bug while fixing image roi scaling.


kwagyeman:
@iabdalkader - This is ready to merge.

Issue #1655, opened 2022-06-02
Failed to run OpenMV example on Arduino Nicla Vision


tejalbarnwal:
**Describe the bug**
I was trying to run the image_drawing_with_custom_palette.py from the examples in the OpenMV IDE. The script does start to execute, but in the middle of the execution, Nicla Vision gets disconnected from the IDE. All other examples under Drawing tab under examples are running fine. Its just this script. What could be the reasons for the same? I did check the USB connection, with Nicla Vision it is working fine.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Examples in OpenMV IDE
2. Click on OpenMV IDE &rarr; Drawing &rarr; `image_drawing_with_custom_palette.py`

The output shown:
![image](https://user-images.githubusercontent.com/64950661/171593183-9511f781-2de9-4e80-870f-c4fd676ee086.png)

Update: After I connect it again to the laptop, there is a new file in the internal storage named ERROR.LOG. It says the following:
```
FATAL ERROR:
MemManage
```
**Please complete the following information**
 - IDE version: 2.9.2
 - Firmware version: 4.3.1
 - Board: Arduino Nicla Vision
 - OS: Windows

A query regarding the same have been put up [here](https://forums.openmv.io/t/failed-to-run-openmv-example/7172) on the OpenMV Forum too

kwagyeman:
Hi, this crashes on all cameras. Debugging now.

PR #1729, opened 2022-09-26
imlib: Fix macro expansion in filter statement.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/1655

Note... this same code with the example used to work. The GCC update changed macro expansion behavior.

### DIFF
--- a/src/omv/imlib/filter.c
+++ b/src/omv/imlib/filter.c
@@ -197,7 +197,8 @@ void imlib_mean_filter(image_t *img, const int ksize, bool threshold, int offset
             }
 
             // Copy any remaining lines from the buffer image...
-            for (int y = IM_MAX(img->h - ksize, 0), yy = img->h; y < yy; y++) {
+            int lines = img->h - ksize, min_lines = IM_MAX(lines, 0);
+            for (int y = min_lines, yy = img->h; y < yy; y++) {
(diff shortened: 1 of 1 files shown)


iabdalkader:
No this has nothing to do with gcc (which hasn't changed since forever), and the macro is used everywhere, if it's broken it means a lot of other code is broken and it should be fixed. I think it's because `img->w/h` are unsigned and `ksize` is signed, the result (say -1) just wraps around and so is bigger than 0.

iabdalkader:
Everywhere where this macro is used like that should be reviewed, for example in line.c:
```
IMAGE_GET_GRAYSCALE_PIXEL_FAST(row_ptr, IM_MAX(x - 1, 0));
```

iabdalkader:
> I was thinking that but img->h is an int32.

No image->h/w have always been `uint32_t`.

kwagyeman:
Huh, okay, you are right. I hovered over the image and saw it say in32_t with the vscode explorer. Weird.

iabdalkader:
> Will find all the places where there's something like this and include them in the PR.

Don't I found an easier way to detect this, will send a PR to fix.

iabdalkader:
Closing because: Superseded by  #1738

Issue #1751, opened 2022-11-24
Making Predictions on arrays with tflite model


BolajiLawal:
**Describe the bug**
I am currently sending data from a wireless sensor to the openmv using uart. The idea is to use a pretrained tflite model to make predictions on the data from the sensor which will be stored as an array. The idea is to use the openmv to take a snapshot if the output from the prediction meets a certain condition/threshold. However, there is currently no way to use tflite model to make predictions on array

**Please complete the following information**
 - IDE version: 2.9.7
 - Firmware version: 4.3.2
 - Board: OPENMV-H7+
 - OS: Windows

chitraa:
Hey since I have a similar requirement, I am open adding this support to openmv and raise a PR. 
Would need some direction @iabdalkader

BolajiLawal:
@chitraa @iabdalkader can you please explain to me how the tf.regression works. Is it possible to set the precision of the output i.e. if I want to 4 decimal places or something. If not what's the default precision, if I get an output of 0.998 will this be approximated to 1?

kwagyeman:
It's a floating point number. The precision is speced by the print statement.

BolajiLawal:
Thanks, so i have a tflite neural network model where the output layer is softmax i.e. should give an array of probabilities. My feeling is if have an input 1D array and use tf.regression, the output 1D array should be an array indicating the probability of the input belonging to each class in the model. But when I try with two different inputs, I get the same answer an outpur array of [0,1]. This is wrong because the examples belong to different classes so should give a different answer. Is there a way I can use tf.regression for this problem or am I doing something wrong?

Issue #1739, opened 2022-10-07
Feature Request: Add support for linear data ML models


tejalbarnwal:
The issue is related to the [this](https://forums.openmv.io/t/running-a-non-linear-regression-model-on-processed-data/7550/9) discussion on the forum.

iabdalkader:
This issue has been mentioned on **OpenMV Forums**. There might be relevant details there:

https://forums.openmv.io/t/running-a-non-linear-regression-model-on-processed-data/7550/10

tejalbarnwal:
Hey, I am not able to find the `libtf_builtin_models.h` file neither in this repository or `openmv/tensorflow-lib` repo. Could you please help me?

iabdalkader:
> find the `libtf_builtin_models.h` file

This file is auto-generated, also my reply above is automated.

tejalbarnwal:
Hey @iabdalkader @kwagyeman , 
I implemented the function myself in the `py_tf.c`, but now I want to debug it. Is there any way to debug the code without running it on OpenMV IDE.
The code I am running right now doesn't involve any input from the development board I am using.
```python
import time, os, tf

labels, net = tf.load_builtin_model('force_quant')

print("input height:", tf.input_height(net))
print("input width:", tf.input_width(net))
print("input channels:", tf.input_channels(net))
print("input datatype:", tf.input_datatype(net))
print("input scale:", tf.input_scale(net))
print("input zero point:", tf.input_zero_point(net))

print("output height", tf.output_height(net))
print("ouput width", tf.output_width(net))
print("ouput channels", tf.output_channels(net))
print("ouput datatype", tf.output_datatype(net))
print("op scale:", tf.output_scale(net))
print("op zero point:", tf.output_zero_point(net))

input_list = [-8, -5, -8,  2, -7, 10, -4, 12, -4, -2, -4,  7, -3, 14,  1, -6,  1, 3,  2, 11,  2, 13,  6, -2,  8,  7,  7, 13,  7, -5, 10,  2, 11,  9, 9, 11]

output = tf.regression(net, 1, input_list)
print(output)
```
I just wanna debug the `tf.regression` function that I implemented. I am currently using `printf` statements and building the firmware and deploying it each time.
Is there any easier way to debug it that you know? maybe running the whole thing on linux machine instead of deploying the firmware everytime on the board?

iabdalkader:
Yes, you can use jlink to debug (build the firmware with `DEBUG=1`).

[...]

PR #1762, opened 2022-12-28
pytf: Added python bindings for regression

tejalbarnwal:
The PR is about feature requested for https://github.com/openmv/openmv/issues/1751 and https://github.com/openmv/openmv/issues/1739

Added binding as `py_tf_regression` to support the functionality of regression models for one-dimensional input and one-dimensional output for the function `libtf_regression_1Dinput_1Doutput` added in `libtf.h`
Presently, it supports only `int8`, `uint8`, and `float32` tflite models.

The following are details about how to use the model:
- Input is taken as an `array` object with type float. Internally, the inout gets mapped to appropriate datatype of `int8`, `uint8`, and `float32`
- Output is returned as a list 
Screenshot from 2022-12-28 11-38-13

### DIFF
--- a/src/omv/modules/py_tf.c
+++ b/src/omv/modules/py_tf.c
@@ -8,10 +8,13 @@
  *
  * Python Tensorflow library wrapper.
  */
+#include <stdio.h>
 #include "py/runtime.h"
 #include "py/obj.h"
 #include "py/objlist.h"
 #include "py/objtuple.h"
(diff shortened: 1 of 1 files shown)


iabdalkader:
Hi, I merged your changes here https://github.com/openmv/openmv/pull/1763 (with some minor formatting fixes) along with the static library updates. Thanks!

PR #1781, opened 2023-02-12
boards/NICLAV: Add readout control to driver and fix image orientation.

kwagyeman:
Adds readout control to the NICLA so you can get a higher FPS: https://youtu.be/R5YjhI1aA5s

You can overclock the sensor in BAYER mode and double the FPS by setting the PLL at 0xF8 to 0x8B.

Also, fixes the image being upside down by default.


iabdalkader:
Tested, and merged.

PR #1802, opened 2023-03-11
sensors/MT9M114: Optimize patch loading.

iabdalkader:
* Optimize and simplify patch loading by converting firmware patches to byte arrays and stashing patch address at the start of the array. This allows loading patches with just one call to cambus_write_bytes, vs the old 1K short REVs and 1K calls to cambus_write_bytes. Since patch loading is done on every reset at 100KHz, this saves a lot of overhead on script run.


kwagyeman:
Seems to be fine. Tested it on the H7 Plus.

PR #1822, opened 2023-04-15
imlib: Add support for AWB and CCM.

kwagyeman:
This PR adds support for some ISP algorithms which allow use to get valid color images out of the colored version of the Global Shutter Image sensor to unblock customer applications. It also can be used to improve the bayer pattern image coming from any sensor we sell.

.awb() performs the grayworld auto white balance algorithm on either an RGB565 image OR the RAW BAYER pattern itself. The code is incredibly fast and takes full advantage of SIMD double pumping.

You may also pass "max=True" to the .awb() method to have to perform the "white patch" algorithm. This algorithm works well in places the grayworld algorithm struggles with. However, by default it produces worse results.

Additionally, I also added color matrix correction support via the .ccm() method. It works like so:

```
import sensor, image, time

sensor.reset()
sensor.set_pixformat(sensor.BAYER)
sensor.skip_frames(time = 2000)

clock = time.clock()

# This matrix is multiplied per pixel:
#    1x3         1x4         4x3 
# [R, G, B] = [r, g, b, 1][1, 0, 0,
#                          0, 1, 0,
#                          0, 0, 1,
#                          0, 0, 0]
# You may also do:
#    1x3         1x3      3x3 
# [R, G, B] = [r, g, b][1, 0, 0,
#                       0, 1, 0,
#                       0, 0, 1]
mat = [1, 0, 0,
       0, 1, 0,
       0, 0, 1,
       0, 0, 0]

while(True):
    clock.tick() 
    img = sensor.snapshot()
    img.awb()
    img.to_rgb565()
    img.ccm(mat)
```

So, it's a generic 3x3 or 4x3 multiplication across every pixel in the image. You can use this to cancel out cross talk between color channels and etc. However, it is not needed to improve the image of the MT9V034 image sensor. awb() is all that is required to get a valid good image from the color global shutter sensor.

Note, the ccm() method only works on an RGB565 image.


kwagyeman:
https://youtu.be/QRHFdJMP8Y8

iabdalkader:
I just want one change, need that new file completely guarded with `#ifdef IMLIB_ENABLE_AWB` and `#ifdef IMLIB_ENABLE_CCM` if you want to split them into two features. It's not enough to let the linker remove deadcode because sometimes some files fail to build on certain platforms. If there's a dependency on `MATH_OPS` you should start detecting this in `imlib_config.h`, ex:
```C
#if defined(IMLIB_ENABLE_AWB) && !defined(IMLIB_ENABLE_MATH_OPS)
#error "AWB requires MATH_OPS"
#endif
```

kwagyeman:
Done. I guarded it with the math ops.

kwagyeman:
Moved it to a new type called ISP ops.

PR #1824, opened 2023-04-18
imlib/apriltag: Add floating point centroid.

kwagyeman:
Adding per: https://forums.openmv.io/t/accuracy-of-april-tags-position/6572/19


iabdalkader:

https://forums.openmv.io/t/accuracy-of-april-tags-position/6572/22

PR #1896, opened 2023-07-12
fix: use urequests provided by micropython-lib

### LINKED ISSUE #1895: Memory leak when using urequests on Nicla vision
**Describe the bug**
I found a memory leak when using the included version of urequets. The sixth request in row always fails with "ENOMEM". To prevent this from happening, I would like to free up the resources by calling "close()" on the Response-object, but the version of urequests provided does not implement this function. 

**To Reproduce**
- run any six (successful) get-requests on the arduino nicla vision
- it does not matter if the request is called in a loop or by interrupt.

**Expected behavior**
Either have no need to free allocated memory or implement some function to have the user clean up after use.
Like it is done [here](https://github.com/micropython/micropython-lib/blob/7128d423c2e7c0309ac17a1e6ba873b909b24fcc/python-ecosys/urequests/urequests.py#L10C9-L10C9)

**Workaround**
a quick workaround is using the original code from [here](https://github.com/micropython/micropython-lib/blob/master/python-ecosys/urequests/urequests.py)

 - IDE version: 3.0.3
 - Firmware version: 4.4.2
 - Board: Arduino Nicla Vision
 - OS: Windows 11

hannesrichter:
I think we do not need to reinvent the wheel here: As you already include _micropython-lib_ as a submodule, i only removed the local copy of the file and added the original as frozen py-file to all applicable boards.

I have run all build pipelines, but could only test the resulting firmware on my _nicla vision_ board. 

fixes #1895


iabdalkader:
We're not reinventing the wheel, our `urequests` has some minor fixes to support our WINC1500 driver, but more importantly our version of the library supports posting files (encoding and adding files to requests), which is Not supported by the one in `micropython-lib`. I thought you just wanted to add the missing `close()` function ?

hannesrichter:
You are right, @iabdalkader . I missed your additions regarding file encoding.

iabdalkader:
> You are right, @iabdalkader . I missed your additions regarding file encoding.

Never mind, thanks for the fix. If those changes ever make it to `micropython-lib` we'll switch to it, no point in keeping duplicates, will close this PR now.

PR #1898, opened 2023-07-16
mimxrt: Handle short packets for debug mode in TinyUSB.

kwagyeman:
Mac sends 384 in high speed mode at max over USB-CDC. Never more.


iabdalkader:
May as well just remove the `if (count == bytes)` , also the commit message should start with `misc:...`

kwagyeman:
Done

PR #1905, opened 2023-07-31
boards/OPENMV_RT1060: Fix blinking LED polarity.

kwagyeman:


iabdalkader:
That's not going to fix the inverted polarity, you need to invert this:

https://github.com/openmv/micropython/blob/openmv-1.20/ports/mimxrt/boards/OPENMV_RT1060/mpconfigboard.h#L28

iabdalkader:
I fixed.

iabdalkader:
Ah I see the issue, this is a Pin object not and LED object, will revert the change. However, that code could use an LED object not sure why I used a Pin.

kwagyeman:
So, change this to a LED object?

iabdalkader:
Does this still work using LED object ? Note you'll need to erase the flash to make sure.

kwagyeman:
Yep! I tested it.

PR #1906, opened 2023-07-31
boards/OPENMV_RT1060: Fix SD Card Detection on power-on.

kwagyeman:
mp_obj_t bdev = MP_OBJ_TYPE_GET_SLOT(&machine_sdcard_type, make_new) (&machine_sdcard_type, 1, 0, args);

Initializes the SD Card pins, clocks, checks CD, and etc. sdcard_detect() should not be called before. sdcard_init_pins() doesn't need to be called directly.

...

SD card is now reliably detected on power on and resets.

### DIFF
--- a/src/omv/ports/mimxrt/main.c
+++ b/src/omv/ports/mimxrt/main.c
@@ -88,6 +88,7 @@ void exec_boot_script(const char *path, bool interruptible) {
 }
 
 int main(void) {
+    bool sdcard_detected = false;
     bool sdcard_mounted = false;
     bool first_soft_reset = true;
 
@@ -139,7 +140,6 @@ soft_reset:
     machine_adc_init();
     #if MICROPY_PY_MACHINE_SDCARD
     machine_sdcard_init0();
-    sdcard_init_pins(&mimxrt_sdcard_objs[MICROPY_HW_SDCARD_SDMMC - 1]);
     #endif
     #if MICROPY_PY_MACHINE_I2S
     machine_i2s_init0();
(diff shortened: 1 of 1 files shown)


iabdalkader:
> SD card is now reliably detected on power on and resets.

Why is SD card detection removed ? The call to `sdcard_detect` ?

kwagyeman:
It's done inside of the make new. As mentioned in my comment you are calling it before the add card peripheral is initialized.

kwagyeman:
machine_sdcard_ioctl -> MP_BLOCKDEV_IOCTL_INIT -> sdcard_detect

kwagyeman:
Card is inited now in the right order.

iabdalkader:
I don't have a debugger, but I think this will reinitialize the SD card & its pins on every soft reset (which also include a 2ms delay), whether there's a card or not, because the card's state is only set to initialized if `ioctl` is called (which in turn calls `sdcard_power_on`).

`machine_sdcard_init_helper` calls both `sdcard_init` and `sdcard_init_pins` (in that order), maybe we can just call those manually on first soft-reset ? Then CD should work (because the clock is enabled & pins are configured)  and if there's a card then call `make_new`. Alternatively, maybe we can simply just configure CD pin as GPIO and check it ?

kwagyeman:
Best fix would be to call:

machine_sdcard_init_helper

They just made it static so I was trying to avoid calling it as I have to edit the MP submodule. Will update PR with this.

kwagyeman:
Wait, this isn't true. machine_sdcard_init_helper is gated by a check to see if the card is already inited.

The current PR is fine.

kwagyeman:
For clarity, makenew does the following:

`mp_map_t kw_args;
    mp_map_init_fixed_table(&kw_args, n_kw, all_args + n_args);

    // Parse args
    mp_arg_val_t args[MP_ARRAY_SIZE(sdcard_init_allowed_args)];
    mp_arg_parse_all(n_args, all_args, &kw_args, MP_ARRAY_SIZE(sdcard_init_allowed_args), sdcard_init_allowed_args, args);

    // Extract arguments
    mp_int_t sdcard_id = args[SDCARD_INIT_ARG_ID].u_int;

    if (!(1 <= sdcard_id && sdcard_id <= MP_ARRAY_SIZE(mimxrt_sdcard_objs))) {
        nlr_raise(mp_obj_new_exception_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("SDCard(%d) doesn't exist"), sdcard_id));
    }

    mimxrt_sdcard_obj_t *self = &mimxrt_sdcard_objs[(sdcard_id - 1)];

    // Initialize SDCard Host
    if (!sdcard_state_initialized(self)) {
        machine_sdcard_init_helper(self);
    }
    return MP_OBJ_FROM_PTR(self);`

iabdalkader:
> but otherwise it only touches the hardware and inits things once.

Not unless `sdcard_power_on` is called from `ioctl` I already mentioned that, please read my comment.

kwagyeman:
Ah, you are right. Mmm, was stating my initial assumption.

Mmm...

kwagyeman:
I think this is the best solution:

`Alternatively, maybe we can simply just configure CD pin as GPIO and check it ?`

Will change to it.

kwagyeman:
@iabdalkader Updated. I think it's what you want now.

PR #1909, opened 2023-08-12
port/imxrt: fix compile error when disable sdram

kaidegit:


iabdalkader:
Hi, the commit message should be "ports/imxrt: Fix compile error when disabling sdram."

PR #1913, opened 2023-08-24
ports/mimxrt: Add machine.CAN driver.

kwagyeman:
After https://github.com/openmv/micropython/pull/114 is merged this enables CAN support for our board.


iabdalkader:
CAN is enabled now.

PR #1914, opened 2023-09-02
imlib/fmath: Fix invalid fp instruction usage on cortex-m4.

kwagyeman:
vcvtm and vcvtp are invalid instructions on the Cortex-M4. It crashes immediately when it executes them.

__ARM_FP is 4 on the cortex-m4 and should be 14 on the cortex-m7

... that said, in practice, something is overriding this output in our make files causing __ARM_FP to always be 4. It appears that passing the argument of the fp unit to GCC causes it to generate the wrong output.

```
kwagyeman@DESKTOP-EF0ULG9:~/GitHub/openmv2/src$ arm-none-eabi-gcc -mcpu=cortex-m7 -mabi=aapcs-linux -mthumb -mfloat-abi=hard -E -dM - < /dev/null | egrep -i 'thumb|arm|cortex|fp[^-]|version|abi' | sort | grep __ARM_FP
#define __ARM_FP 14
```

```
kwagyeman@DESKTOP-EF0ULG9:~/GitHub/openmv2/src$ arm-none-eabi-gcc -mcpu=cortex-m4 -mabi=aapcs-linux -mthumb -mfloat-abi=hard -E -dM - < /dev/null | egrep -i 'thumb|arm|cortex|fp[^-]|version|abi' | sort | grep __ARM_FP
#define __ARM_FP 4
```

But, when you pass the floating point unit:

```
kwagyeman@DESKTOP-EF0ULG9:~/GitHub/openmv2/src$ arm-none-eabi-gcc -mcpu=cortex-m7 -mfpu=fpv5-sp-d16 -mabi=aapcs-linux -mthumb -mfloat-abi=hard -E -dM - < /dev/null | egrep -i 'thumb|arm|cortex|fp[^-]|version|abi' | sort | grep __ARM_FP
#define __ARM_FP 4
```

```
kwagyeman@DESKTOP-EF0ULG9:~/GitHub/openmv2/src$ arm-none-eabi-gcc -mcpu=cortex-m4 -mfpu=fpv4-sp-d16 -mabi=aapcs-linux -mthumb -mfloat-abi=hard -E -dM - < /dev/null | egrep -i 'thumb|arm|cortex|fp[^-]|version|abi' | sort | grep __ARM_FP
#define __ARM_FP 4
```

Not really sure if we should remove passing the floating point unit to GCC. But, this PR fixes the crash on the M4 in the mean time.

Other notes... removing passing the floating point unit adds the following predefines to the M7:

```
__GCC_IEC_559 2 (versus 0 on M4)
__GCC_IEC_559_COMPLEX 2 (versus 0 on M4)
__FP_FAST_FMA 1 (new for M7)
__FP_FAST_FMAL 1 (new for M7)
__FP_FAST_FMAF64 1 (new for M7)
__FP_FAST_FMAF32x 1 (new for M7)
```

My gcc version is `gcc version 10.2.1 20201103 (release) (GNU Arm Embedded Toolchain 10-2020-q4-major)`

Another way to fix this would be to look for the ARM_MATH macro in our header files. But, seems like it will require maintenance per CPU added then.

### DIFF
--- a/src/omv/imlib/fmath.c
+++ b/src/omv/imlib/fmath.c
@@ -8,12 +8,9 @@
  *
  * Fast approximate math functions.
  */
+#include <arm_math.h>
 #include "fmath.h"
 #include "omv_common.h"
-
-#define M_PI      3.14159265f
-#define M_PI_2    1.57079632f
-#define M_PI_4    0.78539816f
 
 const float __atanf_lut[4] = {
     -0.0443265554792128f,    //p7
@@ -55,7 +52,11 @@ float OMV_ATTR_ALWAYS_INLINE fast_sqrtf(float x) {
 int OMV_ATTR_ALWAYS_INLINE fast_floorf(float x) {
     int i;
     asm volatile (
+    #if (__CORTEX_M > 4)
         "vcvtm.S32.f32  %[r], %[x]\n"
+    #else
+        "vcvt.S32.f32  %[r], %[x]\n"
+    #endif
         : [r] "=t" (i)
         : [x] "t"  (x));
     return i;
@@ -63,8 +64,18 @@ int OMV_ATTR_ALWAYS_INLINE fast_floorf(float x) {
 
 int OMV_ATTR_ALWAYS_INLINE fast_ceilf(float x) {
     int i;
+    #if (__CORTEX_M > 4)
     asm volatile (
         "vcvtp.S32.f32  %[r], %[x]\n"
+    #else
+    #pragma GCC diagnostic push
+    #pragma GCC diagnostic ignored "-Wstrict-aliasing"
+    uint32_t max = 0x3f7fffff;
+    x += *((float *) &max);
+    #pragma GCC diagnostic pop
+    asm volatile (
+        "vcvt.S32.f32  %[r], %[x]\n"
+    #endif
         : [r] "=t" (i)
         : [x] "t"  (x));
     return i;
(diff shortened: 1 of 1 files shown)


iabdalkader:

https://forums.openmv.io/t/cam-2-in-current-ide/8697/17

iabdalkader:
> But, this PR fixes the crash on the M4 in the mean time.

Doesn't this also disable that code for M7 ?

iabdalkader:
> causing __ARM_FP to always be 4.

This is not the FP version, it's a set of bits.

iabdalkader:
You're right those specific instructions are not available on CM4, but you shouldn't use the FP macro, you should use CMSIS's `__CORTEX_M` (i.e `__CORTEX_M > 4`), and please don't duplicate the whole function just add if/else around the lines that changed...
Note, `VCVT` will round towards zero so it's not the same, if you want you can use `VCVTR` and set the rounding mode in `FPSCR` but it's probably not worth it.

kwagyeman:
Made the changes requested. Code now properly enables the fast instructions on the M7.

Regarding the rounding up. I just used the format of what was before this code was refactored. However, I did update the floating point value to match the largest value you can specify: 0x3f7fffff which is 0.999999940395355224609375. I wrote it as a hex number so there's no chance that it gets interpreted as 1 by the compiler. Got the value from here: https://www.h-schmidt.net/FloatConverter/IEEE754.html

PR #1918, opened 2023-09-07
drivers: Make all spi based py modules cross platform.

kwagyeman:
This PR is kinda of big. It ended up that way since it touches a lot of modules and a lot of boards. Let me know if it needs to be broken up. Not quite sure how though.

Anyway, most of it is pretty straight forward on using the omv_spi module to replace what wasn't cross platform before. That said:

1. I wanted to avoid changing the vospi bus to avoid making this commit larger. However, the single buffer callback via spi doesn't work with the FLIR Lepton 3. The processor isn't able to keep up. The code for that worked before when you wrote out the entire frame to 64KB... but, not per line. So, I updated the code to use dual buffers like the py_lepton_flir code. This also had the benefit of making the transfer 16-bits per word so it reduces the overhead on the CPU/system-bus.

2. I was able to leverage your .dma_section code to remove the need for the 1K allocations for the flir pin-pong buffers. This is why linker files are changed.

3. To be more compatible with the imxrt micropython base I switched dma control over to their dma_manager file which maintains a free pool of DMA channels. This removes the need to define them statically. In the event there is ever more than one DMA module and DMAMUX the dma_manager should provide that info. In the mean-time... I just made DMA and DMAMUX static in the omv_spi code for the imxrt.

4. For the Arduino Portenta and Nicla Vision I fully speced out all the pins for the external SPI bus and LCD shield pin control.

There's still a lot of work to do to make cross platform modules for PWM, DAC, LCD control to remove the specific parts of that in these modules. However, it can be done later as that code is #ifdef'ed out.

...

Not sure what's going on with the code formatting check. Looks like it can't handle file renames.


iabdalkader:
> Let me know if it needs to be broken up

Yes it does, please break it into multiple PRs with multiple commits, FB changes, hal changes for (stm32, imxrt), module changes for stm32 port, then module changes for imxrt, in its current state I can't sift through all of those changes at once.

> Not sure what's going on with the code formatting check. Looks like it can't handle file renames.

I noticed, working on it, but could use some help debugging it, so for now please leave everything as it is (this PR, your branch).

kwagyeman:
Rebased and pushed.

kwagyeman:
Okay, split it up... might be possible to make it more atomic.

[drivers: Upgrade SPI driver to support py drivers on all boards.](https://github.com/openmv/openmv/pull/1918/commits/991526043cef52a05e6938ddd958cad864812d8a)

This commit is what upgrades the SPI bus drivers. Since the header file is shared for both ports I have to update all the ports at the same time to make the code base still compile. Additionally, since the vospi driver leveraged features that I changed around I had to include those changes too.

This relates to my initial comment: "I wanted to avoid changing the vospi bus to avoid making this commit larger. However, the single buffer callback via spi doesn't work with the FLIR Lepton 3. The processor isn't able to keep up. The code for that worked before when you wrote out the entire frame to 64KB... but, not per line. So, I updated the code to use dual buffers like the py_lepton_flir code. This also had the benefit of making the transfer 16-bits per word so it reduces the overhead on the CPU/system-bus."

[ports/stm32: Update py drivers with new portable spi bus code.](https://github.com/openmv/openmv/pull/1918/commits/2a861cc471955db518d2d5399cf6c577a9229da0)

This commit moves all the SPI buses LCD/TV/FLIR drivers to the shared module folder, fixes the drivers up for usage with the new SPI bus code, and fixes all the pins in the board config files.

This relates to my initial comment: "I was able to leverage your .dma_section code to remove the need for the 1K allocations for the flir pin-pong buffers. This is why linker files are changed."

[ports/mimxrt: Enable support for all py modules.](https://github.com/openmv/openmv/pull/1918/commits/0a054d7e549d6c898c3b7a2198b78f821d9a8b05)

This commit enables things in the imxrt port now that all the heavy lifting has been done.

src/omv/ports/mimxrt/mimxrt_hal.c
```diff
@@ -216,7 +230,7 @@ int mimxrt_hal_spi_init(uint32_t bus_id, bool nss_enable, uint32_t nss_pol) {
     if (nss_enable) {
         omv_gpio_config(spi_pins.ssel_pin, OMV_GPIO_MODE_ALT, OMV_GPIO_PULL_UP, OMV_GPIO_SPEED_MED, -1);
     } else {
-        omv_gpio_config(spi_pins.ssel_pin, OMV_GPIO_MODE_OUTPUT, OMV_GPIO_PULL_UP, OMV_GPIO_SPEED_MED, -1);
+        omv_gpio_config(spi_pins.ssel_pin, OMV_GPIO_MODE_OUTPUT, OMV_GPIO_PULL_UP, OMV_GPIO_SPEED_MED, 5);
```
iabdalkader, line 233: This is wrong please revert it.

src/omv/ports/mimxrt/omv_spi.c
```diff
@@ -78,65 +107,118 @@ static void spi_master_callback(LPSPI_Type *base, void *handle, status_t status,
...
+    uint32_t flags = (OMV_SPI_XFER_DMA | OMV_SPI_XFER_COMPLETE);
+    if ((spi->dma_flags & OMV_SPI_DMA_DOUBLE) && ((spi->xfer_flags & flags) == flags)) {
+        if (spi->xfer_flags & OMV_SPI_XFER_HALF) {
+            spi->xfer_flags &= ~OMV_SPI_XFER_HALF;
+            if (spi->xfer_descr.txData) {
+                spi->xfer_descr.txData -= spi->xfer_descr.dataSize;
+            }
+            if (spi->xfer_descr.rxData) {
+                spi->xfer_descr.rxData -= spi->xfer_descr.dataSize;
+            }
```
iabdalkader, line 126: Just less code...

```suggestion
uint32_t flags = (OMV_SPI_XFER_DMA | OMV_SPI_XFER_COMPLETE);
if ((spi->dma_flags & OMV_SPI_DMA_DOUBLE) && ((spi->xfer_flags & flags) == flags)) {
    int32_t offset = (spi->xfer_flags & OMV_SPI_XFER_HALF) ? -spi->xfer_descr.dataSize : spi->xfer_descr.dataSize;
    if (spi->xfer_descr.txData) {
        spi->xfer_descr.txData += offset;
    }
    if (spi->xfer_descr.rxData) {
        spi->xfer_descr.rxData += offset;
    }
    spi->xfer_flags ^= OMV_SPI_XFER_HALF;
}
```

src/omv/ports/mimxrt/omv_spi.c
```diff
@@ -78,65 +107,118 @@ static void spi_master_callback(LPSPI_Type *base, void *handle, status_t status,
...
+        spi->xfer_descr.dataSize /= 2;
+    }
+
+    // Check if we can optimize the transfer by doing things in 32-bit chunks.
+    if ((!(spi->xfer_descr.dataSize % 4)) && (!(xfer->flags & OMV_SPI_XFER_16_BIT))) {
+        omv_spi_set_frame_size(spi, 32);
+        spi->xfer_descr.configFlags |= kLPSPI_MasterByteSwap;
+    }
+
+    if (spi->xfer_flags & OMV_SPI_XFER_BLOCKING) {
```
iabdalkader, line 204: Please revert this to the original order it used, DMA was checked first, it will make it easier to read the diff.

src/omv/ports/mimxrt/omv_spi.c
```diff
@@ -145,46 +227,64 @@ int omv_spi_transfer_start(omv_spi_t *spi, omv_spi_transfer_t *xfer) {
...
     spi->cs = spi_descr->cs;
-    spi->dma_enabled = config->dma_enable;
+    spi->dma_flags = config->dma_flags;
 
     lpspi_master_config_t spi_config;
     LPSPI_MasterGetDefaultConfig(&spi_config);
 
     spi_config.whichPcs = kLPSPI_Pcs0;
     spi_config.baudRate = config->baudrate;
-    spi_config.bitsPerFrame = config->datasize;
```
iabdalkader, line 179: I looked back at our discussion about this PR on slack, and there was definitely a confusion about this, there's no reason to remove datasize, it's definitely good to have it, and it shouldn't be a flags. The discussion was about `dmasize` not datasize.

src/omv/ports/mimxrt/omv_spi.c
```diff
@@ -145,46 +227,64 @@ int omv_spi_transfer_start(omv_spi_t *spi, omv_spi_transfer_t *xfer) {
 }
 
 int omv_spi_transfer_abort(omv_spi_t *spi) {
-    LPSPI_MasterTransferAbortEDMA(spi->inst, &spi->descr_master_edma);
+    if (spi->dma_flags & (OMV_SPI_DMA_NORMAL | OMV_SPI_DMA_CIRCULAR)) {
+        LPSPI_MasterTransferAbortEDMA(spi->inst, &spi->descr_master_edma);
+    }
+    LPSPI_MasterTransferAbort(spi->inst, &spi->descr_master);
+    LPSPI_MasterInit(spi->inst, &spi->config_backup, BOARD_BOOTCLOCKRUN_LPSPI_CLK_ROOT);
```
iabdalkader, line 234: Why do we need to reconfigure the SPI on abort ?

src/omv/ports/mimxrt/omv_spi.c
```diff
@@ -40,36 +39,66 @@ typedef struct omv_spi_descr {
...
-    { NULL, NULL, { NULL, NULL, 0, 0 }, { NULL, NULL, 0, 0 } },
+    { NULL, NULL, { NULL, NULL, 0 }, { NULL, NULL, 0 } },
     #endif
 };
 
+// LPSPI_MasterTransferEDMA doesn't have any code at all to generate a callback on transmit complete. So, we
+// add one here that simulates that. However, the DMA interrupt will occur before the SPI bus is done sending
+// data. So, we cannot call spi_master_callback below as it will happen before the data is actually sent. Instead,
+// we use the SPI transfer complete interrupt to call spi_master_callback at the right time. Note that we have
+// to zero out the descriptor to prevent LPSPI_MasterTransferHandleIRQ from doing anything we don't want it to do.
```
iabdalkader, line 77: I'm not following, it seems it does have code to call back on transmit complete and this is it (transferDone will be set to true), but why do we need this ?

iabdalkader (changes requested):

src/omv/ports/mimxrt/mimxrt_hal.c
```diff
@@ -87,10 +88,7 @@ void mimxrt_hal_init() {
                          );
     #endif
 
-    // Confgure and enable EDMA
-    edma_config_t edma_config = {0};
-    EDMA_GetDefaultConfig(&edma_config);
-    EDMA_Init(DMA0, &edma_config);
+    dma_init();
```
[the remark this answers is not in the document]
kwagyeman: Per our discussion via slack you where okay using the dma poll stored in micropython so that we could share handles with the MicroPython drivers which will use this. I reverted the DMA0/DMAMUX stuff as asked though.

src/omv/ports/mimxrt/omv_portconfig.h
```diff
@@ -46,6 +46,7 @@
 #define OMV_GPIO_MODE_IT_FALL   ((6 << 17) | IOMUXC_PAD_CTL(DSE, 0U) | IOMUXC_PAD_CTL(HYS, 1U))
 #define OMV_GPIO_MODE_IT_RISE   ((7 << 17) | IOMUXC_PAD_CTL(DSE, 0U) | IOMUXC_PAD_CTL(HYS, 1U))
 #define OMV_GPIO_MODE_IT_BOTH   ((8 << 17) | IOMUXC_PAD_CTL(DSE, 0U) | IOMUXC_PAD_CTL(HYS, 1U))
+
```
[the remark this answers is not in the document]
kwagyeman: Okay.

src/omv/ports/mimxrt/omv_spi.c
```diff
@@ -40,36 +39,66 @@ typedef struct omv_spi_descr {
...
+    spi->descr_master.writeRegRemainingTimes = 0;
+    spi->descr_master.readRegRemainingTimes = 0;
+    spi->descr_master.txBuffIfNull = 0;
+    spi->descr_master.fifoSize = 0;
+    spi->descr_master.isPcsContinuous = 0;
+    spi->descr_master.isByteSwap = 0;
+    spi->descr_master.bytesEachWrite = 0;
+    spi->descr_master.bytesEachRead = 0;
+    spi->descr_master.rxWatermark = 0;
+    spi->descr_master.isTxMask = 0;
```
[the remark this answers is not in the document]
kwagyeman: I wasn't zeroing all variables in the field. But, yeah, I can store variables that I'm not zeroing, zero the rest, and then restore them to make this smaller.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: Uh, yeah, I mean, you wanted to save space in the struct. I literally only use the value as 8/16 so it's effectively a flag. It makes sense to make it a flag.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: Abort leaves the SPI bus broken as it hits the RESET bit in the register... and after pulling my hair out on why it wasn't working I read the docs and it says that after abort you have to reinit the spi bus.

src/omv/ports/mimxrt/mimxrt_hal.c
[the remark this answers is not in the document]
iabdalkader: Yes only to allocate channels not for initialization, I'd like to rely on as less as possible from the port(s) please keep this in mind.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: They only implemented a callback for RX. There's literally no code in the driver for a TX callback. As for how you are supposed to know when the TX transfer completes... no clue. I read the driver code and it was clearly missing though.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: Okay.

src/omv/ports/mimxrt/mimxrt_hal.c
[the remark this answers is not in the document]
kwagyeman: Mmm, looking at the manager it appears I don't have to call dma_init for it to work. It will allow usage of the free pool without calling dma_init(). However, this could lead to bugs in the future not calling dma_init() ourselves as it will try to reinit the EDMA if called by some other code.

Still revert?

src/omv/ports/mimxrt/omv_spi.c
iabdalkader: I didn't want to remove data-size, SPI can use other modes than 8 and 16, also you added a duplicate of the whole config struct so no point in trying to save space anymore.

src/omv/ports/mimxrt/mimxrt_hal.c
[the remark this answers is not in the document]
iabdalkader: It's only called from i2s code, but it should be okay to reinit the DMA, in fact I prefer if we revert back to hard-coded channels and just set them as used in `dma_manager.c` (i.e. reserve them).

src/omv/ports/mimxrt/omv_spi.c
iabdalkader: Can we try to clear that bit ? If this can be avoided we don't need to duplicate the config, where does it say that in the datasheet ?

src/omv/ports/mimxrt/mimxrt_hal.c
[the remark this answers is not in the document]
kwagyeman: That will require an update of MicroPython too. Should that be in another PR?

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: I tried. I really did. If you abort and don't reset the SPI hardware it breaks. It will not work anymore.

I went as far as implementing all the code in the abort function without the reset... but, no, once you clear the DMA TX/RX registers in the SPI bus you have to reset the whole module.

kwagyeman: It says this about the abort function in the driver api header file.

src/omv/ports/mimxrt/omv_spi.c
iabdalkader: It seems to be called after every transfer (tx or rx), and when it's done transferDone should be set. But I'm still not sure why we need this callback ?

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: It was a last resort thing to include the whole config struct again. I did not want to do that. But, I spent a day trying to do a clean approach and got nowhere as the module will not work after abort unless you reinit it.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: As mentioned, they don't setup the callback for TX in their code. That's why I have to do it manually. They only setup the callback for RX.

src/omv/ports/mimxrt/omv_spi.c
```diff
@@ -78,65 +107,118 @@ static void spi_master_callback(LPSPI_Type *base, void *handle, status_t status,
     } else {
         spi->xfer_flags |= OMV_SPI_XFER_FAILED;
         spi->xfer_error = status;
-        if (spi->xfer_flags & OMV_SPI_XFER_DMA) {
-            LPSPI_MasterTransferAbortEDMA(spi->inst, &spi->descr_master_edma);
+        omv_spi_transfer_abort(spi);
+    }
+
+    void *rxBuf = spi->xfer_descr.rxData;
```
iabdalkader, line 113: I think this should be `rxData` or `txData` otherwise it will be NULL in TX transfer callback, even if it's not used by the callback it shouldn't be null.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: E.g. I have to do this to enable the TX callback.

```
 if (xfer->txbuf) {
            // There isn't a race condition here to worry about since the major interrupt status flag
            // will remain asserted even if the DMA transfer was to complete before we enable the interrupt.
            EDMA_SetCallback(&spi->dma_descr_tx, EDMA_LpspiMasterTxCallback, spi);
            EDMA_EnableChannelInterrupts(spi->dma_descr_tx.base, spi->dma_descr_tx.channel,
                                         (uint32_t) kEDMA_MajorInterruptEnable);
        }
```

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: My intention is that it would be null unless it was an rx callback. I can change that though.

src/omv/ports/mimxrt/omv_spi.c
iabdalkader: >  It says this about the abort function in the driver api header file.

Which file ? Link ?

Do we need to call both `LPSPI_MasterTransferAbortEDMA` and `LPSPI_MasterTransferAbort` for a DMA transfer ? Could that be the issue ?

iabdalkader: This can't be right because in the current driver it works without re-initializing the bus and there's an abort, I've tested this with vospi + FLIR.

src/omv/ports/mimxrt/omv_spi.c
iabdalkader: We seem to be going in circles here... What is the purpose of this callback ? I know it's a TX complete callback, but why do we care when TX is complete ? The callback seems to zero the struct after the TX is done, why is that needed ?

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: Abort calls LPSPI_RESET.

```
/*!
 * brief LPSPI master abort transfer which uses an interrupt method.
 *
 * This function aborts a transfer which uses an interrupt method.
 *
 * param base LPSPI peripheral address.
 * param handle pointer to lpspi_master_handle_t structure which stores the transfer state.
 */
void LPSPI_MasterTransferAbort(LPSPI_Type *base, lpspi_master_handle_t *handle)
{
    assert(handle != NULL);

    /* Disable interrupt requests*/
    LPSPI_DisableInterrupts(base, (uint32_t)kLPSPI_AllInterruptEnable);

    LPSPI_Reset(base);

    handle->state                = (uint8_t)kLPSPI_Idle;
    handle->txRemainingByteCount = 0;
    handle->rxRemainingByteCount = 0;
}
```

Which says:

```
/*!
 * brief Restores the LPSPI peripheral to reset state. Note that this function
 * sets all registers to reset state. As a result, the LPSPI module can't work after calling
 * this API.
 * param base LPSPI peripheral address.
 */
void LPSPI_Reset(LPSPI_Type *base)
{
    /* Reset all internal logic and registers, except the Control Register. Remains set until cleared by software.*/
    base->CR |= LPSPI_CR_RST_MASK;

    /* Software reset doesn't reset the CR, so manual reset the FIFOs */
    base->CR |= LPSPI_CR_RRF_MASK | LPSPI_CR_RTF_MASK;

    /* Master logic is not reset and module is disabled.*/
    base->CR = 0x00U;
}
```

You only abort the DMA in the current code:

```
int omv_spi_transfer_abort(omv_spi_t *spi) {
    LPSPI_MasterTransferAbortEDMA(spi->inst, &spi->descr_master_edma);
    return 0;
}
```

This is not applicable when doing non-DMA transfers.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: See the large 4 line comment above the function.

I have to zero out the descriptor to prevent unwanted behavior from occurring in the driver HAL. Then enable the SPI transfer complete interrupt so they happen in the right order.

The DMA interrupt happens before the SPI bus finishes sending data. So, I can't do anything from it. However, I can enable the transfer complete interrupt for the SPI. Which when it fires will run callback code.

If I don't zero most of the descriptor values then it will have side affects and not just call my the callback. See `void LPSPI_MasterTransferHandleIRQ(LPSPI_Type *base, lpspi_master_handle_t *handle)` in the fsl_lpspi.c file.

src/omv/ports/mimxrt/omv_spi.c
iabdalkader: > This is not applicable when doing non-DMA transfers.

Do we need to recover the bus if a blocking/IT transfer fails ? If so then still it should be :

```c
    } else {
       LPSPI_MasterTransferAbort(spi->inst, &spi->descr_master);
       LPSPI_MasterInit(spi->inst, &spi->config_backup, BOARD_BOOTCLOCKRUN_LPSPI_CLK_ROOT);
}
```

kwagyeman: This might be better.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: Note, you can't use the transfer complete interrupt directly from the SPI since that fires per frame sent. You have to get the transfer complete from the DMA first and then enable the interrupt for the SPI to known the end of the data transfer to the physical wire.

src/omv/ports/mimxrt/mimxrt_hal.c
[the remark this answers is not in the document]
kwagyeman: Reverted the change above. Please let me know if you want me to remove: allocate_dma_channel() from the om_spi_dmi_init() method.

I expect MicroPython will use the dma allocator more in the future. So, we need to reserve the channels in it or use it.

src/omv/ports/mimxrt/omv_portconfig.h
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: Done. Will return a pointer to the transmit buffer if no rx buffer is present.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: Updated comment as requested.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: Regarding this, it doesn't need to be in the config struct anymore. The transfer struct now controls the data size. The code will automatically update the bus/dma as required for the transfer.

For other than 8-bit/16-bit modes. I don't have anything to test with for 32-bit transfers. We don't have any device that needs that right now. Would it makes sense to add support for 32-bit transfers in the future? Or do so now.

src/omv/ports/mimxrt/omv_spi.c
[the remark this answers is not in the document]
kwagyeman: Was able to reduce this to:

```
omv_spi_t *spi = (omv_spi_t *) user;
spi->descr_master_edma.state = (uint8_t) kLPSPI_Idle;
spi->descr_master.rxData = NULL;
LPSPI_EnableInterrupts(spi->inst, (uint32_t) kLPSPI_TransferCompleteFlag);
```

kwagyeman:
Updated the PR per feedback. However, i have not retested everything yet. Do not merge.

src/omv/ports/mimxrt/mimxrt_hal.c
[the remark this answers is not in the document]
iabdalkader: > That will require an update of MicroPython too. Should that be in another PR?

Yes please reserve the channels you use in MicroPython, perhaps even find a way to make it configurable with board config files and push it upstream first. For now I guess you can just hard-code them in dma_manager.c

src/omv/ports/stm32/omv_spi.c
```diff
@@ -152,18 +152,80 @@ static void omv_spi_callback(SPI_HandleTypeDef *hspi) {
...
+    omv_spi_callback_error_check(hspi, spi);
+    spi->xfer_flags &= ~OMV_SPI_XFER_HALF;
+
     if (spi->callback) {
-        spi->callback(spi, spi->userdata);
+        spi->callback(spi, spi->userdata, omv_spi_callback_buf(spi));
+    }
+}
+
+static void omv_spi_half_callback(SPI_HandleTypeDef *hspi) {
```
iabdalkader, line 202: I don't think we need the half callback, or those extra functions, this should work:
```suggestion
static void omv_spi_callback(SPI_HandleTypeDef *hspi) {
    omv_spi_t *spi = omv_spi_get_descr(hspi);
    if (spi == NULL) {
        return;
    }

    if (hspi->ErrorCode != HAL_SPI_ERROR_NONE) {
        spi->xfer_flags |= OMV_SPI_XFER_FAILED;
        spi->xfer_error = hspi->ErrorCode;
        omv_spi_transfer_abort(spi);
    } else {
        spi->xfer_flags |= OMV_SPI_XFER_COMPLETE;
    }

    if (spi->callback) {
        uint8_t *buf = (spi->descr->pRxBuffPtr) ? spi->descr->pRxBuffPtr : spi->descr->pTxBuffPtr;
        if (config->dma_flags & OMV_SPI_DMA_DOUBLE) {
            if (spi->xfer_flags & OMV_SPI_XFER_HALF) {
                uint32_t size = (spi->descr->RxXferSize) ? (spi->descr->RxXferSize) : spi->descr->TxXferSize;
                buf += ((size * ((spi->xfer_flags & OMV_SPI_XFER_16_BIT) ? 2 : 1)) / 2);
            }
            spi->xfer_flags ^= OMV_SPI_XFER_HALF;
        }
        spi->callback(spi, spi->userdata, buf);
    }
}
```

iabdalkader (changes requested):

src/omv/ports/mimxrt/omv_spi.c
iabdalkader: > This is the behavior you wanted.

No it's not, again we were talking about the DMA transfer size, not the transfer width. The transfer width does Not change between transfers, it's a set once and forget config option, it can't change between transfers without re-configuring the SPI and DMA, in fact you shouldn't be allowed to, and if it can't change why pass it as a flag in every transfer ? It's a property of the SPI/DMA config not of individual transfers.

Also if you revert the datasize in the config and add `datasize` to `omv_spi_t` there won't be any need to change the frame size every transfer, and code like this can be removed:

```
    uint32_t datasize = 8;
    if (xfer->flags & OMV_SPI_XFER_16_BIT) {
        datasize = 16;
    }

    omv_spi_set_frame_size(spi, datasize);
```

src/omv/ports/stm32/omv_spi.c
kwagyeman: Done

src/omv/boards/OPENMV_RT1060/omv_boardconfig.h
```diff
@@ -58,6 +58,11 @@
 #define OMV_ENABLE_HM01B0          (0)
 #define OMV_ENABLE_PAJ6100         (1)
 #define OMV_ENABLE_FROGEYE2020     (1)
+#define OMV_ENABLE_FIR_MLX90621    (1)
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/common/vospi.c
```diff
@@ -10,154 +10,184 @@
...
         }
 
-        if ((mp_hal_ticks_ms() - tick_start) >= timeout_ms) {
-            // Timeout error.
-            return -1;
+        if (!waiting_for_data) {
+            break;
         }
 
         MICROPY_EVENT_POLL_HOOK
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/common/vospi.c
```diff
@@ -10,154 +10,184 @@
...
+#define VOSPI_SEGS_PER_FRAME     (4)
+#define VOSPI_PACKET_SIZE        (VOSPI_HEADER_WORDS + VOSPI_PID_SIZE_PIXELS) // 16-bits
+#define VOSPI_SEG_SIZE_PIXELS    (VOSPI_PIDS_PER_SEG * VOSPI_PID_SIZE_PIXELS) // 16-bits
+
+#define VOSPI_BUFFER_SIZE        (VOSPI_PACKET_SIZE * 2) // 16-bits
+#define VOSPI_CLOCK_SPEED        20000000 // hz
+#define VOSPI_SYNC_MS            200 // ms
+
+static int vospi_rx_cb_expected_pid = 0;
+static int vospi_rx_cb_expected_seg = 0;
```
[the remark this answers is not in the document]
kwagyeman: Done, but, I use temporary variables with the short names in the callback. These have longer names on purpose. I will shorten them though.

src/omv/common/vospi.c
```diff
@@ -10,154 +10,184 @@
  */
 #include "omv_boardconfig.h"
 #if OMV_ENABLE_VOSPI || OMV_ENABLE_LEPTON
-
-#include <stdint.h>
 #include "py/mphal.h"
 
-#include "vospi.h"
 #include "crc16.h"
+
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/common/vospi.c
```diff
@@ -10,154 +10,184 @@
...
-} vospi_flags_t;
-
-typedef struct _vospi_state {
-    volatile uint32_t pid;
-    volatile uint32_t sid;
-    uint8_t *buffer;
-    uint32_t n_packets;
-    omv_spi_t spi_bus;
-    volatile uint32_t flags;
-} vospi_state_t;
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/common/vospi.c
```diff
@@ -10,154 +10,184 @@
...
+    for (int i = VOSPI_HEADER_WORDS; i < VOSPI_PACKET_SIZE; i++) {
+        int value = base[i];
+        crc = ByteCRC16(value >> 8, crc);
+        crc = ByteCRC16(value, crc);
+    }
+
+    return packet_crc == crc;
+}
+#endif
+
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/ports/mimxrt/mimxrt_hal.c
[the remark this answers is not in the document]
kwagyeman: Reverted. Will send a separate PR for disabling them in the DMA manager after this PR is merged.

kwagyeman:
Updated the PR per feedback. However, I have not retested everything yet. Do not merge.

src/omv/ports/mimxrt/omv_spi.c
```diff
@@ -78,65 +89,118 @@ static void spi_master_callback(LPSPI_Type *base, void *handle, status_t status,
...
-    // Duplicate & stash transfer in spi struct, to avoid recreating
-    // it for circular transfers. NOTE: If circular transfers ever
-    // works this can be removed.
+
+    uint32_t datasize = 8;
+    if (xfer->flags & OMV_SPI_XFER_16_BIT) {
+        datasize = 16;
+    }
+
+    omv_spi_set_frame_size(spi, datasize);
```
iabdalkader, line 164: Sorry this one is on me, but I thought about this some more and I prefer if we set the transfer/word width at config time, for the following reasons:

1- Writes to `TCR` take a number of SPI clock cycles to update, this will be done on * every * single transfer vs just once at config time.
2- This "feature" might be hard to support for some future port, if it is an issue for some port we'll end adding ugly code here that will deinit/init the SPI bus. I prefer if we keep this common API simple so it makes it easier to port it.
3- Most of the time you don't change the word width between transfers. The only code that does that is the LCD driver, it's a very special case forced on every use case (TV, WINC, PIXART, FLIR, etc..)
4- None of the SDKs/HALs etc support changing the word width in transmit/receive functions, the old code wrote to registers directly, so shouldn't try to support this either.

For the LCD driver, can we just reinit the bus when needed ?

iabdalkader (changes requested):

src/omv/ports/mimxrt/omv_spi.c
```diff
@@ -78,65 +89,118 @@ static void spi_master_callback(LPSPI_Type *base, void *handle, status_t status,
...
+                                         (uint32_t) kEDMA_MajorInterruptEnable);
         }
-
-        // Non-blocking transfetr, return immediately and the user
-        // callback will be called when the transfer is done.
-        if (spi->xfer_flags & OMV_SPI_XFER_NONBLOCK) {
-            return 0;
+    } else if (spi->xfer_flags & OMV_SPI_XFER_BLOCKING) {
+        if (LPSPI_MasterTransferBlocking(spi->inst, &spi->xfer_descr) != kStatus_Success) {
+            return -1;
```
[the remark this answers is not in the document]
kwagyeman: When I was first debugging the code this caused an interrupt storm and crashed the processor. Seems to be working now changing it back this this without issue. Will do.

src/omv/common/vospi.c
```diff
@@ -10,127 +10,155 @@
...
-    VOSPI_FLAGS_RESYNC = (1 << 1),
-} vospi_flags_t;
-
 typedef struct _vospi_state {
-    volatile uint32_t pid;
-    volatile uint32_t sid;
-    uint8_t *buffer;
-    uint32_t n_packets;
+    int expected_pid;
+    int expected_seg;
```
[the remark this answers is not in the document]
kwagyeman: There are temporary variables in the vospi code with the same name. I can change seg to sid though.

src/omv/common/vospi.c
```diff
@@ -10,127 +10,155 @@
...
+    uint16_t *framebuffer;
+    bool lepton_3;
     omv_spi_t spi_bus;
-    volatile uint32_t flags;
+    volatile bool waiting_for_data;
+    volatile bool resync_callback_flag;
```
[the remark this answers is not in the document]
kwagyeman: This is not possible. The flags don't work the same way.

waiting_for_data is a producer to consumer flag
resync_callback_flag is a consumer to producer flag

If they share the same variable then there's a race condition.

src/omv/common/vospi.c
```diff
@@ -10,127 +10,155 @@
...
-                vospi.flags |= VOSPI_FLAGS_RESYNC; // lost sync
-                debug_printf("lost sync, packet id:%lu expected id:%lu \n", pid, vospi.pid);
-            }
-        } else if (vospi.n_packets > VOSPI_NUMBER_PACKETS
-                   && pid == VOSPI_SPECIAL_PACKET && sid != vospi.sid) {
-            if (vospi.sid != VOSPI_FIRST_SEGMENT) {
-                vospi.flags |= VOSPI_FLAGS_RESYNC; // lost sync
-                debug_printf("lost sync, segment id:%lu expected id:%lu\n", sid, vospi.sid);
+    int pid = id & 0x0FFF;
+    int seg = ((id >> 12) & 0x7) - 1;
```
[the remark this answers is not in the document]
kwagyeman: Sure

src/omv/common/vospi.c
```diff
@@ -10,127 +10,155 @@
...
+        vospi.expected_pid = 0;
+
+        bool frame_ready = false;
+
+        // For the FLIR Lepton 3 we have to receive all the pids in all the segments.
+        if (vospi.lepton_3) {
+            vospi.expected_seg += 1;
+            if (vospi.expected_seg == VOSPI_SEGS_PER_FRAME) {
+                vospi.expected_seg = 0;
+                frame_ready = true;
```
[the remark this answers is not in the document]
kwagyeman: Yes, but, waiting for data can't share the same variable space (see previous comment). Hence why I just used bools.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: 1. The driver code itself touches the register and changes it per API call. The only reason I have to touch it ourselves is that their code never updates the frame size. It just starts/stops the transaction. Note that I only write the register and delay in the loop if it's not what it should be.

2. Maybe... I don't see how you avoid the spi driver being super painful to make work. But, yes, if I had choosen to init/deinit per call I wouldn't have had to spend so much time learning why the SPI bus didn't work for random. It would have been way less painful. I can agree that getting down to the low-level details of the SPI device was challenging.

3. Sure, but, a new device may require this. Since we have the code working right now why not keep it? For future ports we can do the less effort init/deinit in the driver itself. I agree it would have saved time.

4. Yeah, but, you'd agree the HAL drivers aren't complete for every situation. We pretty much always have to bypass them for the best performance. E.g. the camera driver.

> For the LCD driver, can we just reinit the bus when needed ?

If we do that it should still be inside of the SPI driver right? Note, the LCD module will have to reinit and deinit the bus before and after every display update. I have to do a single byte write before updating the screen and then a single byte write afterwards.

src/omv/common/vospi.c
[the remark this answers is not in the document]
iabdalkader: > There are temporary variables in the vospi code with the same name

It shouldn't be a problem.

src/omv/ports/mimxrt/omv_spi.c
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/common/vospi.c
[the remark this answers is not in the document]
kwagyeman: Oh, yeah, now that you have the vospi in front you are right. It's fine.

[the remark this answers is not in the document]
kwagyeman: Done

src/omv/common/vospi.c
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/common/vospi.c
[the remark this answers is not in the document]
iabdalkader: A race condition how do you mean?

src/omv/ports/mimxrt/omv_spi.c
iabdalkader: I didn't realize it required switching word width on every frame, but why can't you write the frame as bytes ? Which LCD is this for ?

src/omv/common/vospi.c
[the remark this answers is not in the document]
kwagyeman: The flags work by one thread raising them and then the other clearing them. If you put them in the same variable then you have a read-modify-write situation.

So, if both the producer and consumer were setting/clearing flags at the same time you'd have an undefined result in the variable.

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: You can, but, but, then the processor has to byte reverse every 16-bits. So, it's a ton of CPU overhead. By just writing the data out at 16-bits at a time the processor can just fire off a DMA transaction from the RGB565 line buffer and go on it's way.

This for our normal LCD driver. I'm just leveraging how little-endian reads work to dramatically save CPU time. E.g. if the DMA reads data at 8-bits at a time it byte reverses the data versus if it reads it at 16-bits at a time.

src/omv/common/vospi.c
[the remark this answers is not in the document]
iabdalkader: I see how the flags meaning is inverted now, well this should work:
```C
static void vospi_resync() {
    omv_spi_transfer_t spi_xfer = {
    .....
    vospi.flasg &= ~VOSPI_FLAGS_RESYNC;
}

int vospi_snapshot(uint32_t timeout_ms) {
   vospi.flags = VOSPI_FLAGS_CAPTURE;

    do {
        if (vospi.flags & VOSPI_FLAGS_RESYNC) {
            vospi_resync();
        }

        if ((mp_hal_ticks_ms() - tick_start) > timeout_ms) {
            omv_spi_transfer_abort(&vospi.spi_bus);
            vospi.expected_pid = 0;
            vospi.expected_sid = 0;
            vospi.flasg = VOSPI_FLAGS_RESYNC;
            return -1;
        }

        MICROPY_EVENT_POLL_HOOK
    } while (vospi.flags & VOSPI_FLAGS_CAPTURE);
```

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: Updated the PR per feedback. However, I have not retested everything yet. Do not merge.

kwagyeman:
Updated the PR per feedback. However, I have not retested everything yet. Do not merge.

src/omv/ports/mimxrt/omv_spi.c
iabdalkader: I think the LCD driver had a bit to reverse the order.

kwagyeman: Yes, I removed that years ago when I optimized it.

src/omv/common/vospi.c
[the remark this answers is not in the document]
iabdalkader: > So, if both the producer and consumer were setting/clearing flags at the same time you'd have an undefined result in the variable.

The IRQ doesn't set/clear the RSYNC flag.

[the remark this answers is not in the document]
kwagyeman: Mmm, let me think about it. Since the two threads aren't completely async I think what you are doing works.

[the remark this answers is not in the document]
kwagyeman: Here's one race condition:

The producer could set the VOSPI_FLAGS_RESYNC flag at the same time you set the VOSPI_FLAGS_CAPTURE flag.

Then the driver will lock up.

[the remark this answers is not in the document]
iabdalkader: > The producer could set the VOSPI_FLAGS_RESYNC flag at the same time you set the VOSPI_FLAGS_CAPTURE flag.
> 
> Then the driver will lock up. Either capture wouldn't be high or resync wouldn't be high.

And how does two variables fix that situation ?

Also, that's why the original code that you removed had this:
```c
    if (vospi.flags & VOSPI_FLAGS_RESYNC) {
        // Captured a packet before an resync is complete.
        return;
    }
```

src/omv/ports/mimxrt/omv_spi.c
iabdalkader: >  Yes, I removed that years ago when I optimized it.

Please un-optimize it, so we can fix this issue.

src/omv/common/vospi.c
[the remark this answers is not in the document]
iabdalkader: Okay you can keep the bools, but I think a) they need to be volatile b) please remove `bool frame_ready = false;` and just set it directly.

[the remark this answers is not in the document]
kwagyeman: done

kwagyeman:
Testing Status:

OpenMV Cam M4:
* LCD Module: Pass (no triple buffer though, not enough ram)
* TV Module: Pass (no triple buffer though, not enough ram)
* WiFi Module: Pass (able to stream UDP and TCP RTSP video at QQVGA)
* FIR Module: Pass (flir lepton 1 and flir lepton 3 work)

OpenMV Cam F7:
* LCD Module: Pass (triple buffering and non-buffered work)
* TV Module: Pass (no triple buffer though, not enough ram)
* WiFi Module: Pass (able to stream UDP and TCP RTSP video at QQVGA)
* FIR Module: Pass (flir lepton 1 and flir lepton 3 work)

OpenMV Cam H7:
* LCD Module: Pass (triple buffering and non-buffered work)
* TV Module: Pass (no triple buffer though, not enough ram)
* WiFi Module: Pass (able to stream UDP and TCP RTSP video at QVGA)
* FIR Module: Pass (flir lepton 1 and flir lepton 3 work)
* FLIR Lepton As Camera: Pass (flir lepton 1 and flir lepton 3 work)

OpenMV Cam H7 Plus:
* LCD Module: Pass (triple buffering and non-buffered work)
* TV Module: Pass (triple buffering and non-buffered work)
* WiFi Module: Pass (able to stream UDP and TCP RTSP video at VGA)
* FIR Module: Pass (flir lepton 1 and flir lepton 3 work)
* FLIR Lepton As Camera: Pass (flir lepton 1 and flir lepton 3 work)

OpenMV Cam RT1062:
* LCD Module: Pass (triple buffering and non-buffered work)
* TV Module: Pass (triple buffering and non-buffered work)
* FLIR Lepton As Camera: Pass (flir lepton 1 and flir lepton 3 work)

src/omv/ports/mimxrt/omv_spi.c
```diff
@@ -78,43 +89,81 @@ static void spi_master_callback(LPSPI_Type *base, void *handle, status_t status,
...
+        buf = spi->xfer_descr.txData;
+    }
+
+    // The IMXRT doesn't support half complete transfer interrupts (in the lpspi driver) like the STM32
+    // does. So, mimick support for them by toggling the half flag.
+    uint32_t flags = (OMV_SPI_XFER_DMA | OMV_SPI_XFER_COMPLETE);
+    if ((spi->dma_flags & OMV_SPI_DMA_DOUBLE) && ((spi->xfer_flags & flags) == flags)) {
+        int32_t offset = (spi->xfer_flags & OMV_SPI_XFER_HALF) ? -spi->xfer_descr.dataSize : spi->xfer_descr.dataSize;
+        if (spi->xfer_descr.txData) {
+            spi->xfer_descr.txData += offset;
```
iabdalkader, line 107:
```suggestion
```
Since this is no longer a valid mode, I don't think we need this.

iabdalkader (changes requested):

iabdalkader:
Everything else looks good, thanks!

src/omv/boards/ARDUINO_PORTENTA_H7/omv_boardconfig.h
```diff
@@ -351,31 +363,22 @@
...
+#define OMV_SPI_LCD_RS_PIN                    (&omv_pin_C6_GPIO)
+#define OMV_SPI_LCD_RST_PIN                   (&omv_pin_C7_GPIO)
 
-#define OMV_SPI_LCD_BL_DAC                    (DAC1)
-#define OMV_SPI_LCD_BL_DAC_CHANNEL            (DAC_CHANNEL_1)
-#define OMV_SPI_LCD_BL_DAC_CLK_ENABLE()       __HAL_RCC_DAC12_CLK_ENABLE()
-#define OMV_SPI_LCD_BL_DAC_CLK_DISABLE()      __HAL_RCC_DAC12_CLK_DISABLE()
-#define OMV_SPI_LCD_BL_DAC_FORCE_RESET()      __HAL_RCC_DAC12_FORCE_RESET()
-#define OMV_SPI_LCD_BL_DAC_RELEASE_RESET()    __HAL_RCC_DAC12_RELEASE_RESET()
+#define OMV_SPI_LCD_DEF_TRIPLE_BUF            (true)
```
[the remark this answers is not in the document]
kwagyeman: Hi, I just changed the true to 1. I left the name the same to not conflict with our changes to py_lcd.c

Note, this isn't an enable option. It's just changing the default argument to be True for triple buffering versus false on boards with SDRAM.

src/omv/ports/mimxrt/mimxrt_hal.c
```diff
@@ -20,6 +20,7 @@
 #include "fsl_usb_phy.h"
 #include "fsl_device_registers.h"
 #include CLOCK_CONFIG_H
+#include "dma_manager.h"
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/common/vospi.c
```diff
@@ -19,118 +19,151 @@
...
+
+#define VOSPI_SPECIAL_PACKET        (20)
+#define VOSPI_DONT_CARE_PACKET      (0x0F00)
+#define VOSPI_HEADER_DONT_CARE(x)   (((x) & VOSPI_DONT_CARE_PACKET) == VOSPI_DONT_CARE_PACKET)
+#define VOSPI_HEADER_PID(id)        ((id) & 0x0FFF)
+#define VOSPI_HEADER_SID(id)        (((id) >> 12) & 0x7)
 
 typedef enum {
-    VOSPI_FLAGS_RESET  = (1 << 0),
+    VOSPI_FLAGS_FRAME  = (1 << 0),
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/ports/mimxrt/omv_spi.c
kwagyeman: Done

kwagyeman:
PR Updated.

iabdalkader (approved):

PR #1919, opened 2023-09-08
sensors/mt9v0xx: Fix exposure calculation.

kwagyeman:
Reported to be incorrect per customer usage.

Please wait to merge this. Sending files to customer to reverify this commit is right. Will comment when they confirm.


kwagyeman:
Customer approves changes and they fix the timing.

PR #1934, opened 2023-09-23
drivers: Refactor all display modules.

iabdalkader:


src/omv/ports/stm32/modules/py_display.c
```diff
@@ -0,0 +1,649 @@
...
+    uint32_t hactive;
+    uint32_t vactive;
+    uint32_t pixel_clock;
+    uint32_t hsync_len;
+    uint32_t hback_porch;
+    uint32_t hfront_porch;
+    uint32_t vsync_len;
+    uint32_t vback_porch;
+    uint32_t vfront_porch;
+    uint32_t hpol : 1;
```
kwagyeman, line 48: Did you pass the polarities along? They don't seem to be in the array below.

kwagyeman:
Can confirm the LCD and HDMI work on the Pure Thermal.

src/omv/ports/stm32/modules/py_display.c
iabdalkader: They're all low, so I may just remove those fields.

Issue #887, opened 2020-08-24
Method for Full Manual Mode for Sensor OV7725


ajacobs-github:
For full manual control of the OV7725 sensor, you need to disable auto gain, auto white balance, auto exposure, night mode (auto frame rate) and black level calibration (BLC). Only AFTER all of these auto modes are disabled, restore a bunch of offset registers back to their default values, because these get changed based on the initial images by the various auto modes before they are disabled.

Also manually set white balance.
```Python
# Setup Image Sensor format
sensor.set_framesize(sensor.QVGA)
sensor.set_pixformat(sensor.GRAYSCALE)

# Setup for manual gain, exposure, and white balance
sensor.set_auto_gain(False, gain_db=3) # gain in db
sensor.set_auto_whitebal(False)
sensor.set_auto_exposure(False, exposure_us=10000) # exposure in us

# Setup contrast, brightness and saturation
sensor.set_contrast(3) # range -3 to +3
sensor.set_brightness(0) # range -3 to +3
sensor.set_saturation(0) # range -3 to +3

# Disable night mode (auto frame rate) and black level calibration (BLC)
sensor.__write_reg(0x0E, 0b00000000) # Disable night mode
sensor.__write_reg(0x3E, 0b00000000) # Disable BLC

# Set White Balance values manually
sensor.__write_reg(0x01, 0b01111100) # Blue Gain for White Balance
sensor.__write_reg(0x02, 0b01000000) # Red Gain for White Balance
sensor.__write_reg(0x03, 0b01111100) # Green Gain for White Balance

# Restore to default values (these are changed before manual mode takes effect)
sensor.__write_reg(0x2D, 0b00000000) # LSB Insert Dummy Rows (Set to Default 0x00)
sensor.__write_reg(0x2E, 0b00000000) # MSB Insert Dummy Rows (Set to Default 0x00)

sensor.__write_reg(0x35, 0b10000000) # AD Offset B Chan (Set to Default 0x80)
sensor.__write_reg(0x36, 0b10000000) # AD Offset R Chan (Set to Default 0x80)
sensor.__write_reg(0x37, 0b10000000) # AD Offset Gb Chan (Set to Default 0x80)
sensor.__write_reg(0x38, 0b10000000) # AD Offset Gr Chan (Set to Default 0x80)

sensor.__write_reg(0x39, 0b10000000) # B channel offset (Set to Default 0x80)
[...]

PR #1938, opened 2023-09-27
sensors: Add full control.

kwagyeman:
This PR is paid for by Field Data Tech LLC. Thank you for supporting OpenMV.

Working on the information provided by https://github.com/openmv/openmv/issues/887 I was able to add the appropriate functions to the OV7725 sensor to gain full manual control of the sensor to save the current register settings and then wake back up again and re-apply them. Using the demo in this PR you can go from complete sensor power down to running again and getting a frame in 93 ms.

I then extended the same control to the OV5640 which the client is paying for to have fast save and restore. With this PR the OV5640 can be ready to take a picture in about 215 ms from power down. This meets the clients expectations.

...

It was observed that the math in all the gain calculations was inaccurate. So, I had to switch out to using the real expf/logf functions so that you'd get consistent results setting the gain and then reading it back.


src/omv/modules/py_sensor.c
```diff
@@ -656,6 +692,24 @@ static mp_obj_t py_sensor_ioctl(uint n_args, const mp_obj_t *args) {
         }
         #endif
 
+        #if (OMV_ENABLE_OV5640 == 1 || OMV_ENABLE_OV7725 == 1)
```
iabdalkader, line 695:
```suggestion
```
Those should not be gated by sensor type.

src/omv/modules/py_sensor.c
```diff
@@ -1028,6 +1085,10 @@ STATIC const mp_map_elem_t globals_dict_table[] = {
     { MP_OBJ_NEW_QSTR(MP_QSTR_IOCTL_RESET_AUTO_FOCUS),              MP_OBJ_NEW_SMALL_INT(IOCTL_RESET_AUTO_FOCUS)},
     { MP_OBJ_NEW_QSTR(MP_QSTR_IOCTL_WAIT_ON_AUTO_FOCUS),            MP_OBJ_NEW_SMALL_INT(IOCTL_WAIT_ON_AUTO_FOCUS)},
     #endif
+    #if (OMV_ENABLE_OV5640 == 1 || OMV_ENABLE_OV7725 == 1)
```
iabdalkader, line 1088:
```suggestion
```

src/omv/modules/py_sensor.c
```diff
@@ -1121,4 +1185,5 @@ const mp_obj_module_t sensor_module = {
 };
 
 MP_REGISTER_MODULE(MP_QSTR_sensor, sensor_module);
+
```
iabdalkader, line 1188:
```suggestion
```

src/omv/modules/py_sensor.c
```diff
@@ -451,6 +451,33 @@ static mp_obj_t py_sensor_get_rgb_gain_db() {
     });
 }
 
+static mp_obj_t py_sensor_set_auto_blc(uint n_args, const mp_obj_t *args, mp_map_t *kw_args) {
+    int enable = mp_obj_get_int(args[0]);
+    int *regs = NULL;
+    py_helper_keyword_int_array(n_args, args, 1, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_regs),
```
iabdalkader, line 457: New code should use `mp_arg_parse`.

iabdalkader (changes requested):

src/omv/modules/py_sensor.c
kwagyeman: Fixed. Much nicer.

src/omv/modules/py_sensor.c
kwagyeman: Done

src/omv/modules/py_sensor.c
kwagyeman: Done

src/omv/modules/py_sensor.c
kwagyeman: Done

kwagyeman:
I discussed the context idea with the client before work was started. However, they don't actually want the controls to be in a black box.

They will be dynamically generating the exposure/gain values based on the external environment. So, they want to control the exposure and gain precisely to immediately set it to the right value on power on.

src/omv/modules/py_sensor.c
```diff
@@ -451,6 +451,46 @@ static mp_obj_t py_sensor_get_rgb_gain_db() {
     });
 }
 
+static mp_obj_t py_sensor_set_auto_blc(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    enum { ARG_enable, ARG_regs };
+    static const mp_arg_t allowed_args[] = {
+        { MP_QSTR_enable, MP_ARG_REQUIRED | MP_ARG_INT },
+        { MP_QSTR_regs,   MP_ARG_OBJ },
```
iabdalkader, line 458: Is `regs` required ? Seems so, it should be:
```
{ MP_QSTR_regs, MP_ARG_REQUIRED | MP_ARG_OBJ },
```
Or if it's not, then
```
{ MP_QSTR_regs, MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
```

src/omv/modules/py_sensor.c
```diff
@@ -705,6 +705,24 @@ static mp_obj_t py_sensor_ioctl(uint n_args, const mp_obj_t *args) {
         }
         #endif
 
+        #if (OMV_ENABLE_OV7725 == 1)
```
iabdalkader: There's no need to gate this by sensor, if an ioctl is not supported it will raise an exception, when it's implemented for another sensor we don't need to remember removing this line (same for the QSTR below).

iabdalkader: I see you removed it in the following commit.

src/omv/modules/py_sensor.c
kwagyeman: Fixed. Thanks for the catch.

PR #1950, opened 2023-10-04
ports/stm32: Remove hard-coded MDMA channels.

iabdalkader:


src/omv/ports/stm32/sensor.c
```diff
@@ -57,6 +57,17 @@ void ISC_SPI_IRQHandler(void) {
 }
 #endif // ISC_SPI
 
+#if defined(OMV_MDMA_CHANNEL_DCMI_0)
```
kwagyeman, line 60: The ISRs for this should never fire. So, this isn't needed. However, it should also be harmless to have.

src/omv/imlib/jpeg.c
```diff
@@ -1012,8 +1013,12 @@ void JPEG_IRQHandler() {
 }
 
 void jpeg_mdma_irq_handler() {
-    HAL_MDMA_IRQHandler(&JPEG_MDMA_Handle_In);
-    HAL_MDMA_IRQHandler(&JPEG_MDMA_Handle_Out);
+    if (MDMA->GISR0 & (1 << OMV_MDMA_CHANNEL_JPEG_IN)) {
```
kwagyeman, line 1016: Doesn't the IRQ handler it's self check the stream ISR? I guess this saves the jump.

kwagyeman (commented):
Don't see any issues.

src/omv/imlib/jpeg.c
iabdalkader: > I guess this saves the jump.

Exactly.

PR #1956, opened 2023-10-06
imlib: Fix DMA2D cache line access.

kwagyeman:
Previous code left open the door for corruption to happen in rare cases.


iabdalkader:
Thank you for fixing the indents, will pull this and test it, but it looks good.

PR #1962, opened 2023-10-09
imlib: Enable loading images from paths in all situations.

kwagyeman:
This PR globally unlocks loading images from paths in all functions.

E.g.

```
# Hello World Example
#
# Welcome to the OpenMV IDE! Click on the green run arrow button below to run the script!

import sensor
import time

sensor.reset()  # Reset and initialize the sensor.
sensor.set_pixformat(sensor.RGB565)  # Set pixel format to RGB565 (or GRAYSCALE)
sensor.set_framesize(sensor.VGA)  # Set frame size to QVGA (320x240)
#sensor.skip_frames(time=2000)  # Wait for settings take effect.

while True:
    clock.tick()  # Update the FPS clock.
    img = sensor.snapshot()  # Take a picture and return the image.
    img.draw_image("arduino-mini-logo.jpg", 0, 0)
    print(clock.fps())  # Note: OpenMV Cam runs about half as fast when connected
```

arduino-mini-logo

Just works. Pretty much anywhere you can pass an image object you can now pass a path.

### DIFF
--- a/src/omv/modules/py_display.c
+++ b/src/omv/modules/py_display.c
@@ -145,6 +145,7 @@ STATIC mp_obj_t py_display_write(uint n_args, const mp_obj_t *pos_args, mp_map_t
     mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
     mp_arg_parse_all(n_args - 1, pos_args + 1, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
 
+    fb_alloc_mark();
     image_t *image = py_helper_arg_to_image(args[ARG_image].u_obj, 0);
     rectangle_t roi = py_helper_arg_to_roi(args[ARG_roi].u_obj, image);
 
@@ -167,7 +168,6 @@ STATIC mp_obj_t py_display_write(uint n_args, const mp_obj_t *pos_args, mp_map_t
     const uint16_t *color_palette = py_helper_arg_to_palette(args[ARG_color_palette].u_obj, PIXFORMAT_RGB565);
     const uint8_t *alpha_palette = py_helper_arg_to_palette(args[ARG_alpha_palette].u_obj, PIXFORMAT_GRAYSCALE);
 
-    fb_alloc_mark();
     py_display_p_t *display_p = (py_display_p_t *) MP_OBJ_TYPE_GET_SLOT(self->base.type, protocol);
     display_p->write(self, image, args[ARG_x].u_int, args[ARG_y].u_int, x_scale, y_scale, &roi,
                      args[ARG_channel].u_int, args[ARG_alpha].u_int, color_palette, alpha_palette, args[ARG_hint].u_int);
--- a/src/omv/modules/py_helper.h
+++ b/src/omv/modules/py_helper.h
@@ -25,14 +25,8 @@ const void *py_helper_arg_to_palette(const mp_obj_t arg, uint32_t pixfmt);
 rectangle_t py_helper_arg_to_roi(const mp_obj_t arg, const image_t *img);
 void py_helper_arg_to_scale(const mp_obj_t arg_x_scale, const mp_obj_t arg_y_scale,
                             float *x_scale, float *y_scale, rectangle_t *roi);
-image_t *py_helper_keyword_to_image_mutable(uint n_args, const mp_obj_t *args, uint arg_index,
-                                            mp_map_t *kw_args, mp_obj_t kw, image_t *default_val);
-image_t *py_helper_keyword_to_image_mutable_mask(uint n_args, const mp_obj_t *args, uint arg_index,
-                                                 mp_map_t *kw_args);
-image_t *py_helper_keyword_to_image_mutable_color_palette(uint n_args, const mp_obj_t *args, uint arg_index,
-                                                          mp_map_t *kw_args);
-image_t *py_helper_keyword_to_image_mutable_alpha_palette(uint n_args, const mp_obj_t *args, uint arg_index,
-                                                          mp_map_t *kw_args);
+image_t *py_helper_keyword_to_image(uint n_args, const mp_obj_t *args, uint arg_index,
+                                    mp_map_t *kw_args, mp_obj_t kw, image_t *default_val);
 void py_helper_keyword_rectangle(image_t *img, uint n_args, const mp_obj_t *args, uint arg_index,
                                  mp_map_t *kw_args, mp_obj_t kw, rectangle_t *r);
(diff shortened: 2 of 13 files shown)


iabdalkader:

https://forums.openmv.io/t/load-jpeg-image-from-buffer/8831/2

src/omv/modules/py_helper.h
```diff
@@ -15,14 +15,14 @@ extern const mp_obj_fun_builtin_var_t py_func_unavailable_obj;
...
-image_t *py_helper_keyword_to_image_mutable_alpha_palette(uint n_args, const mp_obj_t *args, uint arg_index,
-                                                          mp_map_t *kw_args);
+image_t *py_helper_keyword_to_image(uint n_args, const mp_obj_t *args, uint arg_index,
+                                    mp_map_t *kw_args, mp_obj_t kw, image_t *default_val);
+image_t *py_helper_keyword_to_image_mask(uint n_args, const mp_obj_t *args, uint arg_index,
+                                         mp_map_t *kw_args);
+image_t *py_helper_keyword_to_image_color_palette(uint n_args, const mp_obj_t *args, uint arg_index,
+                                                  mp_map_t *kw_args);
+image_t *py_helper_keyword_to_image_alpha_palette(uint n_args, const mp_obj_t *args, uint arg_index,
+                                                  mp_map_t *kw_args);
```
iabdalkader, line 25:
```suggestion
```
Can you please remove those functions and just call `py_helper_keyword_to_image` directly ? They just seem useless, and when it's time to switch to mp_parse_args, it will be easier to just fix one function (which will accept a single object).

src/omv/modules/py_display.c
```diff
@@ -105,6 +105,7 @@ STATIC mp_obj_t py_display_backlight(uint n_args, const mp_obj_t *args) {
 STATIC MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(py_display_backlight_obj, 1, 2, py_display_backlight);
 
 STATIC mp_obj_t py_display_write(uint n_args, const mp_obj_t *args, mp_map_t *kw_args) {
+    fb_alloc_mark();
```
iabdalkader, line 108: Wouldn't it just be easier to `fb_alloc_mark` in `py_image_cobj` ? It seems one mark precedes every call to `py_image_cobj`.

iabdalkader (changes requested):

src/omv/modules/py_display.c
kwagyeman: That can't be done because py_image_cobj is called multiple times inside a function. The right solution would be to just add fb_alloc_mark() and fb_alloc_free_till_mark() to whatever calls functions in micropython. This will globally remove the need for these anywhere else and would clean all this up.

Would you like to do that?

src/omv/modules/py_helper.h
kwagyeman: Done

src/omv/modules/py_image.c
```diff
@@ -288,8 +288,25 @@ typedef struct _mp_obj_py_image_it_t {
...
-    PY_ASSERT_TYPE(img_obj, &py_image_type);
-    return &((py_image_obj_t *) img_obj)->_cobj;
+    if (MP_OBJ_IS_STR(img_obj)) {
+        #if defined(IMLIB_ENABLE_IMAGE_FILE_IO)
+        const char *path = mp_obj_str_get_str(img_obj);
+        FIL fp;
+        image_t image;
+        img_read_settings_t rs;
+        imlib_read_geometry(&fp, &image, path, &rs);
+        file_buffer_off(&fp);
```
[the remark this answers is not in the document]
iabdalkader: You also need to resolve the conflicts with the other PR that was merged.

src/omv/modules/py_display.c
iabdalkader: I think everything eventually calls `mp_call_function_n_kw` to do the actual call, hooks can be added there, but this is also used to call everything, including callbacks from Python/C.

src/omv/modules/py_image.c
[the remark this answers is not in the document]
kwagyeman: Done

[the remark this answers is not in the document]
kwagyeman: Mmm, code didn't disappear.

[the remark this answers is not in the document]
kwagyeman: Fixed now.

src/omv/modules/py_display.c
kwagyeman: https://github.com/openmv/micropython/pull/115

However, on thinking more about this we should probably just keep manually wrapping things.

I tested the PR out and the system still works. But, small function calls in a loop will see a pretty steep slow down. Like toggling a GPIO pin in a loop.

src/omv/modules/py_image.c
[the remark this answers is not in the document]
iabdalkader: You're right, will see if I can merge this first, then rebase the file refactoring.

src/omv/modules/py_display.c
iabdalkader: > That can't be done because py_image_cobj is called multiple times inside a function

The only 2 functions that call it multiple times are `int_py_tf_segment` and `py_image_to`. For `int_py_tf_segment` it calls it on a list of image objects, so should be safe. For `py_image_to`, and all of the conversion functions that call it `to_xxx/compressed`, it doesn't make any sense to pass a string, why would you:

```Python
img.to_bitmap(copy="path/to_imag") 
```

The `copy=img` feature and arg_other stuff can be dropped, but if you want to keep it, we should assert that `copy` is Not a string.

kwagyeman: Yeah, I agree with dropping all the copy=img stuff. Not needed anymore.

As for it being called multiple times. I mean like, any function that supports a mask argument technically accepts a second image. And then like draw_image(). There are a ton of methods that support mask images. So, I moved the code around whenever there's more than 1 image object in use by the method.

iabdalkader: > I mean like, any function that supports a mask argument technically accepts a second image. And then like draw_image().

A second image does Not cause `fb_alloc_mark` to be called, only a path. Can you point out a specific function where `py_image_cobj` is called twice in a row such that `fb_alloc_mark` is called twice ?

kwagyeman: Difference for example. You could pass an image which could be a path and then a mask: https://github.com/openmv/openmv/blob/master/src/omv/modules/py_image.c#L2383

kwagyeman:
PR is ready to go.

PR #1974, opened 2023-10-13
ports/mimxrt: Add support for fsync pin to snapshot.

kwagyeman:
Enables FYSNC for the MT9V034. Tested and works. Also, noticed the MISO/MOSI pins were swapped. Code previously was working since in ALT function mode it doesn't matter. But, thought to fix it.


iabdalkader:
Those commits need to be edited.

PR #1975, opened 2023-10-13
imlib: Support creating images from bytearrays.

kwagyeman:
- Updates load image to use parse all.
- Adds new feature where you can now pass a bytearray in when making an image object to make a deep copy image from that byte array.

### DIFF
--- a/src/omv/modules/py_image.c
+++ b/src/omv/modules/py_image.c
@@ -6870,95 +6870,77 @@ mp_obj_t py_image_from_struct(image_t *img) {
     return o;
 }
 
-mp_obj_t py_image_load_image(uint n_args, const mp_obj_t *args, mp_map_t *kw_args) {
-    // mode == false -> load behavior
-    // mode == true -> make behavior
-    bool mode = mp_obj_is_integer(args[0]);
-    const char *path = mode ? NULL : mp_obj_str_get_str(args[0]);
+mp_obj_t py_image_load_image(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    enum { ARG_height, ARG_pixformat, ARG_buffer, ARG_copy_to_fb };
+    static const mp_arg_t allowed_args[] = {
+        { MP_QSTR_height,       MP_ARG_INT, {.u_int = -1} },
+        { MP_QSTR_pixformat,    MP_ARG_INT, {.u_int = -1} },
+        { MP_QSTR_buffer,       MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_copy_to_fb,   MP_ARG_BOOL | MP_ARG_KW_ONLY, {.u_bool = false} },
+    };
 
-    mp_obj_t copy_to_fb_obj = py_helper_keyword_object(n_args, args,
-                                                       mode ? 3 : 1, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_copy_to_fb), NULL);
-    bool copy_to_fb = false;
-    image_t *arg_other = NULL;
-
-    if (copy_to_fb_obj) {
-        if (mp_obj_is_integer(copy_to_fb_obj)) {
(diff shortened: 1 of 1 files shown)


iabdalkader:

https://forums.openmv.io/t/load-jpeg-image-from-buffer/8831/6

iabdalkader:
Thanks for starting to use `mp_arg_parse`, but the args are very hacky, I'd like to drop support for `Image(copy_2_fb=other_image)` as it's not used anywhere in the scripts, I searched, but it can still be passed as keyword arg if it must be supported.

```C
    enum { ARG_path, ARG_width, ARG_height, ARG_pixformat, ARG_buffer, ARG_image, ARG_copy_to_fb };
        { MP_QSTR_width,      MP_ARG_INT, {.u_int = 0} },
        { MP_QSTR_height,     MP_ARG_INT, {.u_int = 0} },
        { MP_QSTR_pixformat,  MP_ARG_INT, {.u_int = 0} },
        { MP_QSTR_buffer,     MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
        { MP_QSTR_image,      MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
        { MP_QSTR_copy_2_fb,  MP_ARG_BOOL | MP_ARG_KW_ONLY, {.u_bool = true} },
    };

    if (n_args <= 2) {
        // load from path + copy_2_fb
        const char *path = mp_obj_str_get_str(args[0]);
        }
    }
STATIC MP_DEFINE_CONST_FUN_OBJ_KW(py_image_load_image_obj, 1, py_image_load_image);
```

kwagyeman:
Dropped support for copy_to_fb into other images. This makes the code a lot more readable and reduces complexity. Good call.

I split the function into two code paths. One for file loading and the other for in memory buffers. I did this so that parse_all would give you the right error messages in either situation. Additionally, you'd have to define `width` as an u_obj because it will try to cast it to an int otherwise and error out on you if you passed a path.

As for `copy_2_fb`. I'd like to leave that as `copy_to_fb` to not break user code and left it as default off as this was the old API.

Here's the new behavior:

`Image("path")` - Creates an image object in the micropython heap.
`Image("path", copy_to_fb=True) - Replaces the fb with the image object loaded from disk.

`Image(w, h, pixformat)` - Creates an image object in the micropython heap. Image defaults to 0.
`Image(w, h, pixformat, copy_to_fb=True) - Replaces the fb with the image object. Image defaults to 0.
`Image(w, h, pixformat, buffer) - Creates an image object which uses buffer as the data array. This is a shallow copy.
`Image(w, h, pixformat, buffer, copy_to_fb=True) - Replaces the fb with the image object land data from buffer. This is a deep copy.

Unless the user was using the special copy_to_fb=image usage the new method should work in place in all code paths.

iabdalkader:
This is not working I tested it, creating/loading an image to FB doesn't show up. I think you're reintroducing the same bug that I fixed before, it has to do with when you call `update_jpeg_buffer` it needs to be called after at the end of the function for both load/create.

```
img = image.Image(4, 4, image.RGB565, copy_to_fb=True)
img = image.Image("/test.bmp", copy_to_fb=True)
```

Both don't update FB, this fixes the loading:

```c
        framebuffer_update_jpeg_buffer();
    }
    return py_image_from_struct(&image);
```

Note in the future please test these PRs, I assume you do otherwise I will have to test every PR.

Also I don't see any good reason to have 2 sets of args and parse, it doesn't raise any meaningful exceptions, for example:
```C
image.Image(4, image.RGB565, copy_to_fb=True)
```
Raise this exception: `TypeError: 'pixformat' argument required` because they're all ints (so it thinks you passed width, height) is it more helpful ? We can maybe use more asserts, can check for width min/max range, that's the best the can be done.

I also have comments on the code, you can load the buffer once no need to load it 3 times, and about the shallow copy I'm honestly not 100%  sure it's not going to be collected, the pointer to it is in the C image_t which can be  the framebuffer, so not 100% sure if this is safe. I will need to do some tests and I don't like this function as it is so I will re-write it.

kwagyeman:
Updated with your new function. Works fine.

PR #1976, opened 2023-10-15
misc: Refactor file management code.

iabdalkader:
- I split functions into two groups:
  1-  "low-level" functions that don't raise exceptions and don't do file buffering (file_ll_x) those are 1-1 with FatFS
  2-  Regular functions that are aware of file buffering (file_x)  those have extra args sometimes.
- I replaced all direct calls to FatFs, which might be useful in the future if we want to port the code.
- Very little remaining uses the "low-level" functions, I'm still thinking about porting those to the high level functions.
- `file_buffer_off` is done implicitly now on `file_close()`, but it can still be managed manually, some code needs that.
- removed all of the read_byte, read_word, read_long, those were useless, and kept the write counterparts.
- To read and skip bytes you just pass NULL to `file_read` and a count.


kwagyeman:
Looks good. Was able to review everything.

Issue #1933, opened 2023-09-18
Add callback self pointer to draw_image


kwagyeman:

iabdalkader:
Also, just pass an array or `rectangle_t` ?
```
bool imlib_draw_image_rectangle(...., int *x0, int *x1, int *y0, int *y1);
bool imlib_draw_image_rectangle(...., int *points (or rectangle_t *r));
....

points[P_X0] =..;
```

iabdalkader:
When you do, please remove this:
https://github.com/openmv/openmv/blob/master/src/omv/modules/py_spi_display.c#L161

kwagyeman:
Sure, will do this now.

PR #1992, opened 2023-10-26
imlib: Move draw_image arguments to struct.

kwagyeman:
Resolves https://github.com/openmv/openmv/issues/1933.

### DIFF
--- a/src/omv/imlib/apriltag.c
+++ b/src/omv/imlib/apriltag.c
@@ -11987,7 +11987,7 @@ void imlib_find_apriltags(list_t *out, image_t *ptr, rectangle_t *roi, apriltag_
     img.h = roi->h;
     img.pixfmt = PIXFORMAT_GRAYSCALE;
     img.data = fb_alloc(image_size(&img), FB_ALLOC_NO_HINT);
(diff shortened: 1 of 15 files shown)


iabdalkader:
This is good, almost mergable right away, however I have 2 comments:

```suggestion
```

1- I don't think there's a need for a new type `point_rectangle_t`, please reuse the existing ones, you can just pass two `point_t`, `p0`, `p1` to `imlib_draw_image_rectangle`.

2- Speaking of `imlib_draw_image_rectangle`, I think it should be called `imlib_draw_image_get_bounds` or something like that, that function has always confused me, sounds like it's drawing a rectangle but it's not. Also, may want to return the fact that there's nothing to draw `black` using one of the args, whatever works, like setting `x0` to -1 or something.

kwagyeman:
Done!

kwagyeman:
Git rebase interactive appears to have merged two of my commits...

kwagyeman:
Recent changes have not been tested. Will retest if idea is approved.

kwagyeman:
@iabdalkader - This PR is updated as you wanted it.

iabdalkader (approved):

PR #2005, opened 2023-10-30
scripts: Fix light shield and low power ordering.

kwagyeman:
Was moving too fast. This fixes these scripts order to match what it should be.

New IDE is building with the correct order.


iabdalkader:
The current order is fine I think:
```
50-Board-Control
50-Light-Shield
51-Low-Power
52-LCD-Shield
53-Servo-Shield
54-Thermopile-Shield
55-BLE-Shield
56-Motor-Shield
57-IMU-Shield
58-Distance-Shield
59-TV-Shield
99-Tests
``

kwagyeman:
Light shield was 10 and low power was 11. So, they should be 60 and 61.

PR #2008, opened 2023-10-31
modules: Allow arg_to_image to load images from paths.

kwagyeman:
Targeted now to just two methods.

### DIFF
--- a/src/omv/modules/py_helper.c
+++ b/src/omv/modules/py_helper.c
@@ -24,17 +24,33 @@ mp_obj_t py_func_unavailable(uint n_args, const mp_obj_t *args, mp_map_t *kw_arg
 MP_DEFINE_CONST_FUN_OBJ_KW(py_func_unavailable_obj, 0, py_func_unavailable);
 
 image_t *py_helper_arg_to_image(const mp_obj_t arg, uint32_t flags) {
-    image_t *arg_img = py_image_cobj(arg);
+    image_t *image = NULL;
+    if ((flags & ARG_IMAGE_ALLOC) && MP_OBJ_IS_STR(arg)) {
+        #if defined(IMLIB_ENABLE_IMAGE_FILE_IO)
+        const char *path = mp_obj_str_get_str(arg);
+        FIL fp;
+        image = xalloc(sizeof(image_t));
+        img_read_settings_t rs;
+        imlib_read_geometry(&fp, image, path, &rs);
+        file_close(&fp);
+        image->data = fb_alloc(image_size(image), FB_ALLOC_CACHE_ALIGN);
+        imlib_load_image(image, path);
+        #else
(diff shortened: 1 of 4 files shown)


src/omv/modules/py_helper.c
```diff
@@ -24,6 +24,21 @@ mp_obj_t py_func_unavailable(uint n_args, const mp_obj_t *args, mp_map_t *kw_arg
...
+        imlib_read_geometry(&fp, image, path, &rs);
+        file_close(&fp);
+        image->data = fb_alloc(image_size(image), FB_ALLOC_CACHE_ALIGN);
+        imlib_load_image(image, path);
+        return image;
+        #else
+        mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Image I/O is not supported"));
+        #endif // IMLIB_ENABLE_IMAGE_FILE_IO
+    }
     image_t *arg_img = py_image_cobj(arg);
```
iabdalkader, line 42:
```suggestion
image_t *py_helper_arg_to_image(const mp_obj_t arg, uint32_t flags) {
    image_t *image = NULL;
    if (MP_OBJ_IS_STR(arg)) {
        #if defined(IMLIB_ENABLE_IMAGE_FILE_IO)
        const char *path = mp_obj_str_get_str(arg);
        FIL fp;
        image = xalloc(sizeof(image_t));
        img_read_settings_t rs;
        imlib_read_geometry(&fp, image, path, &rs);
        file_close(&fp);
        image->data = fb_alloc(image_size(image), FB_ALLOC_CACHE_ALIGN);
        imlib_load_image(image, path);
        #else
        mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Image I/O is not supported"));
        #endif // IMLIB_ENABLE_IMAGE_FILE_IO
    } else {
       image = py_image_cobj(arg);
    }
```
This PR is okay, but why not let it fall through the flags check ? This way it can load specific formats, if it's ever needed in the future.

kwagyeman: Yes, this is better. But, now you can cause an alloc that's not protected since you can pass a path at any time :) so we'd need to fix alloc/free all over the place to make sure there were no bugs.

iabdalkader: Right, I see all calls use one flag or the other..In that case would it be better to explicitly pass a new flag like `ARG_IMAGE_ANY | ARG_IMAGE_ALLOC` ?

iabdalkader: No functional change, just makes it clear that image can be allocated.

kwagyeman: Make sense. Let me change it to that.

kwagyeman: Done. Also, made the previous changes you wanted since the new flag prevents the alloc issue.

PR #2015, opened 2023-11-13
sensors/gc2145: Add support for a wider FoV.

kwagyeman:
You can test increasing the field of view with the code below. The frame rate will drop from 20 FPS to 12 FPS. But, you will see quite a lot more in the field of view.

```
import sensor
import time

sensor.reset()  # Reset and initialize the sensor.
sensor.ioctl(sensor.IOCTL_SET_FOV_PREFERRED, False)
sensor.set_pixformat(sensor.RGB565)  # Set pixel format to RGB565 (or GRAYSCALE)
sensor.set_framesize(sensor.QVGA)  # Set frame size to QVGA (320x240)
sensor.skip_frames(time=2000)  # Wait for settings take effect.

while True:
    clock.tick()  # Update the FPS clock.
    img = sensor.snapshot()  # Take a picture and return the image.
    print(clock.fps())  # Note: OpenMV Cam runs about half as fast when connected
```


src/omv/common/sensor.h
```diff
@@ -137,6 +137,8 @@ typedef enum {
     IOCTL_GET_READOUT_WINDOW,
     IOCTL_SET_TRIGGERED_MODE,
     IOCTL_GET_TRIGGERED_MODE,
+    IOCTL_SET_FOV_PREFERRED,
+    IOCTL_GET_FOV_PREFERRED,
```
iabdalkader, line 141: Can you rename this to set/get WIDE_FOV please ?

iabdalkader (changes requested):

src/omv/sensors/gc2145.c
```diff
@@ -726,6 +728,8 @@ static int reset(sensor_t *sensor) {
     readout_w = ACTIVE_SENSOR_WIDTH;
     readout_h = ACTIVE_SENSOR_HEIGHT;
 
+    fov_preferred = false;
```
iabdalkader, line 731:
```suggestion
    wide_fov_enable = false;
```

src/omv/common/sensor.h
iabdalkader: Or FOV_WIDE.

kwagyeman: Done

PR #2018, opened 2023-11-18
Enable OV5640 for GIGA R1

Bexin3:


Bexin3:
It appears there is more work to be done to get this to work

Issue #1838, opened 2023-05-09
Add timeout_char to uart_control example.


kwagyeman:
Add timeout_char to uart_control example.

PR #2021, opened 2023-11-22
scripts/examples: Add timeout_char to uart examples.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/1838


iabdalkader:
Not sure if machine uart supports that arg on every port, you need to check or test it on every port. Also 1000 ms is excessive, it's normally just 200ms which is already a lot of time between characters.

kwagyeman:
Done.

This PR is mainly just to stop forum posts by folks who tried to use the UART at low baud rates and then it doesn't work using the PYB module on the STM32 because they didn't set the timeout_char value. We get less people asking about this because it's heavily documented on the forums now. But this PR is just in case.

PR #2025, opened 2023-11-25
modules/py_tf: Revert `load_to_fb` arg name.

iabdalkader:


iabdalkader:
Would it make sense to rename this to `copy_to_fb` to be consistent with everything else ? @kwagyeman

kwagyeman:
It's not copying an image though, that's why I didn't name it that.

iabdalkader:
> It's not copying an image though, that's why I didn't name it that.

https://docs.openmv.io/library/omv.image.html#image.Image

Loading images with `Image()` doesn't copy either, but it's called `copy_to_fb`.

iabdalkader:
Okay reverted to `load_to_fb`

kwagyeman:
Yes, edge impulse generates scripts with the current naming.

PR #2027, opened 2023-11-25
scripts/examples: Update thermal scripts for new refectored API.

kwagyeman:
Having to pass floats for the scale here to avoid issues made me realize we have to drop the size feature on scale being an int. See: https://github.com/openmv/openmv/pull/2028

### DIFF
--- a/scripts/examples/50-Arduino-Boards/Nano-33-BLE-Sense/53-Thermal/thermal_camera.py
+++ b/scripts/examples/50-Arduino-Boards/Nano-33-BLE-Sense/53-Thermal/thermal_camera.py
@@ -11,6 +11,7 @@ import image
 import time
 import fir
 
+IMAGE_SCALE = 5  # Scale image to 5x.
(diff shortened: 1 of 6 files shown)


scripts/examples/50-Arduino-Boards/Nano-33-BLE-Sense/53-Thermal/thermal_camera.py
```diff
@@ -11,25 +11,16 @@
 import time
 import fir
 
+IMAGE_SCALE = 5.0  # Higher scaling uses more memory.
```
iabdalkader, line 14:
```suggestion
IMAGE_SCALE = 5  # Scale image to 5x.
```

Should be the same everywhere, no need for float.

kwagyeman: Done

PR #2028, opened 2023-11-25
modules: Make int/floating image scaling behavior the same.

kwagyeman:
Having scale turn into a size if you pass an int is too dangerous. Just realized this after going through examples. This makes the behavior scaling and nothing else.

### DIFF
--- a/src/omv/modules/py_display.c
+++ b/src/omv/modules/py_display.c
@@ -159,7 +159,7 @@ STATIC mp_obj_t py_display_write(uint n_args, const mp_obj_t *pos_args, mp_map_t
 
     float x_scale = 1.0f;
     float y_scale = 1.0f;
-    py_helper_arg_to_scale(args[ARG_x_scale].u_obj, args[ARG_y_scale].u_obj, &x_scale, &y_scale, &roi);
+    py_helper_arg_to_scale(args[ARG_x_scale].u_obj, args[ARG_y_scale].u_obj, &x_scale, &y_scale);
 
     if (y_scale < 0 && !self->triple_buffer) {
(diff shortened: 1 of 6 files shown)


iabdalkader:
Dangerous how ? Scale should just have one meaning, a single float multiplied by width/height.

kwagyeman:
I pass a value of scale of 32 in some scripts for example in the drawing examples code. This would have to be made into 32.0. I had to make all the scale values into an explicit float for the thermal API.

Yes, I know that using an int versus float was your intention but making a completely different behavior for float versus int is too complex. we will surely get people posting to the forums rather quickly about this behavior change.

Let's just leave it as a scale.

iabdalkader:
> Yes, I know that using an int versus float was your intention but making a completely different behavior for float versus int is too complex.

Okay, but you don't have to read int vs float then, you can just `get_float`, so this should be enough:

```c
void py_helper_arg_to_scale(const mp_obj_t arg_x_scale, const mp_obj_t arg_y_scale, float *x_scale, float *y_scale) {
    if (arg_x_scale != mp_const_none) {
        *x_scale = mp_obj_get_float(arg_x_scale);
    }
    if (arg_y_scale != mp_const_none) {
        *y_scale = mp_obj_get_float(arg_y_scale);
    }

    if (arg_x_scale == mp_const_none && arg_y_scale != mp_const_none) {
        *x_scale = *y_scale;
    } else if (arg_y_scale == mp_const_none && arg_x_scale != mp_const_none) {
        *y_scale = *x_scale;
    }
}
```

And the other PR, the scale doesn't have to be float either (`IMAGE_SCALE=10` will work).

kwagyeman:
Done

PR #2031, opened 2023-11-27
modules/py_tf: Fix non-square image support.

kwagyeman:
Non-square image support has not been working for a long time. This forum post pointed out the issue: https://forums.openmv.io/t/edge-impulse-result-wrong-on-openmv-h7-plus/8968/3.

Code is now updated to use the new draw image scaling and centering logic. This applies for classification and segmentation/detect.

Then segmentation has been updated to locate the final result in the right location given all the scaling and cropping going on when your image aspect ratio does not match the CNN aspect ratio.

### DIFF
--- a/src/omv/modules/py_tf.c
+++ b/src/omv/modules/py_tf.c
@@ -347,11 +347,6 @@ STATIC void py_tf_input_data_callback(void *callback_data,
     int shift = (params->input_datatype == LIBTF_DATATYPE_INT8) ? GRAYSCALE_MID : 0;
     float fscale = 1.0f / GRAYSCALE_RANGE;
 
-    float xscale = params->input_width / ((float) arg->roi->w);
-    float yscale = params->input_height / ((float) arg->roi->h);
-    // MAX == KeepAspectRationByExpanding - MIN == KeepAspectRatio
-    float scale = IM_MAX(xscale, yscale);
-
     image_t dst_img;
     dst_img.w = params->input_width;
     dst_img.h = params->input_height;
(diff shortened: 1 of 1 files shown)


src/omv/modules/py_tf.c
```diff
@@ -704,6 +699,10 @@ STATIC mp_obj_t int_py_tf_segment(bool detecting_mode, uint n_args, const mp_obj
         image_t *img = py_image_cobj(img_list->items[i]);
         float x_scale = roi.w / ((float) img->w);
         float y_scale = roi.h / ((float) img->h);
+        // MAX == KeepAspectRationByExpanding - MIN == KeepAspectRatio
```
iabdalkader, line 702:
```suggestion
        // MAX == KeepAspectRatioByExpanding - MIN == KeepAspectRatio
```

src/omv/modules/py_tf.c
```diff
@@ -704,6 +699,10 @@ STATIC mp_obj_t int_py_tf_segment(bool detecting_mode, uint n_args, const mp_obj
...
+        float scale = IM_MIN(x_scale, y_scale);
+        int x_offset = fast_floorf((roi.w - (img->w * scale)) / 2.f) + roi.x;
+        int y_offset = fast_floorf((roi.h - (img->h * scale)) / 2.f) + roi.y;
```
iabdalkader, line 705:
```suggestion
        int x_offset = fast_floorf((roi.w - (img->w * scale)) / 2.0f) + roi.x;
        int y_offset = fast_floorf((roi.h - (img->h * scale)) / 2.0f) + roi.y;
```

iabdalkader (changes requested):

src/omv/modules/py_tf.c
kwagyeman: Fixed

PR #2034, opened 2023-11-30
imlib: Fix imlib_midpoint_pool problem

lxowalle:
Fixed the problem that imlib_midpoint _pool cannot select the centered area when processing PIXELFORMAT GRAYSCALE format images.


kwagyeman (approved):

lxowalle:
> You need to fix your commit message to match our commit message format (so, add a period to the message).
> 
> Also, the pooling functions will be deprecated. You should use scale(). I'm not sure if I will keep the midpoint pooling functionality. Is it useful for anything?
> 
> E.g. I could add midpoint pooling to draw_image hints. You have area scaling which replaces the need for mean pooling.
> 
> Whatever the case, mean/midpoint pooling will be removed.

Oh no! I don't see a deprecation note in the manual (https://docs.openmv.io/library/omv.image.html#image.Image.mean_pool), is there any place to know that the API will be deprecated?

kwagyeman:
I'm just thinking about deprecating it. I can make a wrapper around scale to make mean_pool to continue to work. However, it's extra work to continue keeping midpoint pooling around.

How are you using midpoint pooling?

lxowalle:
> I'm just thinking about deprecating it. I can make a wrapper around scale to make mean_pool to continue to work. However, it's extra work to continue keeping midpoint pooling around.
> 
> How are you using midpoint pooling?

I don't actually use midpoint_pool, I just found this when trying to look at his implementation

kwagyeman:
Ah, well, thanks for the fix!

PR #2052, opened 2023-12-22
ports/mimxrt: Service TinyUSB while waiting for sensor frames.

kwagyeman:
Not sure why this works when the IDE is attached but crashes the system without the IDE.

I guess the code to update the frame buffer was saving us. But, that doesn't run anymore when the IDE is not attached.

### DIFF
--- a/src/omv/ports/mimxrt/sensor.c
+++ b/src/omv/ports/mimxrt/sensor.c
@@ -335,7 +335,7 @@ int sensor_snapshot(sensor_t *sensor, image_t *image, uint32_t flags) {
     vbuffer_t *buffer = framebuffer_get_head(FB_NO_FLAGS);
     // Wait for the DMA to finish the transfer.
     for (mp_uint_t ticks = mp_hal_ticks_ms(); buffer == NULL;) {
-        buffer = framebuffer_get_head(FB_NO_FLAGS);
+        MICROPY_EVENT_POLL_HOOK
         if ((mp_hal_ticks_ms() - ticks) > 3000) {
             sensor_abort();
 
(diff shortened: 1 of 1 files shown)


iabdalkader:

https://forums.openmv.io/t/rt1062-trouble-with-save-open-script-as-main-py/9032/6

kwagyeman:
@iabdalkader

Adding another:

```
MICROPY_EVENT_POLL_HOOK
vbuffer_t *buffer = framebuffer_get_head(FB_NO_FLAGS);
// Wait for the DMA to finish the transfer.
for (mp_uint_t ticks = mp_hal_ticks_ms(); buffer == NULL;) {
```

Fixes the user problems here:

https://forums.openmv.io/t/rt1062-trouble-with-save-open-script-as-main-py/9032/13

One is not enough... However, adding another event loop hook seems wrong. From the users code it appears that print() doesn't service TinyUSB if you the CDC port is not open. I also confirmed this via the code. So... it would appear then that if you were to do:

```
while(True):
    print("Hello")
```

You'd crash the system since nothing is polling TinyUSB when nothing is connected to the CDC port. This also crashes the system:

```
while(True):
    pass
```

Any thoughts on how to handle this? I think polling needs to be added to the micropython scheduler or something.

iabdalkader:
> Any thoughts on how to handle this?

Yes I have a fix, and it's not this one, but this might be a good idea in general. I started polling events in "low-level" loop like that a while ago, not just for USB but to schedule callbacks etc.. but we need to think about if we want to let the VM handle pending events (service soft-timer, schedule callbacks etc..) in this loop or not. This is not required to fix the USB.

kwagyeman:
The poll only gets called if there's not a frame available on the first test. So, it doesn't slow down your maximum frame rate if the camera is pumping out images at 100+ FPS. So, it should be fine.

Issue #2042, opened 2023-12-04
SPI Deinit Turns off GPIO Pin


kwagyeman:
So, our current SPI deinit code turns off the SPI SSEL gpio pin: https://github.com/openmv/openmv/blob/master/src/omv/ports/stm32/stm32fxxx_hal_msp.c#L523

This is what we want to do with C drivers. However, with python this causes a problem, e.g.

```
p3 = Pin("P3", mode=Pin.OUTPUT)
p3.value(0)
spi = SPI(2, SPI.MASTER)

spi.send(55)

spi.deinit()
# p3 is now deinited so the below call will not doing anything
p3.value(1)
```

So, what should be the fix? Remove the SSEL omv_gpio_deinit? Or, have the user reinit the Pin object again? Seems like removing the SSEL gpio deinit is the best choice. Note, this is also an issue for the IMXRT. @iabdalkader

Issue #1926, opened 2023-09-09
RPC Library SPI bus transfer is not working


kwagyeman:
Schedule to fix this when doing the update to convert the RPC library to the machine module.

iabdalkader:
This issue has been mentioned on **OpenMV Forums**. There might be relevant details there:

https://forums.openmv.io/t/spi-slave-error-no-attribute-put-short-timeout/8966/4

kwagyeman:
This post goes into detail why this is bugged right now. https://forums.openmv.io/t/spi-deinit-changing-pin-function/8972

Basically, the I/O pin gets deinited by the spi bus and needs to be reinited. This should fix the issue. Not sure if I should make the SPI code though not deinit the SS pin. This is the behavior we want in C code... just not in python land.

@iabdalkader Any thoughts on this? I can just fix the issue in the RPC library easily. Or, should we not deinit the SS pin in the HAL deinit call?

PR #2055, opened 2023-12-23
ports: Don't deinit the SSEL pin on stm32 and mimxrt.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2042

And: https://github.com/openmv/openmv/issues/1926


iabdalkader:
For stm32, you should deinit SSEL in `omv_spi_deinit()`. For imxrt this change is not needed because `mimxrt_hal_spi_deinit` is Not called by `spi.deinit()`.

PR #2056, opened 2023-12-23
scripts/libraries: Change the RPC Interface Library UART Master.

kwagyeman:
Updates the interface library UART Master/Slave to work with Machine UART. Also, removed the person detection code from the library. Tested comms between the RT1062 and H7 Plus.

IMG_8653


iabdalkader:

https://forums.openmv.io/t/rpc-library-import-error-when-using-rt1062-cam/9012/10

iabdalkader:
Note you can use machine.I2C for the master case.

Issue #2066, opened 2023-12-31
Add get/set register settings to SPI display driver.


kwagyeman:
Add hooks so that users can set registers and etc so they can use custom displays.

iabdalkader:
This issue has been mentioned on **OpenMV Forums**. There might be relevant details there:

https://forums.openmv.io/t/spidisplay-in-the-display-module-has-no-effect-on-tftlcd-screens-of-other-sizes/9051/3

PR #2068, opened 2023-12-31
modules/py_display: Add user SPI LCD register control.

kwagyeman:
I added a stub to allow you to execute register reads and writes to the SPI display. I then implemented the SPI register write. I don't believe read will work for our driver as we didn't hookup that line.

Anyway, code is redundant from the DSI read/write. Thoughts what the API should be?

This fixes: https://github.com/openmv/openmv/issues/2066

### DIFF
--- a/src/omv/modules/py_spi_display.c
+++ b/src/omv/modules/py_spi_display.c
@@ -73,13 +73,18 @@ static void spi_switch_mode(py_display_obj_t *self, int bits, bool dma) {
     omv_spi_init(&self->spi_bus, &spi_config);
 }
 
-static void spi_display_command(py_display_obj_t *self, uint8_t cmd, uint8_t arg) {
+static int spi_write(py_display_obj_t *self, uint8_t cmd, uint8_t *args, size_t n_args, bool dcs) {
     omv_gpio_write(OMV_SPI_DISPLAY_RS_PIN, 0);
     spi_transmit(self, (uint8_t []) { cmd }, 1);
     omv_gpio_write(OMV_SPI_DISPLAY_RS_PIN, 1);
-    if (arg) {
-        spi_transmit(self, (uint8_t []) { arg }, 1);
+    if (n_args) {
+        spi_transmit(self, args, n_args);
     }
+    return 0;
+}
+
+static void spi_display_command(py_display_obj_t *self, uint8_t cmd, uint8_t arg) {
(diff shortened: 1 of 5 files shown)


iabdalkader:
We can just have generic display.bus_read/write functions that call both. The SPI functions will just have to accept one extra unused arg (dcs) easier this way, or make bus_read/write accept va args. However I'm not sure this actually fixes anything, the SPI driver already writes some init registers, and it's hard-coded to support a specific controller's init/reset sequence. I don't think this is needed, unless it actually fixes that linked forums issue.

kwagyeman:
I see, I was just trying to give the user an option to write their own custom register settings. But, I agree with going all the way and making a custom controller driver. That way then folks can submit their own drivers.

iabdalkader:

https://forums.openmv.io/t/tft-lcd-screen-module/764/14

kwagyeman:
I updated the dsi read/write to be generic bus read/write calls. 

Then, I added support for spi writing to the display regs via this generic interface. I also added controller support to the spi class.

However, I made it such that if you don't pass a controller, you get a default initialization. I did this not to break all code using the SPI LCD again as the previous change from the lcd module to the display module did.

Anyway, this PR should give folks an easy way to change the LCD settings.

iabdalkader:
Don't you want to move the display controller to Python ?

src/omv/modules/py_spi_display.c
```diff
@@ -394,6 +414,7 @@ STATIC const py_display_p_t py_display_p = {
     #ifdef OMV_SPI_DISPLAY_BL_PIN
     .set_backlight = spi_display_set_backlight,
     #endif
+    .bus_write = spi_write,
```
iabdalkader, line 417: No `bus_read` ?

iabdalkader (approved):

src/omv/modules/py_spi_display.c
kwagyeman: We didn't connect it on the LCD shield so I don't have anything to test with at the moment. I can write a new display controller class for like a 320x240 screen and then test and add read support later.

kwagyeman:
@iabdalkader - I'd like to leave the default one for our SPI shields in C as it makes the code work. Also, it's incredibly minimal. You can pass a display controller for new shields as mentioned above. I can produce that later.

PR #2076, opened 2024-01-03
modules/py_image: Refactor isp ops to use mp_arg_parse.

kwagyeman:
Move these to mp_arg_parse.

Also, I split up awb into parts so that it can be used for software white balance in the future for bayer only cameras we might use.

### DIFF
--- a/src/omv/modules/py_image.c
+++ b/src/omv/modules/py_image.c
@@ -1849,14 +1849,27 @@ STATIC MP_DEFINE_CONST_FUN_OBJ_KW(py_image_flood_fill_obj, 2, py_image_flood_fil
 // ISP Methods
 //////////////
 
-STATIC mp_obj_t py_awb(uint n_args, const mp_obj_t *args, mp_map_t *kw_args) {
-    image_t *arg_img =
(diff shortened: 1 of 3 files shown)


src/omv/modules/py_image.c
```diff
@@ -1849,14 +1849,27 @@ STATIC MP_DEFINE_CONST_FUN_OBJ_KW(py_image_flood_fill_obj, 2, py_image_flood_fil
...
-    imlib_awb(arg_img, arg_max);
-    return args[0];
+    // Parse args.
+    image_t *image = py_helper_arg_to_image(pos_args[0], ARG_IMAGE_UNCOMPRESSED);
+    mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
+    mp_arg_parse_all(n_args - 1, pos_args + 1, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
+
+    uint32_t r_out, g_out, b_out;
+
+    if (args[ARG_max].u_int) {
```
iabdalkader, line 1865:
```suggestion
    if (args[ARG_max].u_bool) {
```

iabdalkader (changes requested):

PR #2082, opened 2024-01-05
scripts/examples: Add working Pure Thermal OpenMV example.

kwagyeman:
Example script was super broken.


scripts/examples/00-HelloWorld/pure_thermal_main.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+try:
+    main()
+except OSError:
+
+    # I2C Bus may be stuck
+    p = pyb.Pin('P4', pyb.Pin.OUT_OD)
+    for i in range(20000):
+        p.value(not p.value())
```
iabdalkader, line 112: The FIR module does this automatically, or is this for another I2C device ?

scripts/examples/index.csv
```diff
@@ -1,6 +1,7 @@
 # "Relative Path Regex", "Board Type Regex", "Sensor Type Regex", "Flatten Regex"
 "examples/00-HelloWorld/blinky.py", ".+", ".+", ""
 "examples/00-HelloWorld/helloworld.py", ".+", "^(?!None).*$", ""
+"examples/00-HelloWorld/pure_thermal_main.py", "(OPENMVPT)", "^(?!None).*$", ""
```
iabdalkader, line 4: I'm not so sure about this, helloworld examples are supposed to be really generic, basic stuff. Maybe you can add a new examples directory for PT in `50-OpenMV-Boards/PureThermal` ?

scripts/examples/00-HelloWorld/pure_thermal_main.py
```diff
@@ -0,0 +1,114 @@
...
+                        color=(255, 255, 255), mono_space=False, scale=2)
+        # img.draw_string(4, 54, 'Distance: %d mm' % distance.read(),
+        #                 color=(255, 255, 255), mono_space=False, scale=2)
+
+        lcd.write(img, hint=(image.BILINEAR | image.CENTER | image.SCALE_ASPECT_KEEP))
+
+
+try:
+    main()
+except OSError:
```
iabdalkader, line 107: Is it okay if we avoid wrapping the whole script in try..except ? Just to make the script look like every other one. Which part can throw an exception exactly ?

scripts/examples/00-HelloWorld/pure_thermal_main.py
kwagyeman: This is for the VLX sensor. It fails to come up on I2C often and will lock up the bus. We didn't connect the shutdown line to it, so there's no way to get the I2C bus unstuck really, except to clock the thing.

If I remove the code for it from the example completely then this is not needed. It's disabled anyway since it causes too many issues.

scripts/examples/index.csv
kwagyeman: Sure, will do.

scripts/examples/00-HelloWorld/pure_thermal_main.py
kwagyeman: The VLX sensor fails to work. So, this was me trying to make the system work by having it go into a reboot loop and clock the VLX bus on error until you boot successfully.

However, this failed. The system will reboot forever in a loop. So, I commented out the VLX code. There's no way to use it reliably if it locks up the I2C bus.

So... if I remove the VLX stuff entirely it should be fine.

kwagyeman:
@iabdalkader Updated

scripts/examples/index.csv
```diff
@@ -27,6 +27,7 @@
 "examples/09-WiFi/WINC1500", "(OPENMV2|OPENMV3|OPENMV4|OPENMV4P|OPENMVPT)", ".+", ""
 "examples/10-Bluetooth", "(OPENMV_RT1060|ARDUINO_GIGA|ARDUINO_PORTENTA_H7|ARDUINO_NICLA_VISION|ARDUINO_NANO_RP2040_CONNECT)", ".+", ""
 "examples/50-OpenMV-Boards/(?!main\.py).*$", "(OPENMV.*)", ".+", "OpenMV-Boards$"
+"examples/50-OpenMV-Boards/62-Pure-Thermal", "(OPENMVPT)", ".+", "OpenMV-Boards$"
```
iabdalkader, line 30: The pure thermal example is added to the menu for all OpenMV boards (because the rule above matches everything), this rule doesn't really do anything useful. I think we'll need to reorder the examples anyway to support board-control examples specific to imxrt.

iabdalkader:
I think the examples should be restructured like this:
```
50-OpenMV-Boards/Shields // Should have all shields examples.
50-OpenMV-Boards/STM32-Boards // current board-control and low-power (low-power is not portable because of extint).
50-OpenMV-Boards/IMXRT-Boards // imxrt board-control and low-power.
50-OpenMV-Boards/Pure-Thermal/00-HelloWorld/blinky.py
50-OpenMV-Boards/Pure-Thermal/00-HelloWorld/helloworld.py
```

Then rules I think should look like this:
```
"examples/50-OpenMV-Boards/Shields", "OPENMV.*", ".+", "OpenMV-Boards$"
"examples/50-OpenMV-Boards/STM32-Boards", "(OPENMV2|OPENMV3|OPENMV4|OPENMV4P|OPENMVPT)", ".+", "OpenMV-Boards$"
"examples/50-OpenMV-Boards/IMXRT-Boards", "OPENMV_RT1060", ".+", "OpenMV-Boards$"
"examples/50-OpenMV-Boards/Pure-Thermal/", "OPENMVPT", ".+", "OpenMV-Boards(/Pure-Thermal)?$"
```

scripts/examples/00-HelloWorld/pure_thermal_main.py
iabdalkader: Maybe move the bus recovery/try/except etc.. to the Python vlx driver ? You know have it try to recover the bus on failure, a number of time, before completely giving up and raising an error.

kwagyeman: I guess, I don't need it for the example, though, so it can be removed for right now.

kwagyeman:
Okay, I will update this when I get home tomorrow.

scripts/examples/index.csv
kwagyeman: Doesn't the "(OPENMVPT)" only match the pure thermal?

iabdalkader:
> To be clear, I should just do the work for updating the Pure Thermal in this PR...

Yes but problem is there's no easy way to display this example just for the pure thermal. The rule in line 29 matches everything inside `50-OpenMV-Boards/*` I tested it locally, and you'll have to do something very verbose to exclude it. So we need to first reorganize the scripts as suggested in https://github.com/openmv/openmv/pull/2082#issuecomment-1881161358 This will also be helpful for imxrt, to add specific board-control examples.

kwagyeman:
I see, okay, this PR will be on hold until I can create a the board specific examples PR.

iabdalkader:
I think this is no longer needed.

iabdalkader:
But it seems to have `main_py.h` so re-open if needed.

kwagyeman:
@iabdalkader - I'll update this PR today to move the file to the right location so you can merge it.

kwagyeman:
@iabdalkader Updated and ready. I tested it and it works.

Issue #2083, opened 2024-01-07
Fix Bayer when used with sensor.set_windowing()


kwagyeman:
It's bugged. Should be easy to fix. Looks like the bayer pattern is just not adjusted based on x/y pixel shift.

PR #2088, opened 2024-01-13
ports: Fix sensor bayer image with cropping and transpose.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2083

### DIFF
--- a/src/omv/ports/stm32/sensor.c
+++ b/src/omv/ports/stm32/sensor.c
@@ -976,6 +976,7 @@ int sensor_snapshot(sensor_t *sensor, image_t *image, uint32_t flags) {
         case PIXFORMAT_BAYER:
             MAIN_FB()->pixfmt = PIXFORMAT_BAYER;
             MAIN_FB()->subfmt_id = sensor->hw_flags.bayer;
+            MAIN_FB()->pixfmt = imlib_bayer_shift(MAIN_FB()->pixfmt, MAIN_FB()->x, MAIN_FB()->y, sensor->transpose);
             break;
         case PIXFORMAT_YUV422: {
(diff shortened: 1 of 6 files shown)


src/omv/ports/stm32/sensor.c
```diff
@@ -976,6 +976,7 @@ int sensor_snapshot(sensor_t *sensor, image_t *image, uint32_t flags) {
         case PIXFORMAT_BAYER:
             MAIN_FB()->pixfmt = PIXFORMAT_BAYER;
             MAIN_FB()->subfmt_id = sensor->hw_flags.bayer;
+            MAIN_FB()->pixfmt = imlib_bayer_shift(MAIN_FB()->pixfmt, MAIN_FB()->x, MAIN_FB()->y, sensor->transpose);
```
iabdalkader, line 979:
```suggestion
            MAIN_FB()->pixfmt = imlib_bayer_shift(PIXFORMAT_BAYER, MAIN_FB()->x, MAIN_FB()->y, sensor->transpose);
```

Note you're setting `MAIN_FB()->pixfmt = PIXFORMAT_BAYER;` twice, line 977 should be removed, the same goes for the other file. Otherwise this looks good to merge.

iabdalkader (changes requested):

src/omv/ports/stm32/sensor.c
kwagyeman: @iabdalkader - It's not the same. pixfmt includes the subfmt_id as a field under it. Your suggestion would zero the subfmt_id.

iabdalkader: You're right, I forgot how that worked.

iabdalkader (approved):

Issue #2081, opened 2024-01-04
Autofocus not working reliably on iMX-RT1060 using FPC camera 


kenobi986:
Auto focus using FPC OV5670 camera on the iMX-RT1060 was not working. I enabled the auto focus feature in omv_boardconfig.h and compiled the bin file and flashed into the board. I was not able to get focus at various distances for some reason. Also, I am wondering if you've examples for the auto-focus (this might be an enhancement). 

I raised the issue on the forums and here it is: 

https://forums.openmv.io/t/autofocus-not-working-reliably-on-imx-rt1060-using-fpc-camera/9061

![image](https://github.com/openmv/openmv/assets/97324805/ad92fd11-d943-43c4-87c0-97f8f1d127a0)

kwagyeman:
Thanks, I'll debug this when I get home next week.

kwagyeman:
Hi, I just tested this feature out. It seems to work. Note that the MCU on the camera decides what focus level to use. Not us. I can't comment on it's results of focusing.

```
import sensor, image, time

sensor.reset()
sensor.set_pixformat(sensor.RGB565)
sensor.set_framesize(sensor.QVGA)
sensor.skip_frames(time = 2000)

sensor.ioctl(sensor.IOCTL_TRIGGER_AUTO_FOCUS)
clock = time.clock()

while(True):
    clock.tick()
    img = sensor.snapshot()
    
    print(clock.fps())
```

This script above results in the system focusing the lens after 2 seconds. I held an item with a barcode in front of it and it did focus as expected.

You can execute sensor.ioctl(sensor.IOCTL_RESET_AUTO_FOCUS) to release the settings or sensor.ioctl(sensor.IOCTL_TRIGGER_AUTO_FOCUS) to focus again.

You cannot call the function in a loop though. The camera can't just continually auto focus. It's a one shot thing you can call every now and then.

kwagyeman:
As for wait on auto focus... that doesn't seem to be that useful.

read reg 0x3029 and see if it's:

0x70 -> IDLE
0x00 -> Focusing
0x10 -> Focused

You'll see it go from 0 to 0x10. The default value will be 0x70.= before anything is executed.

PR #2090, opened 2024-01-13
boards: Enable OV5640 Autofocus firmware and add example.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2081

Now that JPEG is merged we should have space for the firmware since the software decoder is gone for the H7 line.

### DIFF
--- a/scripts/examples/01-Camera/07-Sensor-Control/autofocus.py
+++ b/scripts/examples/01-Camera/07-Sensor-Control/autofocus.py
@@ -1,0 +1,28 @@
+# This work is licensed under the MIT license.
+# Copyright (c) 2013-2024 OpenMV LLC. All rights reserved.
+# https://github.com/openmv/openmv/blob/master/LICENSE
+#
+# Auto Focus Example
+#
+# This example shows off how to auto-focus on cameras that support the feature.
+
+import sensor
+import time
+
+sensor.reset()  # Reset and initialize the sensor.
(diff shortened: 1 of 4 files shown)


iabdalkader:
Can you please filter this example by sensor ? The rules don't make that easy, so you can also just move it to a new directory `09-OV5640/autofocus.py`.

Also please remove `sensor_` prefix from all scripts in `07-Sensor-Control/*`

iabdalkader:
@kwagyeman alternatively you can keep it in the same place, but please rename to `autofocus.py` and check if `IOCTL_TRIGGER_AUTO_FOCUS` ioctl attr exists. We may have other auto-focus sensors in the future.

kwagyeman:
@iabdalkader Did you second idea. Note, I had to use try/except because if you change the camera module don't the board to something which doesn't support autofocus the attribute still exists but will fail to execute.

iabdalkader:
> something that doesn't support autofocus, the attribute still exists but will fail to execute.

Good point. Don't we have runtime sensor id `sensor.get_id()` ?

iabdalkader:
Merged. Can you please remove `sensor_` prefix from all scripts in `07-Sensor-Control/* and `_control`. So
`sensor_auto_gain_control.py` -> `auto_gain.py` in a follow-up PR.

kwagyeman:
Will do.

Issue #2091, opened 2024-01-14
Fix morph arg


kwagyeman:
arg number 5 is int, yet every call passes it as a float

iabdalkader:
Fixed in #2097

PR #2095, opened 2024-01-16
imlib: Fix gcc13.2 compile issues.

kwagyeman:
Fixes https://github.com/openmv/openmv/issues/2091 too.


iabdalkader:
I looked into this some more, and the warnings in `fmath.c` are actually caused by strict aliasing (the wrong warning is outputted maybe because the file suppresses strict aliasing warnings with `pragma`). I removed all the pragma directives and added specific rules for `fmath.c` to disable strict aliasing, and merged this branch in  #2097

PR #2111, opened 2024-01-24
sensors/paj6100: Fix driver for RT1062.

kwagyeman:
The PJA6100 is the only camera which uses an sensor.hw_flags.hsync of 1. All other sensors are 0. That's why this bug in the driver went undetected.

The change in the PAJ6100 register list just resets those registers in the driver back to their chip defaults.

Firmware still works with the H7.


iabdalkader:
Excellent work! Thank you!

PR #2114, opened 2024-01-25
ports: Fix sensor abstraction layer comment.

kwagyeman:


src/omv/ports/mimxrt/sensor.c
```diff
@@ -1,12 +1,12 @@
...
  *
- * Copyright (c) 2013-2023 Ibrahim Abdelkader <iabdalkader@openmv.io>
- * Copyright (c) 2013-2023 Kwabena W. Agyeman <kwagyeman@openmv.io>
+ * Copyright (c) 2013-2024 Ibrahim Abdelkader <iabdalkader@openmv.io>
+ * Copyright (c) 2013-2024 Kwabena W. Agyeman <kwagyeman@openmv.io>
  *
  * This work is licensed under the MIT license, see the file LICENSE for details.
  *
- * Sensor abstraction layer for nRF port.
+ * Sensor abstraction layer for mimxrt port.
```
iabdalkader, line 9:
```suggestion
 * Sensor driver for mimxrt port.
```

This is the implementation of the sensor driver, for each port. It's just a comment that was copied from old code.

iabdalkader (changes requested):

src/omv/ports/mimxrt/sensor.c
kwagyeman: Yep

PR #2116, opened 2024-01-25
ports/all: Allow sensor_abort() to be called from different contexts.

iabdalkader:
This patch removes the duplicated `sensor_abort()` logic in the stm32 and mimxrt ports. This is done by adding a `flush_fifo` argument, which controls the flush operation, and a `in_irq` argument , which indicates whether `sensor_abort()` is called from within an IRQ context or not.

### DIFF
--- a/src/omv/ports/stm32/sensor.c
+++ b/src/omv/ports/stm32/sensor.c
@@ -250,7 +250,7 @@ int sensor_abort() {
         sensor.last_frame_ms_valid = false;
     }
 
-    framebuffer_reset_buffers();
+    framebuffer_flush_buffers(true);
 
     return 0;
 }
@@ -488,7 +488,7 @@ void DCMI_DMAConvCpltUser(uint32_t addr) {
         sensor.last_frame_ms_valid = false;
         // Reset the queue of frames when we start dropping frames.
         if (!sensor.disable_full_flush) {
-            framebuffer_flush_buffers();
+            framebuffer_flush_buffers(false);
         }
         return;
     }
--- a/src/omv/ports/stm32/sensor.c
+++ b/src/omv/ports/stm32/sensor.c
@@ -97,9 +97,8 @@ static int sensor_dma_config() {
     // Set DMA IRQ handle
     dma_utils_set_irq_descr(DMA2_Stream1, &DMAHandle);
 
-    // Configure and enable DMA IRQ Channel
+    // Configure the DMA IRQ Channel
     NVIC_SetPriority(DMA2_Stream1_IRQn, IRQ_PRI_DMA21);
-    HAL_NVIC_EnableIRQ(DMA2_Stream1_IRQn);
     return 0;
 }
 
@@ -109,7 +108,7 @@ void sensor_init0() {
     DCMI_MDMA_Handle1.Instance = MDMA_CHAN_TO_INSTANCE(OMV_MDMA_CHANNEL_DCMI_1);
     #endif
 
-    sensor_abort();
+    sensor_abort(true, false);
 
     // Re-init i2c bus to reset the bus state after soft reset, which
     // could have interrupted the bus in the middle of a transfer.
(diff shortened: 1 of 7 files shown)


src/omv/ports/stm32/sensor.c
```diff
@@ -229,13 +228,17 @@ int sensor_dcmi_config(uint32_t pixformat) {
...
-        HAL_NVIC_EnableIRQ(DMA2_Stream1_IRQn);
+        if (in_irq) {
+            HAL_DMA_Abort_IT(&DMAHandle);
+        } else {
+            HAL_DMA_Abort(&DMAHandle);
+        }
+        HAL_NVIC_DisableIRQ(DMA2_Stream1_IRQn);
         #if defined(OMV_MDMA_CHANNEL_DCMI_0)
         HAL_MDMA_Abort(&DCMI_MDMA_Handle0);
         HAL_MDMA_Abort(&DCMI_MDMA_Handle1);
```
kwagyeman, line 244:
```suggestion
        if (!in_irq) {
            HAL_MDMA_Abort(&DCMI_MDMA_Handle0);
            HAL_MDMA_Abort(&DCMI_MDMA_Handle1);
        }
```

kwagyeman: Looks good otherwise.

kwagyeman:
@iabdalkader Changing the "HAL_NVIC_EnableIRQ(DMA2_Stream1_IRQn);" logic int the STM32 driver seems to break things.

```
import sensor, image, time

sensor.reset()
sensor.set_framebuffers(1)
sensor.skip_frames(time = 2000)

clock = time.clock()

while(True):
    clock.tick()
    img = sensor.snapshot()
    print(clock.fps())
```

Just crashes. If I restore the normal behavior of that IRQ by doing:

```
    if (DCMI->CR & DCMI_CR_ENABLE) {
        DCMI->CR &= ~DCMI_CR_ENABLE;
        if (in_irq) {
            HAL_DMA_Abort_IT(&DMAHandle);
        } else {
            HAL_DMA_Abort(&DMAHandle);
            HAL_NVIC_EnableIRQ(DMA2_Stream1_IRQn);
        }
        #if defined(OMV_MDMA_CHANNEL_DCMI_0)
        if (!in_irq) {
            HAL_MDMA_Abort(&DCMI_MDMA_Handle0);
            HAL_MDMA_Abort(&DCMI_MDMA_Handle1);
        }
        HAL_MDMA_DeInit(&DCMI_MDMA_Handle0);
        HAL_MDMA_DeInit(&DCMI_MDMA_Handle1);
        #endif
        __HAL_DCMI_DISABLE_IT(&DCMIHandle, DCMI_IT_FRAME);
        __HAL_DCMI_CLEAR_FLAG(&DCMIHandle, DCMI_FLAG_FRAMERI);
        first_line = false;
        drop_frame = false;
        sensor.last_frame_ms = 0;
        sensor.last_frame_ms_valid = false;
    }
```

Then is works as expected. I think I'd feel more comfortable leaving the behavior as is. It is possible to change it, though... I just need to reload a lot of the driver state into my head again to workout why the failure is happening. I don't have time today though to focus on this.

kwagyeman:
IMXRT works as expected.

src/omv/ports/stm32/sensor.c
iabdalkader: That doesn't seem to have any effect either. I think it's safe to abort it unconditionally, but I can make that change if it's necessary.

iabdalkader:
> Changing the "HAL_NVIC_EnableIRQ(DMA2_Stream1_IRQn);" logic int the STM32 driver seems to break things.
> 
> ```
> import sensor, image, time
> 
> sensor.reset()
> sensor.set_pixformat(sensor.RGB565)
> sensor.set_framesize(sensor.QVGA)
> sensor.set_framebuffers(1)
> sensor.skip_frames(time = 2000)
> 
> clock = time.clock()
> 
> while(True):
>     clock.tick()
>     img = sensor.snapshot()
>     print(clock.fps())
> ```

I think I just needed to enable the DMA IRQ before `HAL_DCMI_Start_DMA_xx`. I tested this with both OPENMV4/MT9M and OPENMV4P/OV5640,  and they both seem to work fine (with 1 frame buffer or more). Can you test again ?

src/omv/ports/stm32/sensor.c
kwagyeman: The aborts technically have an infinite loop in them. So, you want to not call them in an IRQ.

iabdalkader: Good point, fixed.

kwagyeman:
@iabdalkader - That fixed it. All good.

PR #2120, opened 2024-01-31
modules/py_fir: Fix mlx data root pointer.

kwagyeman:


iabdalkader:
While testing I noticed this causing serious issues because:
1) bss is no longer scanned so it can be collected at any time.
2) `gc_sweep()` will collect it on soft-reboot, and then you have a dangling pointer, which it will try to collect again.

PR #2123, opened 2024-02-01
ports/mimxrt: Enable edma memcpy.

kwagyeman:
Enables EDMA memcpy() on the RT1062. I wrote this first to understand better how edma works. I think I see a path to getting full offload operation. This PR improves performance by 15%. Full offload should provide much more.

Also, I found a way to offload the byte swap to the CSI hardware.

### DIFF
--- a/src/omv/boards/OPENMV_RT1060/omv_boardconfig.h
+++ b/src/omv/boards/OPENMV_RT1060/omv_boardconfig.h
@@ -192,7 +192,12 @@
 
 // Camera interface configuration.
 #define OMV_CSI_BASE                    (CSI)
+#define OMV_CSI_DMA                     (DMA0)
+#define OMV_CSI_DMA_MUX                 (DMAMUX)
+#define OMV_CSI_DMA_CHANNEL_START       (4U)
+#define OMV_CSI_DMA_CHANNEL_COUNT       (2U)
 #define OMV_CSI_XCLK_FREQUENCY          (12000000)
+#define OMV_CSI_DMA_MEMCPY_ENABLE       (1)
 #define OMV_CSI_HW_SWAP_ENABLE          (1)
 
 #define OMV_CSI_D0_PIN                  (&omv_pin_DCMI_D0)
--- a/src/omv/ports/mimxrt/sensor.c
+++ b/src/omv/ports/mimxrt/sensor.c
@@ -24,14 +24,21 @@
 #include "framebuffer.h"
 #include "unaligned_memcpy.h"
 
+#define DMA_LENGTH_ALIGNMENT     (8)
+#define SENSOR_TIMEOUT_MS        (3000)
+
 // Sensor struct.
 sensor_t sensor = {};
 extern uint8_t _line_buf[OMV_LINE_BUF_SIZE];
 
+#if defined(OMV_CSI_DMA)
+static edma_handle_t CSI_EDMA_Handle[OMV_CSI_DMA_CHANNEL_COUNT];
+static int src_inc, src_size, dest_inc_size;
+#endif
+
 #define CSI_IRQ_FLAGS    (CSI_CR1_SOF_INTEN_MASK            \
                           | CSI_CR1_FB2_DMA_DONE_INTEN_MASK \
                           | CSI_CR1_FB1_DMA_DONE_INTEN_MASK)
-//CSI_CR1_RF_OR_INTEN_MASK
 
 void sensor_init0() {
     sensor_abort(true, false);
@@ -187,6 +194,39 @@ void sensor_sof_callback() {
     }
 }
 
+#if defined(OMV_CSI_DMA)
+int sensor_dma_memcpy(void *dma, void *dst, void *src, int bpp, bool transposed) {
+    // EMDA will not perform burst transfers for anything less than 32-byte chunks made of four 64-bit
+    // beats. Additionally, the CSI hardware lacks cropping so we cannot align the source address.
+    // Given this, performance will be lacking on cropped images. So much so that we do not use
+    // the EDMA for anything less than 4-byte transfers otherwise you get sensor timeout errors.
+    if (dest_inc_size < 4) {
+        return -1;
+    }
+
+    edma_handle_t *handle = dma;
+    edma_transfer_config_t config;
+    EDMA_PrepareTransferConfig(&config,
+                               src, // srcAddr
+                               src_size, // srcWidth
+                               src_inc, // srcOffset
+                               dst, // destAddr
+                               transposed ? bpp : dest_inc_size, // destWidth
+                               transposed ? (MAIN_FB()->v * bpp) : dest_inc_size, // destOffset
+                               MAIN_FB()->u * bpp, // bytesEachRequest
+                               MAIN_FB()->u * bpp); // transferBytes
+
+    // Drop the frame if EDMA is not keeping up as the image will be corrupt.
+    if (EDMA_SubmitTransfer(handle, &config) != kStatus_Success) {
+        sensor.drop_frame = true;
+        return 0;
+    }
+
+    EDMA_TriggerChannelStart(handle->base, handle->channel);
+    return 0;
+}
+#endif
+
 void sensor_line_callback(uint32_t addr) {
     // Throttle frames to match the current frame rate.
     sensor_throttle_framerate();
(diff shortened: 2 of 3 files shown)


iabdalkader:
This should be rebased on  #2128

kwagyeman:
Done, on to doing full offload.

kwagyeman:
Okay, did some testing and debugging.

Enabling full offload via using the CSI embedded DMA controllers provides a 47% performance increase. The frame rate at 1280x720 with grayscale streaming jpegs goes from 19 to 28 fps.

It appears the DMA controllers in the CSI hardware are talking to a 64-bit AHB bus that connects to the main 64-bit AXI system that can talk to SDRAM. These controllers can do a INC16 burst at 8-bytes a piece from the RXFIFO. It's no clear how deep the FIFO is... either 128 words or 128 double words. Anyway, interestingly, the DMA controllers handle short counts. Using INC16 doesn't appear to force the total image data size to be a multiple of 128 bytes.

I tried out different INC sizes and found a 1 FPS difference from INC4 to INC16. So, making that as large as possible is important.

Moving on, I'm going to try to get full offload with eDMA working. However, I think it will be slower. eDMA is on a 64-bit AHB bus too. However, it's max burst size is INC4 with 32-byte transfers. It also doesn't handle short counts at all. So, using it will force the image size to be a multiple of that or suffer a huge slowdown. Except for 32-byte bursts it does everything less than that as single non-burst 8/16/32/64-bit transfers which will be terrible for the SDRAM.

iabdalkader:
Hi, I understand this is a draft, but I have some early feedback: The first commit doesn't make much sense to me. A board either has support for DMA memcpy (in which case it's used) or it doesn't. There's no fallback to software in runtime, but the code isn't even trying to do that.

The second commit can be sent in a separate PR, and merged first, but it's also Not a runtime option, it's a fixed hardware characteristic, so it should be a define (`OMV_CSI_HAS_BSWAP`) or something.

Note we don't have a "common" category in the change log, but feel free to add one github/workflows/changelog.json, or use hal.

kwagyeman:
Yeah, you have to fall back to software memcpy though for the IMXRT because the EDMA performance is so bad. You get timeouts from using it for copying lines if the line start/offset aren't a multiple of 4 bytes.

So, I have it setup such that if the number of bytes it's moving is less than 4 you fallback to software.

This brings up though if it even makes sense to use it at all. It technically can provide a minor speed up for copying lines in situations where you are cropping at a multiple of 4 bytes.

The edma on the IMXRT will only do burst transactions for 32 byte aligned 32 byte transfers. Everything else is done as a single read/write using 8/16/32/64 bit widths.

kwagyeman:
Okay, trying to use the eMDA went nowhere. I couldn't figure out how to use it. So, I fell back and used the embedded DMA controllers. This provides a 47% performance boost. Setting the frame rate also works in full offload mode.

src/omv/boards/OPENMV_RT1060/omv_boardconfig.h
```diff
@@ -211,4 +212,11 @@
 #define OMV_CSI_POWER_PIN               (&omv_pin_DCMI_POWER)
 #define OMV_CSI_FSYNC_PIN               (&omv_pin_DCMI_FSYNC)
 
+#define OMV_CSI_HAS_BSWAP               (1)
+
+#define OMV_CSI_DMA                     (DMA0)
+#define OMV_CSI_DMA_MUX                 (DMAMUX)
+#define OMV_CSI_DMA_CHANNEL_START       (4U)
+#define OMV_CSI_DMA_CHANNEL_COUNT       (2U)
```
iabdalkader, line 220: If you'll keep those options please move them after `OMV_CSI_BASE`

kwagyeman: Done

src/omv/ports/mimxrt/sensor.c
```diff
@@ -181,13 +184,74 @@ void sensor_sof_callback() {
...
+    // Clear the FIFO and re/enable DMA.
+    CSI_REG_CR3(CSI) |= (CSI_CR3_DMA_REFLASH_RFF_MASK | CSI_CR3_DMA_REQ_EN_RFF_MASK);
+}
+
+#if defined(OMV_CSI_DMA)
+int sensor_dma_memcpy(void *dma, void *dst, void *src, int bpp, bool transposed) {
+    // EMDA will not perform burst transfers for anything less than 32-byte chunks made of four 64-bit
+    // beats. Additionally, the CSI hardware lacks cropping so we cannot align the source address.
+    // Given this, performance will be lacking on cropped images. So much so that we we do not use
+    // the EDMA for anything less than 4-byte transfers...
```
iabdalkader, line 212:
```suggestion
    // Given this, performance will be lacking on cropped images. So much so that we do not use
    // the EDMA for anything less than 4-byte transfers.
```

iabdalkader (changes requested):

src/omv/ports/mimxrt/sensor.c
```diff
@@ -24,14 +24,21 @@
...
+#define SENSOR_TIMEOUT_MS        (3000)
+
 // Sensor struct.
 sensor_t sensor = {};
 extern uint8_t _line_buf[OMV_LINE_BUF_SIZE];
 
+#if defined(OMV_CSI_DMA)
+static edma_handle_t CSI_EDMA_Handle[OMV_CSI_DMA_CHANNEL_COUNT];
+static int src_inc, src_size, dest_inc_size;
+#endif
```
iabdalkader, line 37: There's a way to add variables like those directly in `sensor_t` struct, see `OMV_SPI_PORT_BITS`

iabdalkader:
I don't see any noticeable performance gain with the 3rd commit enabled, is it for a special case ? Everything else looks okay. If you drop the 3rd commit this PR should move much faster, and maybe we can revisit it later.

iabdalkader:
> The 3rd commit offers all the performance gain.
> 
> Set the resolution to sensor.HD with RGB565 and then compare before and after FPS. You'll notice the FPS increase by 47%.

I did, again, and I still don't see any noticeable performance gain with the 3rd commit enabled.

kwagyeman:
Weird. Let me check.

src/omv/ports/mimxrt/sensor.c
kwagyeman: Fixed the comment.

kwagyeman:
Okay, I ran a test using something particularly processor-intensive and noticed a performance improvement:

```
from ulab import numpy as np
import sensor
import time
import math

# Set to 0 for a grayscale image. Set above 1.0 to pump-up the saturation.
UV_SCALE = 1.0

sensor.reset()
sensor.set_framesize(sensor.WVGA)
sensor.skip_frames(time=2000)

# These are the standard coefficents for converting RGB to YUV.
rgb2yuv = np.array([[    0.299,     0.587,     0.114], # noqa
                    [-0.168736, -0.331264,       0.5], # noqa
                    [      0.5, -0.418688, -0.081312]], dtype=np.float) # noqa

# Now get the inverse so we can get back to RGB from YUV.
yuv2rgb = np.linalg.inv(rgb2yuv)

clock = time.clock()

# r will be the angle by which we rotate the colors on the UV plane by.
r = 0

while True:
    clock.tick()

    # Increment in a loop.
    r = (r + 1) % 360
    a = math.radians(r)

    # This is a rotation matrix which we will apply on the UV components of YUV values.
    # https://en.wikipedia.org/wiki/Rotation_matrix
    rot = np.array([[1,           0,            0], # noqa
                    [0, math.cos(a), -math.sin(a)], # noqa
                    [0, math.sin(a),  math.cos(a)]], dtype=np.float) # noqa

    # This is the scale matrix
    scale = np.array([[1,        0, 0], # noqa
                      [0, UV_SCALE, 0], # noqa
                      [0,        0, UV_SCALE]], dtype=np.float) # noqa

    # Now compute the final matrix using matrix multiplication.
    m = np.dot(yuv2rgb, np.dot(scale, np.dot(rot, rgb2yuv)))

    # Apply the color transformation (m.flatten().tolist() also works)
    img = sensor.snapshot().ccm(m.tolist())

    print(clock.fps())
```

You get 10.15 FPS with the 3rd commit enabled and 9.44 fps with it disabled. I guess this is because I used 32-byte eDMA bust transactions when everything is setup to do so.

So, it's 7.5% faster... that's probably the difference in CPU time used to setup the eDMA.

I'd prefer to keep all speed we can get. I know that the 3rd commit is absolutely the fastest thing you can do. But, it's also... minimal.

iabdalkader (approved):

Issue #2024, opened 2023-11-25
Refactor py_tf.


iabdalkader:
There should be exactly:
- 1 libtf_invoke internal function
- 1 tf_input_callback_data
- 1 tf_output_callback_data.

Segment and detect should be separate functions.

PR #2133, opened 2024-02-07
modules/py_tf: Refactor regression() code to handle arrays correctly.

kwagyeman:
This PR does a few things:

Different ops are now used by the TensorFlow library depending on the board you are using. For the RT1062 and forward, we enable 78 tensorflow ops, up from 34 being enabled previously. This uses 200 KB more flash on the RT1062. However, we have plenty on that system. On the other boards, I increased the number of ops from 34 to 38 to support YOLO.

The regression code has been refactored to accept standard Python arrays. The arrays can either be flattened or have levels. This removes the code being dependent on ulab, which we do not want to tie our code base closely to. ndarrays support a method that outputs their internal data as a list trivially so it's a minor fix to make this work with the new API.

This PR fixes: https://github.com/openmv/openmv/issues/2024

```
m = tf.load("force_float32_quant.tflite")
print(m)
i = [-3, -1, -2,  5, -2, 10, -1,  9,  0,  2,  0,  9,  1, 10,  2, -1,  3,
     5,  3,  9,  3,  9,  6,  2,  6,  7,  5, 10,  6, -1,  7,  4,  7,  8,
     5,  7]
print(len(i))

print(tf.regression(m, I))
# 54.4129
```

Model: https://drive.google.com/file/d/13z1Trd78ExMKRGmAtE2cFbRVcSVsT8EK/view?usp=sharing

```
m = tf.load("force_int_quant.tflite")
print(m)
i = [-3, -1, -2,  5, -2, 10, -1,  9,  0,  2,  0,  9,  1, 10,  2, -1,  3,
     5,  3,  9,  3,  9,  6,  2,  6,  7,  5, 10,  6, -1,  7,  4,  7,  8,
     5,  7]
print(len(i))

print(tf.regression(m, I))
# 53.7833
```

Model: https://drive.google.com/file/d/18rv0yw1m6ohoKAibzMwRlaI1Ptnq1izu/view?usp=sharing


iabdalkader:

https://forums.openmv.io/t/using-cmsis-nn-for-imxrt1060/9200/4

iabdalkader:
> Finally, on the NICLA and PORTETNA I reduced the number of ops so that the firmware can compile.

So those 2 boards won't run YOLO and were switched to minimal ops ? Or were they switched to minimal ops to support YOLO ?

src/omv/modules/py_tf.c
```diff
@@ -774,6 +728,107 @@ STATIC mp_obj_t py_tf_detect(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw
...
+                        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Channel count mismatch!"));
+                    }
+                }
+            } else {
+                mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Column count mismatch!"));
+            }
+        }
+    } else {
+        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Row count mismatch!"));
+    }
```
iabdalkader, line 803: Can you do these sanity checks early in the function so it's more readable ?

src/omv/modules/py_tf.h
```diff
@@ -26,4 +27,15 @@ extern size_t py_tf_putchar_buffer_index;
...
+#if IMLIB_ENABLE_TF == IMLIB_TF_FULLOPS
+#define libtf_get_parameters libtf_all_ops_get_parameters
+#define libtf_invoke libtf_all_ops_invoke
+#elif IMLIB_ENABLE_TF == IMLIB_TF_DEFAULT
+#define libtf_get_parameters libtf_reduced_ops_get_parameters
+#define libtf_invoke libtf_reduced_ops_invoke
+#elif IMLIB_ENABLE_TF == IMLIB_TF_MINIMAL
+#define libtf_get_parameters libtf_minimum_ops_get_parameters
+#define libtf_invoke libtf_minimum_ops_invoke
+#endif
```
iabdalkader, line 40: While we're at it, I would prefer if the names match the options.

```suggestion
#if IMLIB_ENABLE_TF == IMLIB_TF_FULLOPS
#define libtf_get_parameters libtf_get_parameters_fullops
#define libtf_invoke libtf_invoke_fullops
#elif IMLIB_ENABLE_TF == IMLIB_TF_DEFAULT
#define libtf_get_parameters libtf_get_parameters_default
#define libtf_invoke libtf_invoke_default
#elif IMLIB_ENABLE_TF == IMLIB_TF_MINIMAL
#define libtf_get_parameters libtf_get_parameters_minimal
#define libtf_invoke libtf_invoke_minimal
#endif
```

iabdalkader (changes requested):

src/omv/imlib/imlib.h
```diff
@@ -30,6 +30,10 @@
 #include "imlib_config.h"
 #include "omv_boardconfig.h"
 
+#define IMLIB_TF_MINIMAL        (1)
+#define IMLIB_TF_DEFAULT        ((IMLIB_TF_MINIMAL) +1)
+#define IMLIB_TF_FULLOPS        ((IMLIB_TF_DEFAULT) +1)
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/modules/py_tf.c
kwagyeman: I would have to duplicate the same if statements and logic to do this. Are you sure? For the nested raises I don't have the information I need until I call mp_obj_get_array.

iabdalkader: Actually why not just support a flat array ? One can easily flatten the array in Python with ulab ndarray.flatten, or pure Python:
```python
[j for sub in [[1,2,3],[4,5,6],[7,8,9]] for j in sub]
```

kwagyeman: Seems like this would be less user-friendly. I've noticed a few different forum threads and folks opening up issue posts about this function such that it's worth it to make it handle any array people could throw at it. Right now, it will handle a 1d array, a 2d array, and a 3d array without issue.

iabdalkader: Yeah but it's more text, code adds up and then we run out of flash. I'll merge it though.

iabdalkader (approved):

PR #2141, opened 2024-02-12
ports/all: Add a sensor function to reconfigure hardware if/when needed.

iabdalkader:
sensor_dcmi_config function was called every time the pixel format changed, and typically reconfigured the capture interface. This originated from the stm32 port, which required reconfiguring the DCMI if the pixel format changed. However, most ports only require configuring the capture interface once during initialization (for example nrf, rp2 and mimxrt ports).
The new sensor_config function is called after a setting that may require reconfiguring the hardware changes, such as such as window size, frame size, or pixel format. The exact setting that has changed is passed to the function, allowing ports to decide whether to do a full reconfiguration of the hardware, or just ignore it based on the changed setting.

Note on every reset, stm32 configured the DCMI twice in a row, in `sensor_init`, and later in `set_pixformat`. The mimxrt port configured CSI on every soft-reset, when it only needs to configure it once


iabdalkader:
>> This originated from the stm32 port, which required reconfiguring the DCMI if the pixel format changed.

I wonder if we even need to do that, or just configure once, and then flip a bit on pixel format change. Either way, this PR is still very useful.

kwagyeman (approved):
Yeah, it all looks okay. I assume you tested that JPEG mode works still with the STM32H7 and non-JPEG mode works still? If that's correct then it looks good.

iabdalkader:
> Yeah, it all looks okay. I assume you tested that JPEG mode works still with the STM32H7 and non-JPEG mode works still? If that's correct then it looks good.

Yes I tested it, switched between JPEG, RGB, Grayscale, JPEG again etc.. a few times all works, but if you can think of something else that we should test let me know.

iabdalkader:
Note the most of the changes are just indenting code, no actual code has been changed.

kwagyeman:
No, I think it's fine then.

PR #2061, opened 2023-12-29
modules/py_image: Optimize and cleanup all math and binary line ops.

kwagyeman:
The goal of this PR was to cleanup all the line op functions and save firmware space. This PR accomplishes that saving 5kB, optimizes the line op functions with cortex SIMD, and vastly increases functionality by moving line ops to be powered by draw_image().

Fixes: https://github.com/openmv/openmv/issues/1855

For binary functions:
- binary: Now powered by draw image for backend removing a huge amount of code.
- invert: Optimized the loop for 2x/4x rgb565/grayscale speedup.
- and/nand/or/nor/xor/xnor: Now powered by draw image and optimized the loops for 2x/4x rgb565/grayscale speedup.
- Erode/Dilate: Changed Erode threshold argument meaning to match dilate/close/open/top_hat/black_hat.

For imlib:
- Removed the old image_operation runner function and it's helper functions.
- Moved ksize_to_n from py_helper to imlib.

For math functions:
- negate(): Removed as it did the same thing as invert(). Negate() in py_image points to invert().
- Replace(): Removed as draw_image() does this now. Points to draw_image().
- Add/Sub/Mul: Now powered by draw image and optimized the loop for 2x/4x rgb565/grayscale speedup.
- RSub/IMul: Added as new methods that breakout the old reverse/invert arguments of sub/mul.
- div: Removed. Didn't make sense to keep. It does a poor job blending images and can't be made fast.
- Difference: Now powered by draw image and optimized the loop for 2x/4x rgb565/grayscale speedup.
- Blend: Removed as draw_image provides this. Blend() in py_image points to draw_image().
- Top_hat()/Black_hat(): Moved to mathops.c and cleaned up code to save space. 

For stats functions:
- getSimilarity(): Now powered by draw_image(). Also, cleaned up the code. I had to include this code change in this PR to remove the image_operation function as getSimilarity() also depended upon it.

In py_image:
- Draw image() re-written for mp_arg_parse_all and used to power lineops functions.
- All binary methods have had their wrappers re-written for mp_arg_parse_all.
- All mathops methods have had their wrappers re-written for mp_arg_parse_all.
- Get similarity wrapper re-written to use mp_arg_parse_all.

Things to do in the future:
- Binary()'s thresholding code needs to be improved.
- Erode/Dilate: Code needs be cleaned up and there are some speed gains to be had.
[...]

PR #2143, opened 2024-02-12
imlib: Remove mul().

kwagyeman:
blend() is more useful than this feature. mul() effectively implements the image burn operation. It also implements the image screen blending operation when used with the invert flag.

While mul can definitely benefit from performance improvements in the future through SIMD... it isn't really needed. burn/screen are image blending operations which is something for photoshop. Not the OpenMV Cam.

Broke out from: https://github.com/openmv/openmv/pull/2061


iabdalkader:
This needs to be rebased to resolve conflicts.

kwagyeman:
@iabdalkader Updated.

PR #2147, opened 2024-02-13
imlib: Remove cartoon() method.

kwagyeman:
Was disabled on every board already. Just dead code.


iabdalkader:
But why not keep it and enable it for boards that have flash ?

kwagyeman:
The method actually doesn't work. You need a CNN using style transfer to actually do this correctly. Right now, it does a poor job of color filling areas.

PR #2148, opened 2024-02-13
modules/py_image: Use mp_arg_pasrse_all for erode/dilate.

kwagyeman:
Update erode/dilate/close/open/black_hat/top_hat to use mp_arg_pasrse_all().

Also, the meaning of the optional threshold value for erode has been changed to match how dilate's threshold works.

Broke out from: https://github.com/openmv/openmv/pull/2061

### DIFF
--- a/src/omv/modules/py_image.c
+++ b/src/omv/modules/py_image.c
@@ -2118,67 +2118,50 @@ STATIC mp_obj_t py_image_b_xnor(uint n_args, const mp_obj_t *args, mp_map_t *kw_
 }
 STATIC MP_DEFINE_CONST_FUN_OBJ_KW(py_image_b_xnor_obj, 2, py_image_b_xnor);
 
-STATIC mp_obj_t py_image_erode(uint n_args, const mp_obj_t *args, mp_map_t *kw_args) {
-    int arg_ksize =
-        py_helper_arg_to_ksize(args[1]);
-    int arg_threshold =
-        py_helper_keyword_int(n_args, args, 2, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_threshold),
-                              imlib_ksize_to_n(arg_ksize) - 1);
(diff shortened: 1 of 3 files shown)


src/omv/modules/py_image.c
```diff
@@ -2118,65 +2118,48 @@ STATIC mp_obj_t py_image_b_xnor(uint n_args, const mp_obj_t *args, mp_map_t *kw_
...
-        py_helper_arg_to_ksize(args[1]);
-    int arg_threshold =
-        py_helper_keyword_int(n_args, args, 2, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_threshold),
-                              0);
-    image_t *arg_msk =
-        py_helper_keyword_to_image(n_args, args, 3, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_mask), NULL);
+STATIC mp_obj_t py_image_erode(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    return py_image_binary_moprh_op(n_args, pos_args, kw_args, imlib_erode);
+}
+STATIC MP_DEFINE_CONST_FUN_OBJ_KW(py_image_erode_obj, 2, py_image_erode);
```
iabdalkader, line 2149:
```suggestion
STATIC MP_DEFINE_CONST_FUN_OBJ_KW(py_image_erode_obj, 1, py_image_erode);
```
I wasn't sure about this before but I've seen more examples. This should be the minimum number of positional args. Required keyword args should be enforced with `MP_ARG_REQUIRED`. This way if you make `ksize` non-required, you only need to change one place.

iabdalkader (changes requested):

src/omv/modules/py_image.c
kwagyeman: Done

iabdalkader (approved):

Issue #2070, opened 2024-01-02
Debug Issue with jpeg file data causing some desktop programs not to be able to handle it.


kwagyeman:
Our JPEG images have some issue with their data that causes issues. Determine what that is and fix it.

kwagyeman:
On comparison with our STM32 JPEG compressor and the OV5640 (which displays correctly on desktop programs), there is no difference in the JPEG headers. So, the issue is somewhere in the scanline data. Our software compressor has the same issue as the STM32 one... FFMEG throws an error about extra bits. So, will hunt down this now.

![ours](https://github.com/openmv/openmv/assets/5694981/4650f5ee-ada9-4f40-88b6-fee8ebb33954)
![sensor](https://github.com/openmv/openmv/assets/5694981/0cc93736-bf2e-490b-a387-facfb295f633)

kwagyeman:
MJPEG video files created by the OV5640 or our jpeg compressor are treated the same by FFMPEG and VLC. They do not have errors. So, the issue must be related to our programs handle JPEG RTSP streams.

So, I would say our jpeg file format is fine. This is an RTSP specific issue.

kwagyeman:
Okay, determined the issue.

Only FFMPEG can play GRYSCALE MJPEG streams and YUV444/YUV422/YUV420 RGB streams. Other players like VLC can only handle YUV422 RGB Streams (not YUV420).

The reason the OV5640 works versus our JPEG is that does YUV422 JPEGs versus YUV444 JPEGs.

...

Note, VLC can decode YUV444/Grayscale/YUV420 JPEGs fine when reading an MJPEG file video. However, it cannot handle them in an RTSP stream.

...

Not sure what other players problems are, but, it's probably the same issue. They lack code to handle anything but YUV422 RGB which is the most common format.

...

What is the fix for this? I guess we can default JPEG compression to YUV422. But, also add some setting maybe to force YUV422 when necessary.

iabdalkader:
This issue has been mentioned on **OpenMV Forums**. There might be relevant details there:

https://forums.openmv.io/t/rt1062-bugs-in-examples-april-tag-mjpeg-streamer-rtsp-video-streamer-wlan/9036/8

kwagyeman:
Will wait on hardware jpeg to be merged before sending a PR to handle this.

PR #2153, opened 2024-02-16
ports/stm32: Add support for YUV422 hardware JPEG compression.

kwagyeman:
This PR adds support for YUV422 compression in the hardware jpeg decoder on the STM32. YUV422 compression is slightly slower than YUV444. However, it's necessary to support as quite a lot of desktop software cannot process YUV444 JPEGs in MJPEG streams.

The second commit in this PR then changes the default compression level for YUV422 to turn on at 70 percent quality and adjusts all MJPEG streaming scripts to match that. This should fix all MJPEG streaming issues across all OpenMV Cams on most platforms.

Fixes: https://github.com/openmv/openmv/issues/2070

...

For example, the RTSP streamer to VLC doesn't work without this PR. Once you add this PR it works then like magic.

### DIFF
--- a/src/omv/imlib/imlib.h
+++ b/src/omv/imlib/imlib.h
@@ -753,6 +753,13 @@ bool image_get_mask_pixel(image_t *ptr, int x, int y);
 #define JPEG_422_YCBCR_MCU_SIZE    ((JPEG_444_GS_MCU_SIZE) * 4)
 #define JPEG_420_YCBCR_MCU_SIZE    ((JPEG_444_GS_MCU_SIZE) * 6)
 
+typedef enum jpeg_subsampling {
+    JPEG_SUBSAMPLING_AUTO = 0,
+    JPEG_SUBSAMPLING_444  = 0x11, // Chroma subsampling 4:4:4 (No subsampling)
+    JPEG_SUBSAMPLING_422  = 0x21, // Chroma subsampling 4:2:2
+    JPEG_SUBSAMPLING_420  = 0x22, // Chroma subsampling 4:2:0
+} jpeg_subsampling_t;
+
 // Old Image Macros - Will be refactor and removed. But, only after making sure through testing new macros work.
 
 // Image kernels
@@ -964,12 +971,6 @@ typedef enum template_match {
     SEARCH_DS,  // Diamond search
 } template_match_t;
 
-typedef enum  jpeg_subsample {
-    JPEG_SUBSAMPLE_1x1 = 0x11,  // 1x1 chroma subsampling (No subsampling)
-    JPEG_SUBSAMPLE_2x1 = 0x21,  // 2x2 chroma subsampling
-    JPEG_SUBSAMPLE_2x2 = 0x22,  // 2x2 chroma subsampling
-} jpeg_subsample_t;
-
 typedef enum corner_detector_type {
     CORNER_FAST,
     CORNER_AGAST
@@ -1180,7 +1181,7 @@ void imlib_hardware_jpeg_deinit();
 void jpeg_get_mcu(image_t *src, int x_offset, int y_offset, int dx, int dy,
                   int8_t *Y0, int8_t *CB, int8_t *CR);
 void jpeg_decompress(image_t *dst, image_t *src);
-bool jpeg_compress(image_t *src, image_t *dst, int quality, bool realloc);
+bool jpeg_compress(image_t *src, image_t *dst, int quality, bool realloc, jpeg_subsampling_t subsampling);
 bool jpeg_is_valid(image_t *img);
 int jpeg_clean_trailing_bytes(int bpp, uint8_t *data);
 void jpeg_read_geometry(FIL *fp, image_t *img, const char *path, jpg_read_settings_t *rs);
--- a/src/omv/imlib/jpege.c
+++ b/src/omv/imlib/jpege.c
@@ -1509,7 +1509,7 @@ static void jpeg_init(int quality) {
     }
 }
 
-static void jpeg_write_headers(jpeg_buf_t *jpeg_buf, int w, int h, int bpp, jpeg_subsample_t jpeg_subsample) {
+static void jpeg_write_headers(jpeg_buf_t *jpeg_buf, int w, int h, int bpp, jpeg_subsampling_t subsampling) {
     // Number of components (1 or 3)
     uint8_t nr_comp = (bpp == 1)? 1 : 3;
 
@@ -1574,7 +1574,7 @@ static void jpeg_write_headers(jpeg_buf_t *jpeg_buf, int w, int h, int bpp, jpeg
     jpeg_put_bytes(jpeg_buf, m_sof0, sizeof(m_sof0));
     for (int i = 0; i < nr_comp; i++) {
         // Component ID, HV sampling, q table idx
-        jpeg_put_bytes(jpeg_buf, (uint8_t [3]) {i + 1, (i == 0 && bpp == 2)? jpeg_subsample:0x11, (i > 0)}, 3);
+        jpeg_put_bytes(jpeg_buf, (uint8_t [3]) {i + 1, (i == 0 && bpp == 2)? subsampling:0x11, (i > 0)}, 3);
 
     }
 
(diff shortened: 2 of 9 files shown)


iabdalkader:
I think we should Not change sub-sampling with quality anymore, and just use the most common subsampling, known to work with VLC at minimum, by default. The reason for this is obvious, if you lower the quality it will stop working, which will be very confusing. The 1:1 produces the best quality, but the purpose of JPEG compression is Not to produce high quality images.

Note with/without this PR VLC doesn't work for me.

kwagyeman:
Hi, this PR was needed to make VLC work for me. When debugging I discovered that VLC only accepts YUV422 images.

Let me produce a test setup. I was pretty sure outputting YUV422 images was the fix for this.

I agree on fixing the quality to one standard and not dynamically adjusting it.

kwagyeman:
@iabdalkader PR Updated. It was tested on the H7 Plus and RT1060, and VLC was used on Windows and Arch Linux. Everything works as expected now. ffmpeg was tested, and it still works on Windows with the RT1060 and H7 Plus.

Just run the RTSP WLAN streaming script in the IDE, and it should work.

iabdalkader:
I understand what this PR is for, and as I've said before, I tested VLC with/without this PR RTSP streaming does Not work. And I tested it again just now, after the updates, and it's still not working.

image

kwagyeman:
Mmm... this PR appears to fix the issue for me on linux and windows. Let me ping you on slack. I thought this was the fix. It could be not.

iabdalkader:
I did some more testing and research about this, and found out some interesting stuff that explain the mystery: 

1) The reason why streaming wasn't working for me, is that I was missing the `live555` libraries on Arch (installed with `pacman -S live-media`). VLC doesn't pull this automatically when it's installed, but will dynamically loaded it if available. Once I installed this streaming started working.

2) The reason why 422 is required, is Not because VLC can't decode the other sub-sampling, it's because 422 is what's expected for type 0, which is hard-coded in the `rtsp.py` library. It's been documented in the specifications here all along (see https://datatracker.ietf.org/doc/html/rfc2435#section-4.1). I tested both type 0 422 and type 1 and 420 and both work. Based on that, 4:4:4 is just never going to work. 

```
 The two RTP/JPEG types currently defined are described below:

                            horizontal   vertical   Quantization
           types  component samp. fact. samp. fact. table number
         +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
         |       |  1 (Y)  |     2     |     1     |     0     |
         | 0, 64 |  2 (U)  |     1     |     1     |     1     |
         |       |  3 (V)  |     1     |     1     |     1     |
         +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
         |       |  1 (Y)  |     2     |     2     |     0     |
         | 1, 65 |  2 (U)  |     1     |     1     |     1     |
         |       |  3 (V)  |     1     |     1     |     1     |
         +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

3) Note that the specifications say packets shouldn't be fragmented at stack level, to allow partial delivery, but they are, because the MTU is hard-coded to 1400, but lwip is configured by default for 800.

4) The OV5640 in JPEG mode works, but RGB565 doesn't regardless of the subsampling.

How should we proceed from here ? Well after thinking about this some more, maybe it's not a good idea to lose 444 just to enable streaming. Someone may still want the highest possible JPEG quality for taking captures for example. So I think we should add an arg to specify the sub-sampling. For everything it should be auto, determined by quality as it is right now, for RTSP it should be 422. We should use the following in `imlib.h` and I map them it to ST HAL constants when HW JPEG is enabled:

```C
typedef enum  jpeg_subsampling {
    JPEG_SUBSAMPLING_AUTO = 0,
    JPEG_SUBSAMPLING_444 = 1,  // Chroma Subsampling 4:4:4 (No subsampling)
    JPEG_SUBSAMPLING_422 = 2,  // Chroma Subsampling 4:2:2 // type 0
    JPEG_SUBSAMPLING_420 = 3,,  // Chroma Subsampling 4:2:0 // type 1
} jpeg_subsampling_t;
```

kwagyeman:
Thank you for the fantastic work and for debugging this.

Regarding the arg, do you want to add that to all jpeg compression methods or add a special setting in the omv module to force this? Using the omv module to force the setting will result in less code changes all over the place. We already have to do this to disable the frame buffer when doing RTSP streaming.

Please let me know which path you'd like.

Thanks!

iabdalkader:
> Regarding the arg, do you want to add that to all jpeg compression methods or add a special setting in the omv module to force this?

Let's add it to `jpeg_compress` both the software and hardware one, this is why we need the common enum, and then as a keyword arg in `py_image_to` with a default value of auto. This should eventually let you override from the rtsp.py library from Python, something like:
```
.to_jpeg(quality=quality, subsampling=image.JPEG_SUBSAMPLING_422)
```

kwagyeman:
Will do.

kwagyeman:
Okay, updated and I tested the changes on the Nicla and RT1062. I think I got everything you wanted. But, let me know what else to change around.

iabdalkader:
Thanks, I'll try to review and merge it later today.

> I think I got everything you wanted except mapping to HW JPEG constants. But, let me know what else to change around.

But you did already I think. I meant that we should use common constants for SW/HW encoder, and it seems to be done.

src/omv/imlib/jpege.c
```diff
@@ -1645,10 +1645,16 @@ bool jpeg_compress(image_t *src, image_t *dst, int quality, bool realloc) {
...
+        if (subsampling == JPEG_SUBSAMPLING_AUTO) {
+            if (quality <= 35) {
+                jpeg_subsample = JPEG_SUBSAMPLE_2x2;
+            } else if (quality < 60) {
+                jpeg_subsample = JPEG_SUBSAMPLE_2x1;
+            }
+        } else if (subsampling == JPEG_SUBSAMPLING_422) {
             jpeg_subsample = JPEG_SUBSAMPLE_2x1;
+        } else if (subsampling == JPEG_SUBSAMPLING_420) {
+            jpeg_subsample = JPEG_SUBSAMPLE_2x2;
```
iabdalkader, line 1657:
```suggestion
        if (subsampling == JPEG_SUBSAMPLING_AUTO) {
            if (quality <= 35) {
                subsampling = JPEG_SUBSAMPLING_420;
            } else if (quality < 60) {
                subsampling = JPEG_SUBSAMPLING_422;
            }
        }
```

We won't need the old constants anymore, or the `jpeg_subsample` variable.

src/omv/imlib/imlib.h
```diff
@@ -753,6 +753,13 @@ bool image_get_mask_pixel(image_t *ptr, int x, int y);
 #define JPEG_422_YCBCR_MCU_SIZE    ((JPEG_444_GS_MCU_SIZE) * 4)
 #define JPEG_420_YCBCR_MCU_SIZE    ((JPEG_444_GS_MCU_SIZE) * 6)
 
+typedef enum jpeg_subsampling {
+    JPEG_SUBSAMPLING_AUTO = 0,
+    JPEG_SUBSAMPLING_444  = 1, // Chroma subsampling 4:4:4 (No subsampling)
+    JPEG_SUBSAMPLING_422  = 2, // Chroma subsampling 4:2:2
+    JPEG_SUBSAMPLING_420  = 3, // Chroma subsampling 4:2:0
```
iabdalkader, line 760: We should remove the old `jpeg_subsample_t`. Note those HV sampling values are written to SOF0, so we can either set the value here or fix in `jpeg_write_headers`.

```suggestion
    JPEG_SUBSAMPLING_444  = 0x11, // Chroma subsampling 4:4:4 (No subsampling)
    JPEG_SUBSAMPLING_422  = 0x21, // Chroma subsampling 4:2:2
    JPEG_SUBSAMPLING_420  = 0x22, // Chroma subsampling 4:2:0
```

iabdalkader (changes requested):

src/omv/imlib/imlib.h
kwagyeman: Done

src/omv/imlib/jpege.c
kwagyeman: Done

kwagyeman:
@iabdalkader PR updated.

iabdalkader (approved):

PR #2158, opened 2024-02-19
imlib/fmath: Fix fmath function inlining.

kwagyeman:
Compilier wasn't inlining them. Not sure if this is the right fix. Feel free to close the PR and create the proper fix.


iabdalkader:
> Feel free to close the PR and create the proper fix.

No this is fine, but the functions in the header should be `static inline`

iabdalkader:
merged in https://github.com/openmv/openmv/pull/2174

Issue #2064, opened 2023-12-29
Cleanup/optimize erode/dilate


kwagyeman:
General todo.

Will tackle after "modules/py_image: Optimize and cleanup all math and binary line ops." is merged.

PR #2160, opened 2024-02-24
imlib/binary: Improve erode/dilate performance using SIMD.

kwagyeman:
This PR provides huge performance boost to erode and dilate using cortex-m SIMD:

At the VGA resolution for erode/dilate:

---GRAYSCALE---
kernel size 0 (1x1) -> old 31ms -> new 12ms (158% speedup)
kernel size 1 (3x3) -> old 47ms -> new 18ms (161% speedup)
kernel size 2 (5x5) -> old 62ms -> new 25ms (148% speedup)
kernel size 3 (7x7) -> old 79ms -> new 34ms (132% speedup)
kernel size 4 (9x9) -> old 101ms -> new 48ms (110% speedup)
kernel size 5 (11x11) -> old 126ms -> new 65ms (93% speedup)

---RGB565---
kernel size 0 (1x1) -> old 33ms -> new 24ms (37% speedup)
kernel size 1 (3x3) -> old 47ms -> new 35ms (34% speedup)
kernel size 2 (5x5) -> old 62ms -> new 50ms (24% speedup)
kernel size 3 (7x7) -> old 78ms -> new 64ms (21% speedup)
kernel size 4 (9x9) -> old 101ms -> new 88ms (14% speedup)
kernel size 5 (11x11) -> old 130ms -> new 114ms (14% speedup)

Fixes: https://github.com/openmv/openmv/issues/2064

Since no kernel is applied per element, you can use the sliding window approach with erode/dilate. This approach then lends itself nicely to SIMD acceleration since you repeatedly do the same operation on each column. So, I can do 4x the work with grayscale and 2x with RGB565. Finally, this approach illuminates how to speed the code up with helium for an additional 8x and 4x performance boost.

### DIFF
--- a/src/omv/imlib/binary.c
+++ b/src/omv/imlib/binary.c
@@ -939,18 +939,19 @@ static void imlib_erode_dilate(image_t *img, int ksize, int threshold, int e_or_
         case PIXFORMAT_BINARY: {
             buf.data = fb_alloc(IMAGE_BINARY_LINE_LEN_BYTES(img) * brows, FB_ALLOC_NO_HINT);
 
-            for (int y = 0, yy = img->h; y < yy; y++) {
+            for (int y = 0; y < img->h; y++) {
                 uint32_t *row_ptr = IMAGE_COMPUTE_BINARY_PIXEL_ROW_PTR(img, y);
                 uint32_t *buf_row_ptr = IMAGE_COMPUTE_BINARY_PIXEL_ROW_PTR(&buf, (y % brows));
                 int acc = 0;
 
-                for (int x = 0, xx = img->w; x < xx; x++) {
+                for (int x = 0; x < img->w; x++) {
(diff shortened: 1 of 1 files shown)


src/omv/imlib/binary.c
```diff
@@ -939,18 +939,19 @@ static void imlib_erode_dilate(image_t *img, int ksize, int threshold, int e_or_
         case PIXFORMAT_BINARY: {
             buf.data = fb_alloc(IMAGE_BINARY_LINE_LEN_BYTES(img) * brows, FB_ALLOC_NO_HINT);
 
-            for (int y = 0, yy = img->h; y < yy; y++) {
```
iabdalkader, line 942: Thanks for removing those.

kwagyeman: Yup, I'm learn what you want.

PR #2188, opened 2024-04-01
imlib/truetype.c add truetype font support

xue-singtown:
Hi, I add truetype font feature with stb_truetype library.

compare_nottf2
compare_ttf2

1. draw_string will use builtin ttf font, it is a 12.6KB subset of Roboto-Regular.ttf.

generate ttf.h step:
```
pyftsubset Roboto-Regular.ttf --text="\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_\`abcdefghijklmnopqrstuvwxyz{|}~!" --no-hinting
python ttf2c.py --font-file-path=Roboto-Regular.subset.ttf > ttf.h
```

2. I also add a extra font_path param. We can load the subset .ttf file we need, and then draw text on OpenMV, whether it's Chinese, Japanese, Arabic, etc.

char_rotation_correct


iabdalkader:
I do like how the drawing looks, and support for different languages, but this library is a bit of an overkill just to draw a string, and it seems to be very heavy on dynamic memory allocations. I'm not sure if we should even use it at all, let alone make it the only string drawing option. Maybe if we can keep the optimized, simple string drawing function as the default, and branch to TTF with an argument (on boards that have it enabled via an option, ex `IMLIB_ENABLE_TTF`). Thoughts ? @kwagyeman 

Also note that the font-file loading should be done once in Python, not on every call, and passed to the draw_string, this way a font can be loaded once and used many times in a loop, for example:
```Python
fontbuf = open("font", "r").read()
....
while True:
    draw_string(...., font=fontbuf)
```

kwagyeman:
I'll need to go through the code and move all the mallocs to a private heap while drawing is happening to avoid fragmenting the micropython heap. This is doable.

Definitely need font loading to happen, not per call.

As for supporting the old font, I think we should do some benchmarks and see if it matters. Most likely it will. But, I'll be able to make an informed call after knowing the speed difference. Moving everything to ttf font though will just look so much better.

@xue-singtown Awesome work, though, getting it this far.

xue-singtown:
The draw string speed between truetype font  and the previous font as follow:

| Draw string loop | 1 | 10 |  100 |
| ------- | ------- | ------- | ------- |
| TrueType Font  | 6  | 55  | 536 |
| Original Font  | 1  | 7 | 70 |

We use "ABCD\nXXXX\n***.." as the test string to draw.

And, the firmware.bin has increased by 24 KB (Font file occupies 12.6KB) in size.

iabdalkader:
> As for supporting the old font, I think we should do some benchmarks and see if it matters.

It's not just about performance, boards with overflowing flashes can't enable the built-in 12KBs font, so they'll be forced to load a font just to draw a string. The old draw_string must be kept, then it can branch to TTF if:
`TTF is enabled && (built-in font is enabled || a font is loaded)`, otherwise it falls back to the old draw_string, there's no need to even raise errors.

kwagyeman:
@xue-singtown - Now that ROMFS is merged we have a way to get this working thanks to the extra flash onboard different systems. I like https://github.com/tomolt/libschrift/tree/master more than stb given it's smaller size and less mallocs.

Also, I looked at your font subsetting process and that can now be easily integrated into OpenMV IDE given the work I put in to support the latest release. 

image

It's realistic to add an option in the IDE to add a font to ROMFS and handle the filtering processing when adding the font. The sizes of the font are in range to be installable on ROMFS without issues.

Anyway, I will look into this in a few months. We definitely need to upgrade it before shipping the AE3/N6 because the current font is rough.

...

Notes on my thoughts for implementation:

1. IDE will use python tools in a cross platform way to convert fonts for storage on ROMFS. I can re-use a lot of the GUI work I did for other tools to get this done within a day to make it easy to reduce fonts from raw .ttf files and generate small versions.

2. Add new Font() class under img module. This can be passed the romfs ttf file. It can then just memory map the ttf file. I can also make it smart enough to handle loading from disk too like how Model() works.

3. draw_string() will just render the text drawn into a 8bpp alpha buffer. This buffer then will be drawn on the target image using draw_image. This allows the GPU to be invoked to speed up drawing large text blocks and has better alpha blending. This will enhance performance for all OpenMV Cams (e.g. DMA2D on older systems). Additionally, color can be applied here, as well as additional font scaling, etc. This also allows a lot of the rotation/flipping/etc. args to be handled by draw image.

4. A get_text_rect() function or something like that will be added so that you can know the text size and baseline. Similar to what OpenCV provides.

iabdalkader:

https://forums.openmv.io/t/different-font-for-draw-string/10242/8

PR #2189, opened 2024-04-02
modules/py_image: Update all lineops to use the draw_image backend.

kwagyeman:
Brokeout from https://github.com/openmv/openmv/pull/2061.

This PR switches all lineops over to use draw_image as a backend. This allows all lineops to now work between image types and removes the strict geometry checks. Want to add a scaled and rotated by 90 degrees color jpeg image to a grayscale image? Sure. This works now. You can pretty much do anything with compositing images now. Also, about 5KB code space is saved.

...

Fixes: https://github.com/openmv/openmv/issues/1855

Issue #1449, opened 2021-09-08
Move some types out of py_image.c


iabdalkader:
To help reduce the `py_image.c` file size, please move `py_histogram_type`, `py_threshold_type`, `py_percentile_type`, `py_statistics_type` and  `py_similarity_type`  to `py_image_stats.c` ? just like `py_image.c` and `py_imageio.c`

kwagyeman:
Working on this today.

PR #2190, opened 2024-04-03
modules/py_image: Refactor get_similarity() to use draw_image backend.

kwagyeman:
Brokeout from #2061.

GetSimilarity used the lineops backend. This PR switches it over to use the draw image backend so that the one lineop backend can be deleted.

...

"imlib_image_operation" can be deleted in another PR as #2189 has to be merged first before I can remove it.


iabdalkader:
After these cleanups, can we consider https://github.com/openmv/openmv/issues/1449 ?

kwagyeman:
@iabdalkader Yes now that the conference is over.

iabdalkader (approved):

Issue #2063, opened 2023-12-29
Cleanup and .to_x method and and image rotation and transpose support


kwagyeman:
On merge of https://github.com/openmv/openmv/pull/2061 this needs to be done ASAP as there will be no way to vflip/hmirror/transpose and image until this is added back. Make sure to update examples.

PR #2191, opened 2024-04-04
modules/py_image: Update py_image_to with mp_arg_parse_all.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2063

This PR should be reviewed and merged before: https://github.com/openmv/openmv/pull/2189 To avoid loosing support for vflip/hmirror/transpose in place on an image.


iabdalkader:
- `quality` should be a keyword arg.
- The docs need to be updated to add all the args to the `compress_*` functions.
- Please remove duplicate code, like so:
```C
    if ((dst_img.is_bayer || dst_img.is_yuv) &&
        ((x_scale != 1.0f) && (x_scale != -1.0f)) ||
        ((y_scale != 1.0f) && (y_scale != -1.0f)) ||
        (args[ARG_channel].u_int != -1) ||
        (args[ARG_alpha].u_int != 256) ||
        (color_palette != NULL) || (alpha_palette != NULL)) {
        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Only copying/cropping is supported for Bayer/YUV!"));
    }
```
- Do we need `compress_for_ide`,  `compressed` and `compressed_for_ide` ? Seems we can just replace those with args.
- I think we should split `py_image_to` into at least 2 functions: `py_image_convert` (called by all the functions that pass 4 false bools) and `py_image_compress` (that will replace all other compress functions). Yes, some code will be repeated, but it might be better than `py_image_to` accepting args that determine which args it accepts. Let's see how this looks in a separate commit and then decide.

kwagyeman:
> quality should be a keyword arg.

Sure, this will break the API. However, given there are no other positional args I guess it will be easy for users to fix. This will cleanup things. I'll update our scripts per the changed API.

> The docs need to be updated to add all the args to the compress_* functions.

Of course. But, that will be in a doc update PR.

> Please remove duplicate code, like so:

Will do.

> Do we need compress_for_ide, compressed and compressed_for_ide ? Seems we can just replace those with args.

Yeah... this is old API maintenance. We can drop them all and switch to to_jpeg(). Would be better long term but will break the API again.

> I think we should split py_image_to into at least 2 functions: py_image_convert (called by all the functions that pass 4 false bools) and py_image_compress (that will replace all other compress functions). Yes, some code will be repeated, but it might be better than py_image_to accepting args that determine which args it accepts. Let's see how this looks in a separate commit and then decide.

Sure, will do a commit after addressing the comments above ontop of this commit to split it. It's going to be 100% code duplication though more or less.

iabdalkader:
> Yeah... this is old API maintenance. We can drop them all and switch to to_jpeg(). Would be better long term but will break the API again.

I think we should, they're not that popular, especially encode_for_ide, but we can also just alias `to_jpeg` as `compress` for now.

kwagyeman:
Yeah, I don't think it will be too much of a ruckus to drop them.

kwagyeman:
Back on this now.

kwagyeman:
Made all the changes suggested except this one:

> I think we should split py_image_to into at least 2 functions: py_image_convert (called by all the functions that pass 4 false bools) and py_image_compress (that will replace all other compress functions). Yes, some code will be repeated, but it might be better than py_image_to accepting args that determine which args it accepts. Let's see how this looks in a separate commit and then decide.

Not sure if it applies anymore since the compress methods are all gone given the other changes. The only thing that has variable arguments is the copy/copy_to_fb. I could just make the function accept both arguments and then handle the user passing either. This will then eliminate py_image_to accepting args that determine which args it accepts.

iabdalkader:
> Not sure if it applies anymore since the compress methods are all gone given the other changes

Probably not, I'll review it again. But I think you've pushed some unrelated changes in this PR, to draw and some scripts changes that have to do with drawing hints.

iabdalkader (changes requested):

src/omv/modules/py_image.c
```diff
@@ -1106,145 +1066,126 @@ static mp_obj_t py_image_to(pixformat_t pixfmt, const uint16_t *default_color_pa
...
-        PY_ASSERT_TRUE_MSG((size <= buf_size),
-                           "The new image won't fit in the target frame buffer!");
-        // DO NOT MODIFY arg_other YET (as it could point to src_img)!
-        dst_img.data = arg_other->data;
-        // Update now if not in place to possibly freeup framebuffer RAM, otherwise update after.
-        if (dst_img.data != src_img->data) {
-            py_helper_update_framebuffer(&dst_img);
-        }
+    } else if (!copy_to_fb) {
+        bool fb = py_helper_is_equal_to_framebuffer(src_img);
```
iabdalkader, line 1113: I noticed that this function might not be safe to use anymore. The FB head can change before/after this function is called. It worked fine for a while, way before the multiple framebuffers were introduced but probably not anymore.

iabdalkader:
Question, why can't we just add the two args `copy` and `copy_to_fb` ? Then do something like this:

```C
        // Convert to FB.
        py_helper_set_to_framebuffer(&dst_img);
    } else if (args[ARG_copy].u_bool) {
        // Create dynamic copy.
        dst_img.data = xalloc(image_size(&dst_img));
    } else {
        // Convert in place.
        bool fb = py_helper_is_equal_to_framebuffer(src_img);
        size_t buf_size = fb ? framebuffer_get_buffer_size() : image_size(src_img);
        PY_ASSERT_TRUE_MSG((image_size(&dst_img) <= buf_size),
                "The converted image will not fit into the taget frame buffer");
        dst_img.data = src_img->data;
    }
....
....
        py_helper_update_framebuffer(&dst_img);
        framebuffer_update_jpeg_buffer();
    }

    return py_image_from_struct(&dst_img);

```

kwagyeman:
Yeah, I think adding two args is the way to go. Will make that change.

kwagyeman:
Regarding the draw stuff. Without that fix transposing using to_x() doesn't work.

kwagyeman:
Made the changes requested.

scripts/examples/08-RPC-Library/34-Remote-Control/image_transfer_jpg_as_the_remote_device_for_your_computer.py
```diff
@@ -68,7 +68,7 @@ def jpeg_image_snapshot(data):
     pixformat, framesize = bytes(data).decode().split(",")
     sensor.set_pixformat(eval(pixformat))
     sensor.set_framesize(eval(framesize))
-    img = sensor.snapshot().compress(quality=90)
+    img = sensor.snapshot().to_jpeg(quality=90)
```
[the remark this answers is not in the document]
kwagyeman: Done.

src/omv/modules/py_image.c
kwagyeman: The head would change after setting the frame buffers back to 1 from 3. If you force 1 frame buffer first is should be the same.

iabdalkader (approved):

PR #2194, opened 2024-04-14
imlib/draw: Optimize set_pixel_aa.

kwagyeman:
Improved the blending math used to be more accurate. The previous version does the divide too early, which can result in artifacts.

Removed SIMD code from this. It's not going to make much of a difference here, as there's no loop.


iabdalkader:
Merge ?

iabdalkader (approved):

kwagyeman:
Yes

PR #2198, opened 2024-04-15
Add files via upload

Sander-01:


Sander-01:
I'm not sure I'm doing this right, I just want to build a firmware using the documents that I've updated. The firmware is to use with Open MV with an arduino nicla vision. I don't know what am I doing wrong but I cannot create the firmware. Thank you

kwagyeman:
Hi, you need to fork your own repo. After you do that just enable github actions on your own repo. Then any commits you make to that repo will build the firmware.

Do not send PRs to the main repo.

PR #2201, opened 2024-04-17
imlib/draw: Fix out of bound crash issue with set_pixel_aa.

kwagyeman:
The previous code did not bounds check before reading. This can cause the system to crash.

### DIFF
--- a/src/omv/imlib/draw.c
+++ b/src/omv/imlib/draw.c
@@ -89,21 +89,28 @@ static void point_fill(image_t *img, int cx, int cy, int r0, int r1, int c) {
 }
 
 static void imlib_set_pixel_aa(image_t *img, int x, int y, int err, int c) {
-    // float alpha = 1 - err / 256.0;
-    void *row_ptr = imlib_compute_row_ptr(img, y);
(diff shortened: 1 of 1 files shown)


src/omv/imlib/draw.c
```diff
@@ -89,42 +89,46 @@ static void point_fill(image_t *img, int cx, int cy, int r0, int r1, int c) {
 }
 
 static void imlib_set_pixel_aa(image_t *img, int x, int y, int err, int c) {
-    // float alpha = 1 - err / 256.0;
```
iabdalkader, line 92: Can we instead check and return ?

```C
if (!((0 <= x) && (x < img->w) && (0 <= y) && (y < img->h))) {
    // Out of bounds.
    return;
}

kwagyeman: Sure, yeah. That's faster.

kwagyeman: Updated.

kwagyeman: Note, I removed imlib_set_pixel() because that would bounds check again.

iabdalkader (approved):

PR #2206, opened 2024-04-22
examples/ServoShield: Change to SoftI2c for RT.

kidswong999:
Use SoftI2C to support OpenMV RT.


iabdalkader:
This should also work on the RT1062.

kwagyeman:
@kidswong999 - Any updates on this PR?

iabdalkader:
Can you provide some more details about why this is needed ?

kwagyeman:
@iabdalkader - Verified this pull request fixes an issue with I2C. I2C() requires an ID to be passed.

`I2C(1)`

Works without the keywords. But, will not necessarily match on the STM32 boards.

PR #2207, opened 2024-04-29
misc: Increase delay after reset/power-down.

iabdalkader:
The Portenta sometimes fails to detect HM01B0 a soft-reset fixes it. I think this happens because reset is connected to an LDO, not to the sensor directly, so maybe it needs more time to settle down after it's enabled. Increasing the delay after reset/power-down seemed to fix it for me, but then I couldn't get it to fail again without this change, so I'm not sure if it's really needed.


iabdalkader:

https://forums.openmv.io/t/portenta-h7-portenta-vision-shield-ethernet-runtimeerror-sensor-control-failed-no-hw-issue/9522/8

iabdalkader:
> @iabdalkader - Rather not globally increase this time. Maybe just have it increased for the Portenta only?

Yes we can do that, although this only runs once on boot.

iabdalkader:
@kwagyeman updated

iabdalkader:
I want to include this fix in the release. This will be the last release before updating to mp 1.23.

kwagyeman (commented):
Looks good.

PR #2210, opened 2024-05-11
imlib/pool: Remove obsolete pooling functions.

kwagyeman:
Midpoint pooling was never actually used. Mean pooling has been replaced by area scaling built into draw image... and thus available everywhere. AREA scaling in draw_image also handles fractional scaling and not just integer scaling.


iabdalkader:
This needs to be rebased.

kwagyeman:
Done

iabdalkader (approved):

PR #2211, opened 2024-05-12
Kwabena/cleanup and optimize mean

kwagyeman:
Mean now uses mp_arg_parse_all and has been optimized using SIMD:

On the H7 Plus - Mean(3) - (e.g. a 7x7 kernel)

QQVGA Grayscale Before: 5.48ms, After: 3.87ms - 42% speedup
QVGA Grayscale Before: 18.03ms, After: 11.38ms - 58% speedup
VGA Grayscale Before: 64.23ms, After: 37.14ms - 73% speedup
SVGA Grayscale Before: 99.20ms, After: 56.27ms - 76% speedup
XGA Grayscale Before: 157.93ms, After: 88.20ms - 79% speedup

QQVGA RGB565 Before: 7.77ms, After: 6.81ms - 14% speedup
QVGA RGB565 Before: 25.74ms, After: 21.97ms - 17% speedup
VGA RGB565 Before: 96.44ms, After: 80.47ms - 19% speedup
SVGA RGB565 Before: 150.40ms, After: 128.10ms - 17% speedup
XGA RGB565 Before: 264.70ms, After: 230.10ms - 15% speedup

Things are setup for a 4x speedup with Helium later.


iabdalkader:
Before/after this PR, mean produces images that are very different, not visually but numerically. I don't think this is right, they should be the same. Here's an example:
```diff
24c24
< 00000170: 24ab 44ab 65b3 85b3 85b3 65b3 65b3 45ab  $.D.e.....e.e.E.
---
> 00000170: 24ab 44ab 65b3 65b3 85b3 65b3 65b3 45ab  $.D.e.e...e.e.E.
```

kwagyeman:
@iabdalkader - This PR is ready for review. See the link to the spreadsheet in the first comment for the board comparison numbers.

Regarding the images being slightly different in values. This was because I changed the way the divider was calculated. I found a way to keep it the same as before for all non-helium boards. For helium boards, it will have 1-bit less precision.

src/omv/Makefile
```diff
@@ -110,7 +110,8 @@ SRCS += $(wildcard ports/$(PORT)/*.c)
 
 OBJS = $(addprefix $(BUILD)/, $(SRCS:.c=.o))
 ifeq ($(USE_CLANG),1)
-CLANG_OBJS = $(BUILD)/imlib/bayer.o
+CLANG_OBJS = $(BUILD)/imlib/bayer.o \
+			 $(BUILD)/imlib/filter.o
```
kwagyeman, line 114: This is aligned with tabs. Let me know if it should be spaces.

PR #2214, opened 2024-05-13
modules/py_image: Update morph() to use arg parse.

kwagyeman:
morph()/gaussian()/laplacian() updated to use mp_arg_parse.

### DIFF
--- a/src/omv/modules/py_image.c
+++ b/src/omv/modules/py_image.c
@@ -2424,161 +2424,213 @@ static MP_DEFINE_CONST_FUN_OBJ_KW(py_image_midpoint_obj, 2, py_image_midpoint);
 #endif // IMLIB_ENABLE_MIDPOINT
 
 #ifdef IMLIB_ENABLE_MORPH
-static mp_obj_t py_image_morph(uint n_args, const mp_obj_t *args, mp_map_t *kw_args) {
-    image_t *arg_img =
-        py_helper_arg_to_image(args[0], ARG_IMAGE_MUTABLE);
-    int arg_ksize =
-        py_helper_arg_to_ksize(args[1]);
+static mp_obj_t py_image_morph(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    enum { ARG_mul, ARG_add, ARG_threshold, ARG_offset, ARG_invert, ARG_mask };
+    static const mp_arg_t allowed_args[] = {
+        { MP_QSTR_mul, MP_ARG_OBJ | MP_ARG_KW_ONLY,  {.u_rom_obj = MP_ROM_NONE } },
+        { MP_QSTR_add, MP_ARG_OBJ | MP_ARG_KW_ONLY,  {.u_rom_obj = MP_ROM_NONE } },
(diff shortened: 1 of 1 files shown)


src/omv/modules/py_image.c
```diff
@@ -2345,60 +2345,102 @@ static MP_DEFINE_CONST_FUN_OBJ_KW(py_image_midpoint_obj, 2, py_image_midpoint);
...
+        { MP_QSTR_invert, MP_ARG_BOOL | MP_ARG_KW_ONLY,  {.u_bool = false } },
+        { MP_QSTR_mask, MP_ARG_OBJ | MP_ARG_KW_ONLY,  {.u_rom_obj = MP_ROM_NONE} },
+    };
+
+    // Parse args.
+    image_t *image = py_helper_arg_to_image(pos_args[0], ARG_IMAGE_MUTABLE);
+    int ksize = py_helper_arg_to_ksize(pos_args[1]);
+    int k_2 = ksize * 2;
+    int n = k_2 + 1;
+    int n2 = n * n;
```
iabdalkader, line 2364:
```suggestion
```
Please remove these, and in the future please avoid doing that.

iabdalkader (changes requested):

src/omv/modules/py_image.c
kwagyeman: Removed k_2 and n2.

src/omv/modules/py_image.c
```diff
@@ -2424,161 +2424,201 @@ static MP_DEFINE_CONST_FUN_OBJ_KW(py_image_midpoint_obj, 2, py_image_midpoint);
...
+    mp_obj_get_array(pos_args[2], &len, &items);
+
+    mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
+    mp_arg_parse_all(n_args - 3, pos_args + 3, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
 
     fb_alloc_mark();
 
-    int *arg_krn = fb_alloc(n * sizeof(int), FB_ALLOC_NO_HINT);
-    int arg_m = 0;
+    int *krn = fb_alloc((n * n) * sizeof(int), FB_ALLOC_NO_HINT);
```
iabdalkader, line 2452: If there's an upper limit for `n`, something sane, you could just do this, and save memory and use faster RAM:

```suggestion
    int krn[n][n];
```

iabdalkader (approved):

src/omv/modules/py_image.c
kwagyeman: Good catch. Not really sure why that's fballoced... looks like it was very old code I just re-factored.

kwagyeman: Updated

iabdalkader: Now that it's stack based, should there be a sanity check for the max size ? For example a 100  ? Same for pascal. Also, are the mark/free needed ?

kwagyeman: Mark and free are needed for the actual function. I can add a sanity check on the size.

kwagyeman: Added a limit to a ksize of 15 (n==31).

iabdalkader (approved):

PR #2215, opened 2024-05-13
modules/py_image: Move ksize to be pos_arg.

kwagyeman:
py_helper_arg_to_ksize() checks that ksize is greater than or equal to 0. So, using it again.


iabdalkader:
> py_helper_arg_to_ksize() checks that ksize is greater than or equal to 0. So, using it again.

More context ?

kwagyeman:
py_helper_arg_to_ksize() was used in the original code to protect you from passing a kernel size below 0. So, it needed to be added back.

iabdalkader:
> py_helper_arg_to_ksize() was used in the original code to protect you from passing a kernel size below 0. So, it needed to be added back.

Okay, but why switch it from keyword arg to positional ? If it's missing it will now just say missing a positional argument vs the arg name. Either way is fine with me, just pointing out that subtle change.

kwagyeman:
It just had to do with the fact that py_helper_arg_to_ksize() takes a micropython object. You can't pass it an int. I guess I could have created a new method that accepts an int. Was just trying to re-use what we already have.

iabdalkader (approved):

PR #2221, opened 2024-05-21
ports/stm32: Reset vbuffer state when dropping frames.

iabdalkader:
When dropping a frame the current vbuffer state was Not reset, which caused the following frame(s) to be corrupted, probably also overwriting the next vbuffer.
This bug is what caused transposed high-resolution frames to be corrupted. If the DMA is not keeping up you should simply just see slower FPS or no frames at all but definitely never corrupted frames, since they're dropped.

Test code now never corrupts a frame:
```Python
import sensor
import time
import image

sensor.reset()
sensor.set_transpose(True)

clock = time.clock()
while True:
    clock.tick()
    sensor.snapshot().to_rainbow(color_palette=image.PALETTE_IRONBOW)
    print(clock.fps())
```

### DIFF
--- a/src/omv/ports/stm32/sensor.c
+++ b/src/omv/ports/stm32/sensor.c
@@ -405,6 +405,12 @@ void HAL_DCMI_FrameEventCallback(DCMI_HandleTypeDef *hdcmi) {
     sensor.first_line = false;
     if (sensor.drop_frame) {
         sensor.drop_frame = false;
+        // If the frame was dropped, the buffer will not change, so its state
+        // must be reset.
+        vbuffer_t *buffer = framebuffer_get_tail(FB_PEEK);
+        if (buffer) {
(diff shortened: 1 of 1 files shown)


kwagyeman (approved):
Please add the same code to the imxrt too and any other sensor drivers. Or move into the framebuffer code.

iabdalkader:
> Please add the same code to the imxrt too and any other sensor drivers. Or move into the framebuffer code.

I'm looking at imxrt, don't think it has the same issue. As for moving it to framebuffer, I don't think it's possible because only the sensor code knows when the frame is dropped, then it gets reused again for the next frame.

iabdalkader:
IMXRT doesn't have that issue, so the other PRs are needed.

kwagyeman:
Cool

Issue #2218, opened 2024-05-18
Enable sys.settrace in firmware.


kidswong999:
With sys.settrace, we can use the single-step debugging function on the board.
This is very useful for debugging.

We just nee add `MICROPY_PY_SYS_SETTRACE` in MicroPython options.

iabdalkader:
> We just nee add `MICROPY_PY_SYS_SETTRACE` in MicroPython options.

I've looked into this, and the docs says it impacts the performance, but more importantly it adds about 34 KBs to the firmware, so I'm not sure if we should enable it.

kidswong999:
Because settrace requires additional performance and locals() cannot be implemented, we can turn it off and wait until micropython improves this feature in the future.

PR #2224, opened 2024-05-28
micropython: enable sys.settrace.

xue-singtown:
#2218 
I enabled OPENMV4, OPENMV4P, OPENMV_RT1060


iabdalkader:
Not sure about this, see my comment here: https://github.com/openmv/openmv/issues/2218#issuecomment-2131106002

PR #2134, opened 2024-02-08
modules/py_tf: Add support for yolov3, yolov5 and yolov7.

kwagyeman:
Adds support for YOLOV3, V5, V7 to detect().

Depends on https://github.com/openmv/openmv/pull/2133.

image (18)

PR #2227, opened 2024-06-03
modules/py_tf: Add generic CNN processing support.

kwagyeman:
Replaces: https://github.com/openmv/openmv/pull/2134

...

This PR adds support for a new method called invoke() which takes and input image in, runs tensorflow, and outputs a a 1d arrary of the results of on the micropython heap organized in h/w/c order. This is the default behavior, though its not very efficent when used with large models.

Invoke() can additionally take in a callback method though that can access the model output while it's still in the frame buffer stack for much more efficent operation. The callback is passed a new "model_output" object which has a subscr [] method which can get any value from the output sensor that's organized in h/w/c order. You can get individual values or slices. The "model_output" class also provides all the necessary dimension information for you to write whatever post processing you need in python. Whatever this callback returns then will be the result of invoke().

To make post-processing easier the invoke_output class contains "add_bounding_box" and "get_bounding_boxes" methods which encapulsate common operations for object detector networks. This allows the python post processing code to just need to be written for reading the output of a network. The python code does not need to reimplement non-max impression which is handled by the bounding box code.

Finally, control over input scaling was added to support models trained on different input scales. The defaults match the previous input scaling behavior. Note, input scaling is only needed for models which have floating point input heads.

...

Here's how the API would look for processing a YOLOV5 Model:

```Python
import time
import tf
import image

model = 'yolo/lpd-yolov5-int8-quantized.tflite'

# Alternatively, models can be loaded from the filesystem storage.
net = tf.load(model, load_to_fb=True)
labels = ["plate"]
print(net)

colors = [  # Add more colors if you are detecting more than 7 types of classes at once.
    (255, 0, 0),
    (0, 255, 0),
    (255, 255, 0),
    (0, 0, 255),
    (255, 0, 255),
    (0, 255, 255),
    (255, 255, 255),
]

@micropython.native
def yolov5(model, o):
    oh, ow, oc = model.output_shape
    if oh!= 1:
        raise(ValueError("Expected model output height to be 1!"))
    if oc < 6:
        raise(ValueError("Expected model output channels to be >= 6"))
    # cx, cy, cw, ch, score, classes[n]
    ol = oh * ow * oc
    ih, iw, ic = model.input_shape
    m = time.ticks_ms()
    for i in range(0, ol, oc):
        score = o[i + 4]
        if (score > 0.5):
            cx = o[i + 0]
            cy = o[i + 1]
            cw = o[i + 2] * 0.5
            ch = o[i + 3] * 0.5
            xmin = (cx - cw) * iw
            ymin = (cy - ch) * ih
            xmax = (cx + cw) * iw
            ymax = (cy + ch) * ih
            labels = o[i + 5: i + oc]
            label_index = max(enumerate(labels), key=lambda x: x[1])[0]
            o.add_bounding_box(xmin, ymin, xmax, ymax, score, label_index)
    boxes = o.get_bounding_boxes()
    print("time", time.ticks_diff(time.ticks_ms(), m))
    return boxes

clock = time.clock()
while True:
    clock.tick()

    img = image.Image("yolo/plate.jpg", copy_to_fb=True)
    img.to_rgb565()

    for i, detection_list in enumerate(
        net.predict(img, callback=yolov5, scale=tf.SCALE_NONE)
    ):
        print("********** %s **********" % labels[i])
        for d in detection_list:
            print(d.rect(), "score", d.output())
            img.draw_rectangle(d.rect(), color=colors[i], thickness=2)

    print(clock.fps(), "fps", end="\n")
```

Post-processing a 6300 output head from a 320x320 input YOLOV5 model takes 7ms with this code on the H7 Plus. Comparable C code is 1ms. So, while the python is slower is it not unusable. Note that it's imperative to filter detections by the result threshold. You want to do as little work as possible in the loop since there's so much stuff to check. Also, you should add the @micropython.native to the post-processing function to make it as fast as possible, it would be 11ms otherwise without it. Finally, @micropython.viper cannot be used as the model_output class is not exposing the RAW output of the CNN but converting values on the fly as you request data into floating point form.

### DIFF
--- a/src/omv/modules/py_tf.c
+++ b/src/omv/modules/py_tf.c
@@ -215,6 +215,14 @@ STATIC mp_obj_t int_py_tf_load(mp_obj_t path_obj, bool alloc_mode, bool helper_m
         fb_alloc_mark_permanent(); // tf_model->model_data will not be popped on exception.
     }
 
+    tf_model->input_shape = mp_obj_new_tuple(3, (mp_obj_t []) {mp_obj_new_int(tf_model->params.input_height),
+                                                               mp_obj_new_int(tf_model->params.input_width),
+                                                               mp_obj_new_int(tf_model->params.input_channels)});
+
+    tf_model->output_shape = mp_obj_new_tuple(3, (mp_obj_t []) {mp_obj_new_int(tf_model->params.output_height),
+                                                                mp_obj_new_int(tf_model->params.output_width),
+                                                                mp_obj_new_int(tf_model->params.output_channels)});
+
     return tf_model;
 }
 
@@ -265,9 +273,19 @@ STATIC py_tf_model_obj_t *py_tf_load_alloc(mp_obj_t path_obj) {
     }
 }
 
+typedef enum {
+    PY_TF_SCALE_NONE,
+    PY_TF_SCALE_0_1,
+    PY_TF_SCALE_S1_1,
+    PY_TF_SCALE_S128_127
+} py_tf_scale_t;
+
 typedef struct py_tf_input_callback_data {
     image_t *img;
     rectangle_t *roi;
+    py_tf_scale_t scale;
+    float mean[3];
+    float stdev[3];
 } py_tf_input_callback_data_t;
 
 STATIC void py_tf_input_callback(void *callback_data,
@@ -275,31 +293,50 @@ STATIC void py_tf_input_callback(void *callback_data,
                                  libtf_parameters_t *params) {
     py_tf_input_callback_data_t *arg = (py_tf_input_callback_data_t *) callback_data;
 
-    // Disable checking input scaling and zero-point. Nets can be all over the place on the input
-    // scaling and zero-point but still work with the code below.
-
-    // if (params->input_datatype == LIBTF_DATATYPE_UINT8) {
-    //     if (fast_roundf(params->input_scale * GRAYSCALE_RANGE) != 1) {
-    //         mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected model input scale to be 1/255!"));
-    //     }
-
-    //     if (params->input_zero_point != 0) {
-    //         mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected model input zero point to be 0!"));
-    //     }
-    // }
-
-    // if (params->input_datatype == LIBTF_DATATYPE_INT8) {
-    //     if (fast_roundf(params->input_scale * GRAYSCALE_RANGE) != 1) {
-    //         mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected model input scale to be 1/255!"));
-    //     }
-
-    //     if (params->input_zero_point != -GRAYSCALE_MID) {
-    //         mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected model input zero point to be -128!"));
-    //     }
-    // }
-
     int shift = (params->input_datatype == LIBTF_DATATYPE_INT8) ? GRAYSCALE_MID : 0;
-    float fscale = 1.0f / GRAYSCALE_RANGE;
+    float fscale = 1.0f, fadd = 0.0f;
+
+    switch (arg->scale) {
+        case PY_TF_SCALE_0_1:
+            fscale = 1.0f / 255.0f;
+            break;
+        case PY_TF_SCALE_S1_1:
+            fscale = 2.0f / 255.0f;
+            fadd = -1.0f;
+            break;
+        case PY_TF_SCALE_S128_127:
+            fscale = 255.0f / 127.0f;
+            fadd = -128.0f;
+            break;
+        case PY_TF_SCALE_NONE:
+        default:
+            break;
+    }
+
+    float fscale_r = fscale, fadd_r = fadd;
+    float fscale_g = fscale, fadd_g = fadd;
+    float fscale_b = fscale, fadd_b = fadd;
+
+    // To normalize the input image we need to subtract the mean and divide by the standard deviation.
+    // We can do this by applying the normalization to fscale and fadd outside the loop.
+
+    // Red
+    fadd_r = (fadd_r - arg->mean[0]) / arg->stdev[0];
+    fscale_r /= arg->stdev[0];
+
+    // Green
+    fadd_g = (fadd_g - arg->mean[1]) / arg->stdev[1];
+    fscale_g /= arg->stdev[1];
+
+    // Blue
+    fadd_b = (fadd_b - arg->mean[2]) / arg->stdev[2];
+    fscale_b /= arg->stdev[2];
+
+    // Grayscale -> Y = 0.299R + 0.587G + 0.114B
+    float mean = (arg->mean[0] * 0.299f) + (arg->mean[1] * 0.587f) + (arg->mean[2] * 0.114f);
+    float std = (arg->stdev[0] * 0.299f) + (arg->stdev[1] * 0.587f) + (arg->stdev[2] * 0.114f);
+    fadd = (fadd - mean) / std;
+    fscale /= std;
 
     image_t dst_img;
     dst_img.w = params->input_width;
@@ -328,7 +365,7 @@ STATIC void py_tf_input_callback(void *callback_data,
             float *model_input_f32 = (float *) model_input;
 
             for (; size >= 0; size -= 1) {
-                model_input_f32[size] = model_input_u8[size] * fscale;
+                model_input_f32[size] = (model_input_u8[size] * fscale) + fadd;
             }
         } else {
             if (shift) {
@@ -356,9 +393,9 @@ STATIC void py_tf_input_callback(void *callback_data,
 
             for (; size >= 0; size -= 1, rgb_size -= 3) {
                 int pixel = model_input_u16[size];
-                model_input_f32[rgb_size] = COLOR_RGB565_TO_R8(pixel) * fscale;
-                model_input_f32[rgb_size + 1] = COLOR_RGB565_TO_G8(pixel) * fscale;
-                model_input_f32[rgb_size + 2] = COLOR_RGB565_TO_B8(pixel) * fscale;
+                model_input_f32[rgb_size] = (COLOR_RGB565_TO_R8(pixel) * fscale_r) + fadd_r;
+                model_input_f32[rgb_size + 1] = (COLOR_RGB565_TO_G8(pixel) * fscale_g) + fadd_g;
+                model_input_f32[rgb_size + 2] = (COLOR_RGB565_TO_B8(pixel) * fscale_b) + fadd_b;
             }
         } else {
             uint16_t *model_input_u16 = (uint16_t *) model_input;
@@ -374,34 +411,26 @@ STATIC void py_tf_input_callback(void *callback_data,
     }
 }
 
-STATIC void py_tf_classify_output_callback(void *callback_data,
-                                           void *model_output,
-                                           libtf_parameters_t *params) {
+STATIC void py_tf_output_callback(void *callback_data,
+                                  void *model_output,
+                                  libtf_parameters_t *params) {
     mp_obj_t *arg = (mp_obj_t *) callback_data;
-
-    if (params->output_height != 1) {
-        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected model output height to be 1!"));
-    }
-
-    if (params->output_width != 1) {
-        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected model output width to be 1!"));
-    }
-
-    *arg = mp_obj_new_list(params->output_channels, NULL);
+    size_t len = params->output_height * params->output_width * params->output_channels;
+    *arg = mp_obj_new_list(len, NULL);
 
     if (params->output_datatype == LIBTF_DATATYPE_FLOAT) {
-        for (int i = 0, ii = params->output_channels; i < ii; i++) {
+        for (size_t i = 0; i < len; i++) {
             ((mp_obj_list_t *) *arg)->items[i] =
                 mp_obj_new_float(((float *) model_output)[i]);
         }
     } else if (params->output_datatype == LIBTF_DATATYPE_INT8) {
-        for (int i = 0, ii = params->output_channels; i < ii; i++) {
+        for (size_t i = 0; i < len; i++) {
             ((mp_obj_list_t *) *arg)->items[i] =
                 mp_obj_new_float( ((float) (((int8_t *) model_output)[i] - params->output_zero_point)) *
                                   params->output_scale);
         }
     } else {
-        for (int i = 0, ii = params->output_channels; i < ii; i++) {
+        for (size_t i = 0; i < len; i++) {
             ((mp_obj_list_t *) *arg)->items[i] =
                 mp_obj_new_float( ((float) (((uint8_t *) model_output)[i] - params->output_zero_point)) *
                                   params->output_scale);
@@ -409,139 +438,10 @@ STATIC void py_tf_classify_output_callback(void *callback_data,
     }
 }
 
-STATIC mp_obj_t py_tf_classify(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
-    enum { ARG_roi, ARG_min_scale, ARG_scale_mul, ARG_x_overlap, ARG_y_overlap };
-    static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_roi, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
-        { MP_QSTR_min_scale, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
-        { MP_QSTR_scale_mul, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
-        { MP_QSTR_x_overlap, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
-        { MP_QSTR_y_overlap, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
-    };
-
-    // Parse args.
-    mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
-    mp_arg_parse_all(n_args - 2, pos_args + 2, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
-
-    image_t *image = py_helper_arg_to_image(pos_args[1], ARG_IMAGE_ANY);
-    rectangle_t roi = py_helper_arg_to_roi(args[ARG_roi].u_obj, image);
-    float min_scale = py_helper_arg_to_float(args[ARG_min_scale].u_obj, 1.0f);
-    float scale_mul = py_helper_arg_to_float(args[ARG_scale_mul].u_obj, 0.5f);
-    float x_overlap = py_helper_arg_to_float(args[ARG_x_overlap].u_obj, 0.0f);
-    float y_overlap = py_helper_arg_to_float(args[ARG_y_overlap].u_obj, 0.0f);
-
-    // Sanity checks
-    if ((min_scale <= 0.0f) || (min_scale > 1.0f)) {
-        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("0 < min_scale <= 1"));
-    }
-
-    if ((scale_mul < 0.0f) || (scale_mul >= 1.0f)) {
-        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("0 <= scale_mul < 1"));
-    }
-
-    if ((x_overlap != -1.f) && ((x_overlap < 0.0f) || (x_overlap >= 1.0f))) {
-        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("0 <= x_overlap < 1"));
-    }
-
-    if ((y_overlap != -1.0f) && ((y_overlap < 0.0f) || (y_overlap >= 1.0f))) {
-        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("0 <= y_overlap < 1"));
-    }
-
-    fb_alloc_mark();
-    py_tf_alloc_putchar_buffer();
-
-    py_tf_model_obj_t *model = py_tf_load_alloc(pos_args[0]);
-    uint8_t *tensor_arena = fb_alloc(model->params.tensor_arena_size, FB_ALLOC_PREFER_SPEED | FB_ALLOC_CACHE_ALIGN);
-
-    mp_obj_t objects_list = mp_obj_new_list(0, NULL);
-
-    for (float scale = 1.0f; scale >= min_scale; scale *= scale_mul) {
-        // Either provide a subtle offset to center multiple detection windows or center the only detection window.
-        for (int y = roi.y + ((y_overlap != -1.0f)
-                ? (fmodf(roi.h, (roi.h * scale)) / 2.0f)
-                : ((roi.h - (roi.h * scale)) / 2.0f));
-             // Finish when the detection window is outside of the ROI.
-             (y + (roi.h * scale)) <= (roi.y + roi.h);
-             // Step by an overlap amount accounting for scale or just terminate after one iteration.
-             y += ((y_overlap != -1.0f) ? (roi.h * scale * (1.0f - y_overlap)) : roi.h)) {
-            // Either provide a subtle offset to center multiple detection windows or center the only detection window.
-            for (int x = roi.x + ((x_overlap != -1.0f)
-                    ? (fmodf(roi.w, (roi.w * scale)) / 2.0f)
-                    : ((roi.w - (roi.w * scale)) / 2.0f));
-                 // Finish when the detection window is outside of the ROI.
-                 (x + (roi.w * scale)) <= (roi.x + roi.w);
-                 // Step by an overlap amount accounting for scale or just terminate after one iteration.
-                 x += ((x_overlap != -1.0f) ? (roi.w * scale * (1.0f - x_overlap)) : roi.w)) {
-
-                rectangle_t new_roi;
-                rectangle_init(&new_roi, x, y, roi.w * scale, roi.h * scale);
-
-                if (rectangle_overlap(&roi, &new_roi)) {
-                    // Check if new_roi is null...
-
-                    py_tf_input_callback_data_t py_tf_input_callback_data;
-                    py_tf_input_callback_data.img = image;
-                    py_tf_input_callback_data.roi = &new_roi;
-
-                    mp_obj_t py_tf_classify_output_callback_data;
-
-                    if (libtf_invoke(model->model_data,
-                                     tensor_arena,
-                                     &model->params,
-                                     py_tf_input_callback,
-                                     &py_tf_input_callback_data,
-                                     py_tf_classify_output_callback,
-                                     &py_tf_classify_output_callback_data) != 0) {
-                        // Note can't use MP_ERROR_TEXT here.
-                        mp_raise_msg(&mp_type_OSError, (mp_rom_error_text_t) py_tf_putchar_buffer);
-                    }
-
-                    py_tf_classification_obj_t *o = m_new_obj(py_tf_classification_obj_t);
-                    o->base.type = &py_tf_classification_type;
-                    o->x = mp_obj_new_int(new_roi.x);
-                    o->y = mp_obj_new_int(new_roi.y);
-                    o->w = mp_obj_new_int(new_roi.w);
-                    o->h = mp_obj_new_int(new_roi.h);
-                    o->output = py_tf_classify_output_callback_data;
-                    mp_obj_list_append(objects_list, o);
-                }
-            }
-        }
-    }
-
-    fb_alloc_free_till_mark();
-
-    return objects_list;
-}
-STATIC MP_DEFINE_CONST_FUN_OBJ_KW(py_tf_classify_obj, 2, py_tf_classify);
-
 STATIC void py_tf_segment_output_callback(void *callback_data,
                                           void *model_output,
                                           libtf_parameters_t *params) {
     mp_obj_t *arg = (mp_obj_t *) callback_data;
-
-    // Disable checking output scaling and zero-point. Nets can be all over the place on the output
-    // scaling and zero-point but still work with the code below.
-
-    // if (params->output_datatype == LIBTF_DATATYPE_UINT8) {
-    //     if (fast_roundf(params->output_scale * GRAYSCALE_RANGE) != 1) {
-    //         mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected model output scale to be 1/255!"));
-    //     }
-
-    //     if (params->output_zero_point != 0) {
-    //         mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected model output zero point to be 0!"));
-    //     }
-    // }
-
-    // if (params->output_datatype == LIBTF_DATATYPE_INT8) {
-    //     if (fast_roundf(params->output_scale * GRAYSCALE_RANGE) != 1) {
-    //         mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected model output scale to be 1/255!"));
-    //     }
-
-    //     if (params->output_zero_point != -GRAYSCALE_MID) {
-    //         mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected model output zero point to be -128!"));
-    //     }
-    // }
 
     int shift = (params->output_datatype == LIBTF_DATATYPE_INT8) ? GRAYSCALE_MID : 0;
 
@@ -578,9 +478,12 @@ STATIC void py_tf_segment_output_callback(void *callback_data,
 }
 
 STATIC mp_obj_t py_tf_segment(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
-    enum { ARG_roi };
+    enum { ARG_roi, ARG_scale, ARG_mean, ARG_stdev };
     static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_roi, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} }
+        { MP_QSTR_roi, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_scale, MP_ARG_INT | MP_ARG_KW_ONLY, {.u_int = PY_TF_SCALE_0_1} },
+        { MP_QSTR_mean, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_stdev, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
     };
 
     // Parse args.
@@ -596,9 +499,15 @@ STATIC mp_obj_t py_tf_segment(uint n_args, const mp_obj_t *pos_args, mp_map_t *k
     py_tf_model_obj_t *model = py_tf_load_alloc(pos_args[0]);
     uint8_t *tensor_arena = fb_alloc(model->params.tensor_arena_size, FB_ALLOC_PREFER_SPEED | FB_ALLOC_CACHE_ALIGN);
 
-    py_tf_input_callback_data_t py_tf_input_callback_data;
-    py_tf_input_callback_data.img = image;
-    py_tf_input_callback_data.roi = &roi;
+    py_tf_input_callback_data_t py_tf_input_callback_data = {
+        .img = image,
+        .roi = &roi,
+        .scale = args[ARG_scale].u_int,
+        .mean = {0.0f, 0.0f, 0.0f},
+        .stdev = {1.0f, 1.0f, 1.0f}
+    };
+    py_helper_arg_to_float_array(args[ARG_mean].u_obj, py_tf_input_callback_data.mean, 3);
+    py_helper_arg_to_float_array(args[ARG_stdev].u_obj, py_tf_input_callback_data.stdev, 3);
 
     mp_obj_t py_tf_segment_output_callback_data;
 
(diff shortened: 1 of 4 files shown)


src/omv/modules/py_tf.c
```diff
@@ -829,6 +842,346 @@ STATIC mp_obj_t py_tf_regression(mp_obj_t model_obj, mp_obj_t array_obj) {
...
+    { MP_ROM_QSTR(MP_QSTR_y),                   MP_ROM_PTR(&py_tf_model_output_y_obj) },
+    { MP_ROM_QSTR(MP_QSTR_w),                   MP_ROM_PTR(&py_tf_model_output_w_obj) },
+    { MP_ROM_QSTR(MP_QSTR_h),                   MP_ROM_PTR(&py_tf_model_output_h_obj) },
+    { MP_ROM_QSTR(MP_QSTR_input_height),        MP_ROM_PTR(&py_tf_model_output_input_height_obj) },
+    { MP_ROM_QSTR(MP_QSTR_input_width),         MP_ROM_PTR(&py_tf_model_output_input_width_obj) },
+    { MP_ROM_QSTR(MP_QSTR_input_channels),      MP_ROM_PTR(&py_tf_model_output_input_channels_obj) },
+    { MP_ROM_QSTR(MP_QSTR_output_height),       MP_ROM_PTR(&py_tf_model_output_output_height_obj) },
+    { MP_ROM_QSTR(MP_QSTR_output_width),        MP_ROM_PTR(&py_tf_model_output_output_width_obj) },
+    { MP_ROM_QSTR(MP_QSTR_output_channels),     MP_ROM_PTR(&py_tf_model_output_output_channels_obj) },
+    { MP_ROM_QSTR(MP_QSTR_value),               MP_ROM_PTR(&py_tf_model_output_value_obj) },
```
iabdalkader, line 1090: Can we replace those with attributes ? See example in
`ndarray_properties_attr` in ulab or `ssl_context_attr` in `extmod/modtls_mbedtls.c`
And replace `value()` with `subscr` (for this there are many examples in `modules/py_image.c`)

kwagyeman: Does subscr support 3D access? I guess I could make the property names shorter, and then it doesn't matter.

kwagyeman: Done.

kwagyeman:
[lpd-yolov5-int8-quantized.zip](https://github.com/user-attachments/files/15541888/lpd-yolov5-int8-quantized.zip)

Here are all the files to test with.

src/omv/modules/py_tf.c
kwagyeman: Note, I removed 3D subscript access. This was slowing things down by 100x.

kwagyeman:
Replaced classify and regression with invoke.

iabdalkader:

https://forums.openmv.io/t/image-segmentation-in-openmv-h7-plus/9674/2

iabdalkader:
I'm testing this and there's a problem with both libtf and this PR, there's no support for multioutput models. Libtf just returns the first output:
```C
        input_callback(input_callback_data, interpreter.input(0)->data.data, params);

        if (interpreter.Invoke() != kTfLiteOk) {
            error_reporter->Report("Invoke() failed!");
            return 1;
        }

        output_callback(output_callback_data, interpreter.output(0)->data.data, params);
```

src/omv/modules/py_tf.c
```diff
@@ -803,31 +708,384 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+    o->y = mp_obj_new_int(arg->roi->y);
+    o->w = mp_obj_new_int(arg->roi->w);
+    o->h = mp_obj_new_int(arg->roi->h);
+    o->rect = mp_obj_new_tuple(4, (mp_obj_t []) {o->x, o->y, o->w, o->h});
+    o->input_height = mp_obj_new_int(params->input_height);
+    o->input_width = mp_obj_new_int(params->input_width);
+    o->input_channels = mp_obj_new_int(params->input_channels);
+    o->output_height = mp_obj_new_int(params->output_height);
+    o->output_width = mp_obj_new_int(params->output_width);
+    o->output_channels = mp_obj_new_int(params->output_channels);
```
iabdalkader, line 1013: input/output_shape tuples  ?

src/omv/modules/py_tf.c
```diff
@@ -803,31 +708,384 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+                                         &py_tf_input_callback_data,
+                                         py_tf_invoke_output_callback,
+                                         &py_tf_invoke_output_callback_data);
+        } else {
+            invoke_result = libtf_invoke(model->model_data,
+                                         tensor_arena,
+                                         &model->params,
+                                         py_tf_input_callback,
+                                         &py_tf_input_callback_data,
+                                         py_tf_classify_output_callback,
```
iabdalkader, line 1074:
```suggestion
                                         py_tf_classify_output_callback,
```
I don't think should be called `py_tf_classify_output_callback` anymore, maybe just `output_callback`.

src/omv/modules/py_tf.c
```diff
@@ -803,31 +708,384 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+    o->input_width = mp_obj_new_int(params->input_width);
+    o->input_channels = mp_obj_new_int(params->input_channels);
+    o->output_height = mp_obj_new_int(params->output_height);
+    o->output_width = mp_obj_new_int(params->output_width);
+    o->output_channels = mp_obj_new_int(params->output_channels);
+    list_init(&o->bounding_boxes, sizeof(py_tf_model_output_bounding_box_lnk_data_t));
+    *(arg->out) = mp_call_function_1(arg->callback, o);
+}
+
+STATIC mp_obj_t py_tf_invoke(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
```
iabdalkader, line 1018: How about `py_tf_predict` like tensorflow ?

iabdalkader (changes requested):

src/omv/modules/py_tf.c
iabdalkader: Can we index using a tuple like numpy does ? Not this will be important for multi-output models, example:
```Python
In [28]: import numpy as np
In [29]: a = np.array([[1, 2, 3], [4, 5, 6]])
In [30]: a[(0, 2)]
Out[30]: 3
```

kwagyeman: No, you can't use any tuples. Their creation causes a malloc. The speed slow-down is ridiculous.

I got to the current way the API works after starting with thinking we could do tuples and realizing it was a hard no.

kwagyeman:
Ah, I didn't know multiple output models were a thing. Okay, let me think about that.

src/omv/modules/py_tf.c
```diff
@@ -803,31 +708,384 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+    }
+
+    return mp_const_none;
+}
+STATIC MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(py_tf_model_output_add_bounding_box_obj, 7, 7,
+                                           py_tf_model_output_add_bounding_box);
+
+// This convenience function will perform non-maximal-suppression on the bounding boxes you have collected
+// using the add_bounding_box function. It returns the final list of bounding boxes mapped to the image
+// roi for you and separated by label index onto their own lists. You do not have to use it to parse model output.
```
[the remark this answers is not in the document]
kwagyeman: Will rebase the commit.

src/omv/modules/py_tf.c
```diff
@@ -803,31 +708,384 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+    mp_arg_parse_all(n_args - 1, pos_args + 1, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
+
+    py_tf_model_output_obj_t *self_in = (py_tf_model_output_obj_t *) pos_args[0];
+    float threshold = py_helper_arg_to_float(args[ARG_threshold].u_obj, 0.1f);
+    float sigma = py_helper_arg_to_float(args[ARG_sigma].u_obj, 0.1f);
+
+    // We're using soft non-max supression below with a gaussian as this provides the best results.
+    // The gaussian is used to give a soft score penalty to overlapping boxes. On loop
+    // entry "bounding_boxes" is sorted but after each iteration we have to pick the next highest next
+    // score again given the score penalty changes the order.
```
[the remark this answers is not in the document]
kwagyeman: Will rebase the commit.

src/omv/modules/py_tf.c
```diff
@@ -265,41 +265,79 @@ STATIC py_tf_model_obj_t *py_tf_load_alloc(mp_obj_t path_obj) {
...
-    int shift = (params->input_datatype == LIBTF_DATATYPE_INT8) ? GRAYSCALE_MID : 0;
-    float fscale = 1.0f / GRAYSCALE_RANGE;
+    float fscale_r = fscale, fadd_r = fadd;
+    float fscale_g = fscale, fadd_g = fadd;
+    float fscale_b = fscale, fadd_b = fadd;
+
+    // ImageNet Mean and StdDev Normalization.
+    //
+    // To normalize the input image we need to subtract the mean and divide by the standard deviation.
+    // We can do this by applying the normalization to fscale and fadd outside the loop.
```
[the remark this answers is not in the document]
kwagyeman: It's based on the ImageNet Mean/Std. These are known values.

> Also I'm not sure that scaling the mean and std first and then using them to normalize the image is the same as normalizing the image and then scaling it. Any proof that they are ?

I was following what Jan mentioned we needed to support:

{ label: 'Pixels ranging 0..1 (not normalized)', value: '0..1' },
{ label: 'Pixels ranging -1..1 (not normalized)', value: '-1..1' },
{ label: 'Pixels ranging -128..127 (not normalized)', value: '-128..127' },
{ label: 'Pixels ranging 0..255 (not normalized)', value: '0..255' },
{ label: 'PyTorch (scale to 0..1, then normalize using ImageNet mean/std)', value: 'torch' },

So, scale first and then normalize. The normalization operation is to subtract the mean and then divide by the stdev. I then worked out the math to avoid having to do two multiplies/adds in the loop and got it down to just one mutliply/add by factoring this back in.

E.g.

(((value * scale) + offset) - mean) / stdiv

Offset and mean and be combined.

(((value * scale) + (offset-mean)) / stdiv

Then stdev can be distributed

((value * scale)/stdiv) + ((offset-mean)/stdiv)

So fmul becomes (scale/fdiv) and fadd becomes ((offset-mean)/stdiv)

src/omv/modules/py_tf.c
```diff
@@ -803,31 +708,384 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+    list_init(&o->bounding_boxes, sizeof(py_tf_model_output_bounding_box_lnk_data_t));
+    *(arg->out) = mp_call_function_1(arg->callback, o);
+}
+
+STATIC mp_obj_t py_tf_invoke(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    enum { ARG_roi, ARG_callback, ARG_input_scale };
+    static const mp_arg_t allowed_args[] = {
+        { MP_QSTR_roi, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_callback, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_input_scale, MP_ARG_INT | MP_ARG_KW_ONLY, {.u_int = PY_TF_SCALE_0_1} },
```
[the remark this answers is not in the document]
kwagyeman: It needs to be PY_TF_SCALE_0_1 or classify's default behavior breaks for all our current customers.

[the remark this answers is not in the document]
kwagyeman: Was trying to match Jan's request with regards to the scaling options. 0-255 was called out. That said, yes, if it effectively applies, no scale.

src/omv/modules/py_tf.c
```diff
@@ -578,9 +479,10 @@ STATIC void py_tf_segment_output_callback(void *callback_data,
 }
 
 STATIC mp_obj_t py_tf_segment(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
```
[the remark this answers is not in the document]
kwagyeman: It outputs [h][w][c]. It could be replaced with invoke() and then a custom callback that makes an image. However, this should be done in another PR.

I'll need to think about how this should work... we don't want to end up in a situation where the time it takes the build the output image is so slow in Python that you can't get over 1 FPS when we have new faster chips. I think the best thing to do here is wait until I have a test network. Then I can try to figure out a pattern that's flexible and has good performance.

src/omv/modules/py_tf.c
```diff
@@ -803,31 +708,384 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+typedef struct py_tf_model_output_obj {
+    mp_obj_base_t base;
+    rectangle_t *roi;
+    void *model_output;
+    libtf_parameters_t *params;
+    // Pre-compute for lookup speed.
+    size_t output_size;
+    mp_obj_t x, y, w, h, rect;
+    mp_obj_t input_height, input_width, input_channels;
+    mp_obj_t output_height, output_width, output_channels;
```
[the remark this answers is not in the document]
kwagyeman: I don't mind allocating a tuple internally, but, the result to the end user has to be static. We cannot allocate tuples dynamically without a huge speed hit.

Are you thinking:

result.input_shape[0] versus result.input_height, result.input_shape[1] versus result.input_width, and result.input_shape[2] versus result.input_channels?

src/omv/modules/py_tf.c
```diff
@@ -803,31 +708,384 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+
+typedef struct py_tf_model_output_bounding_box_lnk_data {
+    rectangle_t rect;
+    float score;
+    int label_index;
+} py_tf_model_output_bounding_box_lnk_data_t;
+
+// This is a convenience function that will collect bounding boxes for you so that you can get
+// a non-maximal-suppression of them later. You do not have to use it to parse model output.
+// Not using mp_arg_parse_all() on purpose here to make sure this method is fast.
```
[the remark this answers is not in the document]
kwagyeman: Will rebase the commit.

src/omv/modules/py_tf.c
```diff
@@ -803,31 +708,384 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+STATIC mp_obj_t py_tf_model_output_add_bounding_box(uint n_args, const mp_obj_t *pos_args) {
+    enum { ARG_self, ARG_xmin, ARG_ymin, ARG_xmax, ARG_ymax, ARG_score, ARG_label_index };
+    py_tf_model_output_obj_t *self_in = (py_tf_model_output_obj_t *) pos_args[ARG_self];
+
+    py_tf_model_output_bounding_box_lnk_data_t lnk_data;
+    lnk_data.score = mp_obj_get_float(pos_args[ARG_score]);
+
+    if ((lnk_data.score >= 0.0f) && (lnk_data.score <= 1.0f)) {
+        float xmin = mp_obj_get_float(pos_args[ARG_xmin]);
```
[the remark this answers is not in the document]
kwagyeman: Will do.

src/omv/modules/py_tf.c
```diff
@@ -803,31 +708,384 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+
+    py_tf_model_output_obj_t *self_in = (py_tf_model_output_obj_t *) pos_args[0];
+    float threshold = py_helper_arg_to_float(args[ARG_threshold].u_obj, 0.1f);
+    float sigma = py_helper_arg_to_float(args[ARG_sigma].u_obj, 0.1f);
+
+    // We're using soft non-max supression below with a gaussian as this provides the best results.
+    // The gaussian is used to give a soft score penalty to overlapping boxes. On loop
+    // entry "bounding_boxes" is sorted but after each iteration we have to pick the next highest next
+    // score again given the score penalty changes the order.
+
```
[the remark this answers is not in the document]
kwagyeman: Yeah, it's quite useful. Could use it other places. Will do.

src/omv/modules/py_tf.c
kwagyeman: I will do this after clarification: https://github.com/openmv/openmv/pull/2227#discussion_r1632393552

src/omv/modules/py_tf.c
kwagyeman: Sure, I have no ideas for what to call this. If you have a better name let me know.

src/omv/modules/py_tf.c
kwagyeman: Sure, I called it invoke because it was a tensorflow thing.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
iabdalkader: > It's based on the ImageNet Mean/Std. These are known values.

If the network is trained on any other dataset, the mean and std will be different than those hard-coded values from image net. Imagine a network trained on bright/dark images. The mean and standard deviation should be args just like the scaling.

Second, about the optimization above it's just not producing the same values, unless I messed a step ?

```Python
import numpy as np

# Set the random seed for reproducibility
np.random.seed(42)

# Step 1: Generate a 4x4 random ndarray with values from 0 to 255
array = np.random.randint(0, 256, size=(4, 4))
print("Original array:\n", array)

# Step 2: Calculate the mean and standard deviation
m = np.mean(array)
s = np.std(array)
print("\nMean:", m)
print("Standard Deviation:", s)

# Step 3: Standardize the array
standardized_array = (array - m) / s
print("\nStandardized array:\n", standardized_array)

# Step 4: Scale the array from 0 to 1
min_val = np.min(standardized_array)
max_val = np.max(standardized_array)
scaled_array = (standardized_array - min_val) / (max_val - min_val)
print("\nScaled array:\n", scaled_array)

# Mean and STD were calculated on an array of 0->255
m /= 255
s /= 255

# Pretty much copied from the C code
fadd = 0.0
fscale = 1.0

# PY_TF_SCALE_0_1_NORMALIZED
fscale = 1.0 / 255.0

# Grayscale
fadd = (fadd - m) / s
fscale /= s
scaled_array = array * fscale + fadd
print("\nScaled array:\n", scaled_array)

```

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
iabdalkader: Okay we can leave `PY_TF_SCALE_0_1` as the default, but then `None` should be no-scaling (meaning both 0-255 options aren't really needed).

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: Let me check.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: Okay, will do.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: Just looked into this. The implementation right now is specific to bounding box detectors. I should wait to rebase it until I see how I'd use it again in another place.

src/omv/modules/py_tf.c
kwagyeman: Done

src/omv/modules/py_tf.c
kwagyeman: Done

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: Yeah, the scale you are using is different: scaled_array = (standardized_array - min_val) / (max_val - min_val)

So, that wouldn't necessarily be 1/255. My code is taking advantage of the fact that the input range is fixed and known to be 0-255. It doesn't try to maximize the dynamic range though because it doesn't need to to. The net just expects the floats to be within a certain range.

> If the network is trained on any other dataset, the mean and std will be different than those hard-coded values from image net. Imagine a network trained on bright/dark images. The mean and standard deviation should be args just like the scaling.

True, this specifically has to do with the with models that started from transfer learning off of something initially trained on ImageNet. 

Okay, let me allow you to customize this.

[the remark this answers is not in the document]
kwagyeman: Done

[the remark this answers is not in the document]
iabdalkader: > Yeah, the scale you are using is different: scaled_array = (standardized_array - min_val) / (max_val - min_val)
> 
> So, that wouldn't necessarily be 1/255. My code is taking advantage of the fact that the input range is fixed and known to be 0-255. It doesn't try to maximize the dynamic range though because it doesn't need to to. The net just expects the floats to be within a certain range.

Even if I use 1/255 I get the same output more or less:
```
Scaled array:
 [[0.4        0.70196078 0.36078431 0.05490196]
 [0.41568627 0.27843137 0.7372549  0.07843137]
 [0.4        0.4745098  0.82352941 0.83921569]
 [0.29019608 0.79215686 0.34117647 0.45490196]]
```

On the other hand, the C code doesn't seem to do any scaling, unless I missed a step:
```
Scaled array:
 [[-0.27098482  0.98410276 -0.4339832  -1.70537062]
 [-0.20578546 -0.77627982  1.13080131 -1.60757159]
 [-0.27098482  0.03871212  1.48939776  1.55459711]
 [-0.7273803   1.35899905 -0.5154824  -0.04278708]]
```

> True, this specifically has to do with the with models that started from transfer learning off of something initially trained on ImageNet.

Imagenet weights are very popular so they should be the default, but if you train something from scratch on a custom dataset, it should allow you to provide the mean and standard deviation of the input.

src/omv/modules/py_tf.c
iabdalkader: Understood.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
iabdalkader: In `py_tf_invoke_output_callback` you already allocate tuples for the `rect`. There won't be an extra alloc when you access that tuple in `py_tf_model_output_attr` via `self->input_shape`.

Btw why is there a `rect` and `x, y, w, h` it seems redundant.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
iabdalkader: Okay no problem. Note you may want to split the comments above so the code formatter doesn't complain.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
iabdalkader: Also it seems that all this information is already accessible from the model object, right ? If that's the case, why duplicate it in the output object and waste time copying it all over ? The callback should have access to the model object in Python. They still use functions but we can switch them to attributes.

iabdalkader:
> Ah, I didn't know multiple output models were a thing. Okay, let me think about that.

They're very popular, it's also possible to have multiple inputs (probably less common). I'm attaching a multi-output model for testing. It outputs 2 vectors/lists `([p0, p1], [x1, y1, x2, y2])`

> Okay, looked into this more. Adding multiple output support should be done in a different PR. I can talk
 about what needs to be done in our meeting. We'll need to vectorize a lot of stuff, so it's changes all over the place in py_tf. So, we should do it as a separate task.

No problem. But until it's done the utility of `tf_predict` is very limited.

[model.tflite.zip](https://github.com/user-attachments/files/15767720/model.tflite.zip)

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
iabdalkader: If it outputs a small image/map, we can pass it back to some C code along with the main image to do the processing. A function in py_image for example.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: The script did the order of operations wrong.

'PyTorch (scale to 0..1, then normalize using ImageNet mean/std)',

Note here that the normalized/standardized image is not from 0->1. The input just needs to be in that range since the ImageNet mean/stdev are like mean = [0.485, 0.456, 0.406] and std = [0.229, 0.224, 0.225]. So, things won't work unless the input values to this are in the range of 0->1.

```
import numpy as np

# Set the random seed for reproducibility
np.random.seed(42)

# Step 1: Generate a 4x4 random ndarray with values from 0 to 255
array = np.random.randint(0, 256, size=(4, 4))
print("Original array:\n", array)

# Step 2: Calculate the mean and standard deviation
m = np.mean(array)
s = np.std(array)
print("\nMean:", m)
print("Standard Deviation:", s)

# Step 3: Scale the array from 0 to 1
scaled_array = array / 255.0
print("\nScaled array:\n", scaled_array)

# Mean and STD were calculated on an array of 0->255
m /= 255
s /= 255

scaled_array_2 = (scaled_array - m) / s
print("\nScaled->Standarized array:\n", scaled_array_2)

# Pretty much copied from the C code
fadd = 0.0
fscale = 1.0

# PY_TF_SCALE_0_1_NORMALIZED
fscale = 1.0 / 255.0

# Grayscale
fadd = (fadd - m) / s
fscale /= s
scaled_array = array * fscale + fadd
print("\nC Code Scaled->Standarized array:\n", scaled_array)
```

```
Original array:
 [[102 179  92  14]
 [106  71 188  20]
 [102 121 210 214]
 [ 74 202  87 116]]

Mean: 118.625
Standard Deviation: 61.350300529011264

Standardized array:
 [[-0.27098482  0.98410276 -0.4339832  -1.70537062]
 [-0.20578546 -0.77627982  1.13080131 -1.60757159]
 [-0.27098482  0.03871212  1.48939776  1.55459711]
 [-0.7273803   1.35899905 -0.5154824  -0.04278708]]

Scaled array:
 [[0.44  0.825 0.39  0.   ]
 [0.46  0.285 0.87  0.03 ]
 [0.44  0.535 0.98  1.   ]
 [0.3   0.94  0.365 0.51 ]]

Scaled array:
 [[-0.27098482  0.98410276 -0.4339832  -1.70537062]
 [-0.20578546 -0.77627982  1.13080131 -1.60757159]
 [-0.27098482  0.03871212  1.48939776  1.55459711]
 [-0.7273803   1.35899905 -0.5154824  -0.04278708]]
```

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/modules/py_tf.c
kwagyeman: Done

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: I changed everything to shape into the model output object but didn't switch to passing the main object in the callback.  I also removed x/y/w/h and just have the rect object. Since we released the API previously as functions it will break any code in the wild to turn them into attributes. The "()" at the end of any previous function call that gets turned into an attribute will cause MicroPython to then try to call the value of the attribute returned.

I agree we should use attributes moving forward. But, I don't think we should go break the API of older stuff.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: Okay, will redo the segment code once we have a network to test with.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: Example code updated.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
iabdalkader: > The script did the order of operations wrong.

Not necessarily, if the model expects the input to be 0->1 then the script's order is the right order. That's what matters most, the expected input values ranges. Things get lost in chats maybe you want to double check this. For now we can just merge it as it's not that popular anyway.

src/omv/modules/py_tf.c
```diff
@@ -265,41 +265,83 @@ STATIC py_tf_model_obj_t *py_tf_load_alloc(mp_obj_t path_obj) {
     }
 }
 
+typedef enum {
+    PY_TF_SCALE_NONE,
+    PY_TF_SCALE_0_1,
+    PY_TF_SCALE_N1_1,
+    PY_TF_SCALE_N128_127
```
iabdalkader, line 272:
```suggestion
    PY_TF_SCALE_S1_1,
    PY_TF_SCALE_S128_127
```

src/omv/modules/py_tf.c
```diff
@@ -265,41 +265,83 @@ STATIC py_tf_model_obj_t *py_tf_load_alloc(mp_obj_t path_obj) {
...
+    PY_TF_G_STDDEV,
+    PY_TF_B_STDDEV,
+    PY_TF_INPUT_NORMALIZATION_SIZE
+} py_tf_input_normalization_t;
+
 typedef struct py_tf_input_callback_data {
     image_t *img;
     rectangle_t *roi;
+    py_tf_input_scale_t input_scale;
+    float input_normalization[PY_TF_INPUT_NORMALIZATION_SIZE];
```
iabdalkader, line 289:
```suggestion
    py_tf_input_scale_t input_scale;
    float input_stats[PY_TF_STATS_SIZE];
```

src/omv/modules/py_tf.c
```diff
@@ -596,9 +503,13 @@ STATIC mp_obj_t py_tf_segment(uint n_args, const mp_obj_t *pos_args, mp_map_t *k
...
     uint8_t *tensor_arena = fb_alloc(model->params.tensor_arena_size, FB_ALLOC_PREFER_SPEED | FB_ALLOC_CACHE_ALIGN);
 
-    py_tf_input_callback_data_t py_tf_input_callback_data;
+    py_tf_input_callback_data_t py_tf_input_callback_data =
+    {.input_normalization = {0.0f, 0.0f, 0.0f, 1.0f, 1.0f, 1.0f}};
     py_tf_input_callback_data.img = image;
     py_tf_input_callback_data.roi = &roi;
+    py_tf_input_callback_data.input_scale = args[ARG_input_scale].u_int;
+    py_helper_arg_to_float_array(args[ARG_input_normalization].u_obj,
+                                 py_tf_input_callback_data.input_normalization, PY_TF_INPUT_NORMALIZATION_SIZE);
```
iabdalkader, line 512:
```suggestion
```suggestion
    py_tf_input_callback_data_t py_tf_input_callback_data = {
        .img = image,
        .roi = &roi,
        .input_scale = args[ARG_input_scale].u_int,
        .input_stats = {0.0f, 0.0f, 0.0f, 1.0f, 1.0f, 1.0f},
    };
    if (args[ARG_input_stats].u_obj != mp_const_none) {
        py_helper_arg_to_float_array(args[ARG_input_stats].u_obj, py_tf_input_callback_data.input_stats, PY_TF_INPUT_STATS_SIZE);
    }
```

src/omv/modules/py_tf.c
```diff
@@ -620,11 +531,13 @@ STATIC mp_obj_t py_tf_segment(uint n_args, const mp_obj_t *pos_args, mp_map_t *k
 STATIC MP_DEFINE_CONST_FUN_OBJ_KW(py_tf_segment_obj, 2, py_tf_segment);
 
 STATIC mp_obj_t py_tf_detect(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
-    enum { ARG_roi, ARG_thresholds, ARG_invert };
+    enum { ARG_roi, ARG_thresholds, ARG_invert, ARG_input_scale, ARG_input_normalization };
```
iabdalkader, line 534: Should be the same as `py_tf_segement` see the review comments above.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
iabdalkader: > I agree we should use attributes moving forward. But, I don't think we should go break the API of older stuff.

Please implement what we discussed in the meeting. The API to access those attributes isn't that common, before the callback you had no reason to read the input/output shape or channels. Also APIs change, we don't have to get stuck with a poor API. The change is for the better.

src/omv/modules/py_tf.c
```diff
@@ -265,41 +265,83 @@ STATIC py_tf_model_obj_t *py_tf_load_alloc(mp_obj_t path_obj) {
...
+
+typedef enum {
+    PY_TF_R_MEAN,
+    PY_TF_G_MEAN,
+    PY_TF_B_MEAN,
+    PY_TF_R_STDDEV,
+    PY_TF_G_STDDEV,
+    PY_TF_B_STDDEV,
+    PY_TF_INPUT_NORMALIZATION_SIZE
+} py_tf_input_normalization_t;
```
iabdalkader, line 283:
```suggestion
typedef enum {
    PY_TF_R_MEAN,
    PY_TF_G_MEAN,
    PY_TF_B_MEAN,
    PY_TF_R_STDDEV,
    PY_TF_G_STDDEV,
    PY_TF_B_STDDEV,
    PY_TF_INPUT_STATS_SIZE
} py_tf_input_stats_t;

A better, shorter, name for this is `input_stats` which they are.
```

src/omv/modules/py_tf.c
```diff
@@ -578,9 +483,11 @@ STATIC void py_tf_segment_output_callback(void *callback_data,
 }
 
 STATIC mp_obj_t py_tf_segment(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
-    enum { ARG_roi };
+    enum { ARG_roi, ARG_input_scale, ARG_input_normalization };
```
iabdalkader, line 486:
```suggestion
    enum { ARG_roi, ARG_input_scale, ARG_input_stats };
```

src/omv/modules/py_tf.c
```diff
@@ -578,9 +483,11 @@ STATIC void py_tf_segment_output_callback(void *callback_data,
...
     static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_roi, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} }
+        { MP_QSTR_roi, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_input_scale, MP_ARG_INT | MP_ARG_KW_ONLY, {.u_int = PY_TF_SCALE_0_1} },
+        { MP_QSTR_input_normalization, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
```
iabdalkader, line 490:
```suggestion
        { MP_QSTR_input_stats, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
```

src/omv/modules/py_tf.c
```diff
@@ -803,31 +720,382 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+    *(arg->out) = mp_call_function_1(arg->callback, o);
+}
+
+STATIC mp_obj_t py_tf_predict(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    enum { ARG_roi, ARG_callback, ARG_input_scale, ARG_input_normalization };
+    static const mp_arg_t allowed_args[] = {
+        { MP_QSTR_roi, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_callback, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_input_scale, MP_ARG_INT | MP_ARG_KW_ONLY, {.u_int = PY_TF_SCALE_0_1} },
+        { MP_QSTR_input_normalization, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
```
iabdalkader, line 1030:
```suggestion
        { MP_QSTR_input_stats, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
```
Should be the same as `py_tf_segement` please see review comments above.

src/omv/modules/py_tf.c
```diff
@@ -803,31 +720,382 @@ STATIC void py_tf_regression_input_callback(void *callback_data,
...
+        image_t *image = py_helper_arg_to_image(pos_args[1], ARG_IMAGE_ANY);
+        rectangle_t roi = py_helper_arg_to_roi(args[ARG_roi].u_obj, image);
+
+        py_tf_input_callback_data_t py_tf_input_callback_data =
+        {.input_normalization = {0.0f, 0.0f, 0.0f, 1.0f, 1.0f, 1.0f}};
+        py_tf_input_callback_data.img = image;
+        py_tf_input_callback_data.roi = &roi;
+        py_tf_input_callback_data.input_scale = args[ARG_input_scale].u_int;
+        py_helper_arg_to_float_array(args[ARG_input_normalization].u_obj,
+                                     py_tf_input_callback_data.input_normalization, PY_TF_INPUT_NORMALIZATION_SIZE);
```
iabdalkader, line 1064: Should be the same as `py_tf_segement` please see review comments above.

iabdalkader (changes requested):

iabdalkader:
Just one remaining major issue with the PR: the duplicate attributes/functions.

src/omv/modules/py_tf.c
kwagyeman: This is an outdated comment. I changed it previously to what was discussed in the meeting with mean/std.

src/omv/modules/py_tf.c
kwagyeman: This is an outdated comment. I changed it previously to what was discussed in the meeting with mean/std.

src/omv/modules/py_tf.c
kwagyeman: This is an outdated comment. I changed it previously to what was discussed in the meeting with mean/std.

src/omv/modules/py_tf.c
kwagyeman: This is an outdated comment. I changed it previously to what was discussed in the meeting with mean/std.

src/omv/modules/py_tf.c
kwagyeman: This is an outdated comment. I changed it previously to what was discussed in the meeting with mean/std.

src/omv/modules/py_tf.c
kwagyeman: This is an outdated comment. I changed it previously to what was discussed in the meeting with mean/std.

src/omv/modules/py_tf.c
kwagyeman: This is an outdated comment. I changed it previously to what was discussed in the meeting with mean/std.

src/omv/modules/py_tf.c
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/modules/py_tf.c
kwagyeman: Done

src/omv/modules/py_tf.c
kwagyeman: Done

Note: py_helper_arg_to_float_array does "if (args[ARG_input_stats].u_obj != mp_const_none)" internally.

iabdalkader:
> It's weird that you had PR comments on something I updated. Maybe I pushed the update wrong. Anyway, fixing things.

I think you forgot to push and all my comments were on some intermediate state.

iabdalkader:
@kwagyeman This looks fantastic now, great job thanks! 

Just one last comment though, the model object should be passed to the output callback, it doesn't need to be an attribute of the output. It's just one less level of indirection to access attributes in the callback:

```Python
callback(output):
   output.model.input_shape[0]
# vs
callback(model, output):
   model.input_shape[0]
```

To be clear it doesn't need to be passed to the callback either, since it can be accessed globally, but if it's more convenient to do so then it's fine with me.

kwagyeman:
Done, PR updated.

iabdalkader (approved):

iabdalkader:

https://forums.openmv.io/t/heap-ram-limitations-and-tf-module/9790/9

Issue #2219, opened 2024-05-20
Fix Invalidate Cache Issues


kwagyeman:
https://community.st.com/t5/stm32-mcus-products/maintaining-cpu-data-cache-coherence-for-dma-buffers/td-p/95746

Go through code and add any invalidates that need to be added.

PR #2228, opened 2024-06-05
imlib/draw: Add missing speculative read invalidates.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2219

### DIFF
--- a/src/omv/imlib/draw.c
+++ b/src/omv/imlib/draw.c
@@ -801,7 +801,7 @@ void imlib_draw_row_setup(imlib_draw_row_data_t *data) {
                 cfg.pCLUT = clut;
                 cfg.CLUTColorMode = DMA2D_CCM_ARGB8888;
                 cfg.Size = 255;
-                #if defined(MCU_SERIES_F7) || defined(MCU_SERIES_H7)
+                #if __DCACHE_PRESENT
                 SCB_CleanDCache_by_Addr(clut, 256 * sizeof(uint32_t));
                 #endif
                 HAL_DMA2D_CLUTLoad(&data->dma2d, cfg, 1);
@@ -828,6 +828,10 @@ void imlib_draw_row_setup(imlib_draw_row_data_t *data) {
         data->dma2d.LayerCfg[1].ChromaSubSampling = DMA2D_NO_CSS;
         #endif
         HAL_DMA2D_ConfigLayer(&data->dma2d, 1);
+        #if __DCACHE_PRESENT
+        // SCB_InvalidateDCache_by_Addr does nothing if dsize is 0.
+        data->dma2d_invalidate_dsize = 0;
+        #endif
     } else {
         data->row_buffer[1] = data->row_buffer[0];
     }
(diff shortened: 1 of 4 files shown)


src/omv/imlib/draw.c
```diff
@@ -766,6 +766,10 @@ void imlib_draw_row_setup(imlib_draw_row_data_t *data) {
         data->dma2d.LayerCfg[1].ChromaSubSampling = DMA2D_NO_CSS;
         #endif
         HAL_DMA2D_ConfigLayer(&data->dma2d, 1);
+        #if __DCACHE_PRESENT
+        // SCB_InvalidateDCache_by_Addr does nothing is dsize is 0.
```
iabdalkader, line 770:
```suggestion
        // SCB_InvalidateDCache_by_Addr does nothing if dsize is 0.
```

iabdalkader (changes requested):

src/omv/imlib/draw.c
kwagyeman: Fixed

kwagyeman:
I rechecked needing all of them. They are all required to follow the spec of invalidating cached reads to the DMA buffer.

src/omv/ports/stm32/jpeg.c
```diff
@@ -744,6 +751,9 @@ void jpeg_decompress(image_t *dst, image_t *src) {
 
     if ((JPEG_state.jpeg_descr.Conf.ColorSpace == JPEG_YCBCR_COLORSPACE) && dst->is_color) {
         HAL_DMA2D_DeInit(&DMA2D_Handle);
+
+        // Ensure any cached reads are dropped.
+        SCB_InvalidateDCache_by_Addr((uint32_t *) dst->data, image_size(dst));
```
iabdalkader, line 756: The whole dst image is invalidated right after configuring the DMA2D: https://github.com/openmv/openmv/blob/19d70d69e887cc64f67df653cc02f81db68be911/src/omv/ports/stm32/jpeg.c#L506-L510

Then every row/block is invalidated before the transfer: https://github.com/openmv/openmv/blob/19d70d69e887cc64f67df653cc02f81db68be911/src/omv/ports/stm32/jpeg.c#L712-L714

Now here `dst` is invalidated again after all transfers are complete.  The first invalidate is definitely Not needed, as every block will be invalidated anyway before the transfer, or all 3 of them can be replaced with just one invalidate after the `DMA2D_PollForTransfer`.

iabdalkader (changes requested):

src/omv/ports/stm32/jpeg.c
kwagyeman: Yeah, that's why I dropped `SCB_InvalidateDCache_by_Addr((uint32_t *) rp, dst->w * dy * sizeof(uint16_t));` in the PR given the whole image is already invalidated before starting DMA2D.

> Also possible to replace all 3 of them with just one invalidate before/after the DMA2D_PollForTransfer.

Yeah, you could, there's less overhead doing them all in one go though before and after than before and after every row though right?

iabdalkader: > before and after everything than before and after every row, right?

I think so.

kwagyeman: Okay, then this PR is ready.

iabdalkader: BTW if you want to save some time and optimize this, you could try invalidating the previous line/block *before* the poll.

iabdalkader: That is invalidate while the DMA is transferring the next line/block.

kwagyeman: Implemented the suggestion. Works.

iabdalkader (approved):

PR #2231, opened 2024-06-16
modules/py_tf: Refactor py_tf module.

iabdalkader:
* Implement log handler.
* Remove all extra load functions (`load`, `load_builtin_model`, and the newly added `tf.Model()` now all call the same function)
* Remove module-level functions. Code should load the model first, the use `model.predict()`. This is a breaking change, but loading models on the fly was never a good feature.
* Model fb_memory will be free'd in finalizer.  The model's fb_memory is automatically free'd when it's delete (i.e., `del model`).


kwagyeman (approved):

kwagyeman:
@iabdalkader - Was able to follow the PR by looking at the RAW file.

PR #2234, opened 2024-06-20
modules/py_tf: Refactor predict callback code.

kwagyeman:
This PR does the following:

0: A bug commit is included to make loaded networks work still.

1: I noticed that I was doing free() and then new() to move items between lists. So, I updated the list code to have the ability to move items between lists without doing that. I then noticed the list code could be significantly cleaned up and reduced all the logic down to just two functions list_link() and list_unlink(). The rest of the functions are then 1-2 liners around these.

2: I refactored the NMS code into it's own object. I like this improved interface for it. I left the new object in py_tf for the moment. It can be moved into a new module as it has no dependencies on anything in py_tf. However, I'm unsure what that new module should be named if we want to do that.

3: I merged the model output object with the model object.

4: I removed the classification object/method and updated the tensorflow example scripts. Note that the removal of this object/method is a breaking API change. That said, it's very easy to fix. We will need to inform EdgeImpulse of this, but providing help and support for the change to our users should be simple (removing detect() will be much harder given the need for a python callback).

(These commits are in the same PR because they build on each other).

Here's the new API:

```
import time
import tf
import image

model = 'yolo/lpd-yolov5-int8-quantized.tflite'

# Alternatively, models can be loaded from the filesystem storage.
net = tf.load(model, load_to_fb=True)
labels = ["plate"]
print(net)

colors = [  # Add more colors if you are detecting more than 7 types of classes at once.
    (255, 0, 0),
    (0, 255, 0),
    (255, 255, 0),
    (0, 0, 255),
    (255, 0, 255),
    (0, 255, 255),
    (255, 255, 255),
]

@micropython.native
def yolov5(model, rect):
    oh, ow, oc = model.output_shape
    if oh != 1:
        raise(ValueError("Expected model output height to be 1!"))
    if oc < 6:
        raise(ValueError("Expected model output channels to be >= 6"))
    # cx, cy, cw, ch, score, classes[n]
    ol = oh * ow * oc
    ih, iw, ic = model.input_shape
    nms = tf.NMS(iw, ih, rect)
    m = time.ticks_ms()
    for i in range(0, ol, oc):
        score = model[i + 4]
        if (score > 0.5):
            cx = model[i + 0]
            cy = model[i + 1]
            cw = model[i + 2] * 0.5
            ch = model[i + 3] * 0.5
            xmin = (cx - cw) * iw
            ymin = (cy - ch) * ih
            xmax = (cx + cw) * iw
            ymax = (cy + ch) * ih
            labels = model[i + 5: i + oc]
            label_index = max(enumerate(labels), key=lambda x: x[1])[0]
            nms.add_bounding_box(xmin, ymin, xmax, ymax, score, label_index)
    boxes = nms.get_bounding_boxes()
    print("time", time.ticks_diff(time.ticks_ms(), m))
    return boxes

clock = time.clock()
while True:
    clock.tick()

    img = image.Image("yolo/plate.jpg", copy_to_fb=True)
    img.to_rgb565()

    for i, detection_list in enumerate(
        net.predict(img, callback=yolov5, scale=tf.SCALE_NONE)
    ):
        print("********** %s **********" % labels[i])
        for rect, score in detection_list:
            print(rect, "score", score)
            img.draw_rectangle(rect, color=colors[i], thickness=2)

    print(clock.fps(), "fps", end="\n")
```

Since I removed the need for the output object now things can be accessed in more pythonic way. I like how things are improved.

### DIFF
--- a/scripts/examples/03-Machine-Learning/00-TensorFlow/tf_object_detection.py
+++ b/scripts/examples/03-Machine-Learning/00-TensorFlow/tf_object_detection.py
@@ -14,10 +14,10 @@ import math
 sensor.reset()  # Reset and initialize the sensor.
 sensor.set_pixformat(sensor.RGB565)  # Set pixel format to RGB565 (or GRAYSCALE)
 sensor.set_framesize(sensor.QVGA)  # Set frame size to QVGA (320x240)
-sensor.set_windowing((240, 240))  # Set 240x240 window.
 sensor.skip_frames(time=2000)  # Let the camera adjust.
 
 min_confidence = 0.4
+threshold_list = [(math.ceil(min_confidence * 255), 255)]
 
 # Load built-in FOMO face detection model
 labels, net = tf.Model("fomo_face_detection")
@@ -36,18 +36,38 @@ colors = [  # Add more colors if you are detecting more than 7 types of classes 
     (255, 255, 255),
 ]
 
+# FOMO outputs an image per class where each pixel in the image is the centroid of the trained
+# object. So, we will get those output images and then run find_blobs() on them to extract the
+# centroids. We will also run get_stats() on the detected blobs to determine their score.
+# The Non-Max-Supression (NMS) object then filters out overlapping detections and maps their
+# position in the output image back to the original input image. The callback then returns a
+# list per class which each contain a list of (rect, score) tuples representing the detected
+# objects.
+
+
+def fomo_callback(model, rect):
+    out = model.output[0]
+    oh, ow, oc = model.output_shape
+    nms = tf.NMS(ow, oh, rect)
+    for i in range(oc):
+        img = out.get_image(i)
+        blobs = img.find_blobs(threshold_list, x_stride=1, area_threshold=1, pixels_threshold=1)
+        for b in blobs:
+            rect = b.rect()
+            x, y, w, h = rect
+            score = img.get_statistics(thresholds=threshold_list, roi=rect).l_mean() / 255.0
+            nms.add_bounding_box(x, y, x + w, y + h, score, i)
+    return nms.get_bounding_boxes()
+
+
 clock = time.clock()
 while True:
     clock.tick()
 
     img = sensor.snapshot()
 
-    # detect() returns all objects found in the image (splitted out per class already)
-    # we skip class index 0, as that is the background, and then draw circles of the center
-    # of our objects
-
     for i, detection_list in enumerate(
-        net.detect(img, thresholds=[(math.ceil(min_confidence * 255), 255)])
+        net.predict(img, callback=fomo_callback)
     ):
         if i == 0:
             continue  # background class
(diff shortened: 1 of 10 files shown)


kwagyeman:
Get the model, and test image here: https://github.com/openmv/openmv/pull/2227

iabdalkader:
Hi,

* First 2 commits are okay 👍🏼 but I already merged the first commit, so you should drop it.
* We should move NMS to its own file, and then either add it to the image module (`nms = image.NMS()`) ? Or to the tf module (`nms = tf.NMS()`) either way is fine with me.
* I may have misspoken about the output object. By merge I actually meant just make the output object an attribute of the model, and then remove the duplicated things like params. So in code this would look like:
```Python
def model_callback(model, rect):
    model.output[0]
```
Initially and before running inference, the output should be `mp_const_none`.

* Please use MP_OBJ_TO_PTR and MP_OBJ_FROM_PTR everywhere you cast objects to pointers and back.

Note that we need to wrap up this PR pretty quickly because it's blocking all my work.

src/omv/modules/py_tf.c
```diff
@@ -120,11 +120,84 @@ STATIC mp_obj_t py_tf_model_output_subscr(mp_obj_t self_in, mp_obj_t index, mp_o
     return MP_OBJ_NULL; // op not supported
 }
 
+STATIC mp_obj_t py_tf_model_output_get_image(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
```
iabdalkader, line 125: This should be a generic image module function, in py_image.c, passed the raw output, scale etc..

iabdalkader:
@kwagyeman I moved this to  #2235 after dropping the last 3 commits. For the regression example, we'll add it back but there will be a place for the models in the repo (instead of linking to shopify content). As for ulab array support in regression, we can support ulab or plain arrays but not both.

scripts/examples/03-Machine-Learning/00-TensorFlow/tf_object_detection.py
```diff
@@ -36,30 +36,49 @@
...
+    nms = tf.NMS(ow, oh, rect)
+    for i in range(oc):
+        img = out.get_image(i)
+        blobs = img.find_blobs(threshold_list, x_stride=1, area_threshold=1, pixels_threshold=1)
+        for b in blobs:
+            rect = b.rect()
+            x, y, w, h = rect
+            score = img.get_statistics(thresholds=threshold_list, roi=rect).l_mean() / 255.0
+            nms.add_bounding_box(x, y, x + w, y + h, score, i)
+    return nms.get_bounding_boxes()
```
iabdalkader, line 60: Should consider moving this post processing function to an extension module. The same for yolo, if we ever add an example.

kwagyeman: Yeah, I think edge impulse will need to generate that since FOMO is there model. This will serve as a template code for them to provide.

For yolo and other models... for the next gen system push we're probably going to want to pick a library of models we can leverage and then create an extension module with all the post processing callbacks.

src/omv/modules/py_tf.c
kwagyeman: We should extend Image() to accept a list[] so you can create an image from that. Or, have like a image.from_list() function in the image module.

iabdalkader: @kwagyeman Could implement that change ? I want to test the output callback more and that would make it a lot easier for me to do so. But please keep changes to py_tf to absolute minimum as I already have stashed changes for that file.

kwagyeman: Yeah, it would be in py_image.c so, no changes to py_tf.

PR #2239, opened 2024-06-23
TensorFlow: Switch to upstream TFLM and refactor all TensorFlow modules.

iabdalkader:
This PR mainly replaces the few years old TensorFlow library with newer, faster, and smaller libraries built directly from TFLM upstream (see https://github.com/openmv/tflm/). It also includes a few more improvements, and necessary changes:

* The MicroPython TF module is decoupled from the TensorFlow library, allowing support for more DL/ML libraries and engines in the future. The ML backend has been completely redesigned; the model object can now be passed directly to the backend, allowing it to initialize the model internally.
*  The ML backend's state/memory is now persistent (surviving across invocations). This improves the inference speed by around 20% and allows support for models that require persistent memory, such as LSTM.
* The `tf` (now called `ml`) module has also been mostly rewritten to handle model input/output shapes and data properly, and to support models with multiple outputs.
* The built-in model generator has been updated to allow fine-grained control over which models get built into the firmware image. Models can be embed conditionally using the index file. Models listed in `models/index.txt` will be generated with an enable macro, which can be defined per board in `imlib_config.h` files. Other models in models/ will be built-in by default.This PR enables FOMO for all boards, and enables the audio processing models for boards with onboard MICs.
* Linker scripts have been updated to support more GC blocks, and boards' memory configurations have been optimized to maximize the available GC heap memory for use for the backend's persistent state.
* The obsolete micro-speech module no longer works with the updated TFLM, since feature generation has been removed and replaced by another audio preprocessor model. This module has been replaced with a pure MicroPython Micro-Speech implementation.
* Finally, the audio module now supports setting the number of samples dynamically at runtime. This makes Micro Speech's implementation easier, because it require a very specific number of samples.


iabdalkader:

https://forums.openmv.io/t/py-tf-update-in-development-release-firmware/9741/4

PR #2243, opened 2024-07-01
modules/py_ml: Add support for multi-input tensor models.

kwagyeman:
[modules/py_ml: Add support for multi-input tensor models.](https://github.com/openmv/openmv/commit/3934c323b7444330de8353ed6aedaf9abbdc3b91)

This commit adds support for handling multiple input tensors in models. predict() now takes an array of objects equal to the number of input tensors it has. The input objects can either be ndarrays or ml.Image() objects.

The new ml.Image() object is a wrapper for images and their arguments that you would have passes previously to ml.predict(). By moving these to an object you can now pass different images with different arguments if you needed to.

Finally, the callback option has been simplified so that the callback is just passed the (model, inputs, and outputs). Where `inputs` is what was passed to predict.

...

Per our meeting, here's the proposed API with this PR:

```
def cb(model, inputs, outputs):
    some_ndarray1, ml_image, some_ndarray2 = inputs
    # and outputs is the list of output tensors
    roi = ml_image.roi
    ...

ml.predict([some_ndarray1, ml.Image(img, roi=(x, y, w, h), scale=ml.SCALE_NONE), some_ndarray2], callback=cb)
```

Arguments for the image are stored with it in the array and then are available with it in the callback.

The alternative API as discussed:

```
def cb(model, outputs, rects):
    # and outputs is the list of output tensors
    roi = rects[0]
    ...

ml.predict([some_ndarray1, img, some_ndarray2], roi=[(x, y, w, h)], scaling=[ml.SCALE_NONE], callback=cb)
# where the values in the arrays for roi/scaling are applied per image discovered in the input list.
```

The scaling value is unlikely to be needed in the callback. So, just rects can be supplied. The alternative API could also pass a global scaling value for all images in the Model() constructor versus having it in the predict call. I think we want to avoid a global value, though, for everything.

Either API above should future proof us for a few years. I think my suggestion is better designed though as all arguments are in one place and you have access to them in the callback. I agree, though it's an extra object. The alternative API is very simple comparatively and gives us just what we need and not extra. But, it may be less future-proof.

...

[modules/py_ml: Fully vectorize all input and output scaling/datatypes.](https://github.com/openmv/openmv/commit/89345830f84066773413d761412af2ec8224934b)

Fully vectorizes the scaling, zero-point, and datatype values. We should do this now so that if it does become the case in the future that a model has differently scaled tensor heads we have support already complete and don't have to break the API again. Given the backend can change and we may not necessarily always be running tensorflow models our code should handle each tensor being different.

[modules/py_ml_nms: Fix tf->ml naming.](https://github.com/openmv/openmv/commit/583c826e8ca836d7ac860e97cc575b51bb310e25)

This commit fixes a bug with `ml` module renaming of things.

[lib/tflm: Code cleanup.](https://github.com/openmv/openmv/commit/f9ab07cef14ac3e98fd4004cad3c929c5d09ef1b)

Minor code cleanup in the backend.

[modules/py_ml: Add support for arrays.](https://github.com/openmv/openmv/pull/2243/commits/1e6b2e693d9f611170a7c2b8fd77bc42a4e13d30)

This commit adds support for arrays. We should have this so that we have a backdoor to pass input to a model that needs some high-dimensional vector support. The code is much simpler now and just accepts a flattened list, which has to be the same length as the tensor shape. It looks roughly the same as the array code but doesn't have the shape-matching enforcement.

### DIFF
--- a/scripts/examples/03-Machine-Learning/00-TensorFlow/tf_image_classification.py
+++ b/scripts/examples/03-Machine-Learning/00-TensorFlow/tf_image_classification.py
@@ -43,7 +43,7 @@ while True:
     # This combines the labels and confidence values into a list of tuples
     # and then sorts that list by the confidence values.
     sorted_list = sorted(
-        zip(labels, model.predict(img)[0]), key=lambda x: x[1], reverse=True
+        zip(labels, model.predict([ml.Image(img)])[0]), key=lambda x: x[1], reverse=True
     )
     for i in range(5):
         print("%s = %f" % (sorted_list[i][0], sorted_list[i][1]))
--- a/scripts/examples/03-Machine-Learning/00-TensorFlow/tf_object_detection.py
+++ b/scripts/examples/03-Machine-Learning/00-TensorFlow/tf_object_detection.py
@@ -45,9 +45,9 @@ colors = [  # Add more colors if you are detecting more than 7 types of classes 
 # position in the output image back to the original input image. The function then returns a
 # list per class which each contain a list of (rect, score) tuples representing the detected
 # objects.
-def fomo_post_process(model, output, rect):
+def fomo_post_process(model, input, output):
     n, oh, ow, oc = model.output_shape[0]
-    nms = ml.NMS(ow, oh, rect)
+    nms = ml.NMS(ow, oh, input[0].roi)
     for i in range(oc):
         img = image.Image(output[0], shape=(oh, ow, 1), strides=(i, oc), scale=(255, 0))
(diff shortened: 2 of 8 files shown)


kwagyeman:
I can't figure out the fix for getting it to build on the IMXRT.

`#define MICROPY_FLOAT_IMPL MICROPY_FLOAT_IMPL_FLOAT`

Seems to fix the backend file not finding the right floating point object to use. However, this isn't a fix. Not sure where this is getting dropped in the build.

PR #2247, opened 2024-07-04
ports/stm32: Add mic gain support to py_audio when using dfsdm.

kwagyeman:
Gain support works now. It doesn't seem to really improve the ability of the mic to listen. I still have to say "Yes" and "No" at the Nicla with a commanding voice. I do not need to be right next to it, though.


iabdalkader:
> . I still have to say "Yes" and "No" at the Nicla with a commanding voice.

I noticed that too. It's likely the training data, it might be trained on synthesized voices.

PR #2258, opened 2024-07-08
scripts/libraries: Update normalization to handle 3D/4D image tensors.

kwagyeman:
Tested with FOMO and YOLO.

### DIFF
--- a/scripts/libraries/ml/ml/preprocessing.py
+++ b/scripts/libraries/ml/ml/preprocessing.py
@@ -36,7 +36,7 @@ class Normalization:
         buffer, shape, dtype = args
         # Create an image using the input tensor as buffer.
         if len(shape) != 4:
-            raise ValueError("Expected 4D input tensor (batches, height, width, channels)")
+            raise ValueError("Expected input tensor with shape: (1, H, W, C)")
         b, h, w, c = shape
         if b != 1:
             raise ValueError("Expected batches to be 1")
--- a/scripts/libraries/ml/ml/preprocessing.py
+++ b/scripts/libraries/ml/ml/preprocessing.py
@@ -35,8 +35,15 @@ class Normalization:
 
         buffer, shape, dtype = args
         # Create an image using the input tensor as buffer.
-        pixfmt = image.GRAYSCALE if shape[3] == 1 else image.RGB565
-        img = image.Image(shape[2], shape[1], pixfmt, buffer=buffer)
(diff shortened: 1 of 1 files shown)


kwagyeman:
TODO: PR should be updated to just check if the channels are 1/3 and batch size is 1.

iabdalkader:
This might be stay as it is if the batch size removal is not merged.

scripts/libraries/ml/ml/preprocessing.py
```diff
@@ -35,8 +35,18 @@ def __call__(self, *args):
...
-        pixfmt = image.GRAYSCALE if shape[3] == 1 else image.RGB565
-        img = image.Image(shape[2], shape[1], pixfmt, buffer=buffer)
+        if len(shape) == 3:
+            h, w, c = shape
+        elif len(shape) == 4:
+            b, h, w, c = shape
+            if b != 1:
+                raise ValueError("Expected 1 image per batch")
+        else:
+            raise ValueError("Expected 3D or 4D input tensor ([batch,] height, width, channels)")
```
iabdalkader, line 45:
```suggestion
        if len(shape) != 4 or shape[0] != 1:
                raise ValueError("Expected ([batch,] height, width, channels)")
        b, h, w, c = shape
```
This for sure will always be `b, h, w, c`

iabdalkader (changes requested):

scripts/libraries/ml/ml/preprocessing.py
kwagyeman: Yes, will reduce it to just sanity checks.

scripts/libraries/ml/ml/preprocessing.py
```diff
@@ -35,8 +35,15 @@ def __call__(self, *args):
 
         buffer, shape, dtype = args
         # Create an image using the input tensor as buffer.
-        pixfmt = image.GRAYSCALE if shape[3] == 1 else image.RGB565
-        img = image.Image(shape[2], shape[1], pixfmt, buffer=buffer)
+        if len(shape) != 4:
+            raise ValueError("Expected 4D input tensor (batches, height, width, channels)")
```
iabdalkader, line 39:
```suggestion
            raise ValueError("Expected input tensor with shape: (1, H, W, C)")
```

iabdalkader (approved):

PR #2262, opened 2024-07-09
scripts/examples: Fix light shield example.

### LINKED ISSUE #2240: Light shield on the RT1062 using pyb
For the RT1062 built-in examples, the light shield example is using pyb which is not supported by the RT1062.

**To Reproduce**
Steps to reproduce the beh
avior:
1. Connect the RT1062 board.
2. Click File->Examples->Light Shield->Light.py
3. Run script.
4. See error "ImportError: no module named 'pyb'"

The expected behavior is for the script to not use pyb and *possibly* use machine instead.

Script modifications will be needed to support the change.

 - IDE version: 4.1.5
 - Firmware version: 4.5.6
 - Board: RT1062
 - OS: Windows

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2240

I was not able to use the PWM module on the H7. It's missing from the machine module for all STM32 boards. So, I just made the code support both ways.

### DIFF
--- a/scripts/examples/50-OpenMV-Boards/60-Shields/68-Light-Shield/light.py
+++ b/scripts/examples/50-OpenMV-Boards/60-Shields/68-Light-Shield/light.py
@@ -1,13 +1,19 @@
 # This work is licensed under the MIT license.
-# Copyright (c) 2013-2023 OpenMV LLC. All rights reserved.
+# Copyright (c) 2013-2026 OpenMV LLC. All rights reserved.
 # https://github.com/openmv/openmv/blob/master/LICENSE
 #
+# This example shows off using the light shield with the Machine Module.
+
 import time
-from pyb import Pin, Timer
+from machine import PWM, Pin
 
 # 50kHz pin6 timer2 channel1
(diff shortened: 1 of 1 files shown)


iabdalkader:
Yeah missing `machine` classes is frustrating. Note that we have a `machine` extension, we can just wrap `timer` or `umachine.PWM` and then this script should work without changes. You'll need to implement all the PWM functions here though:
https://docs.micropython.org/en/latest/library/machine.PWM.html#machine.PWM

kwagyeman:
Good idea. I'll wrap the machine module and add the missing functionality.

kwagyeman:
@iabdalkader - Trying to finish this PR and clear this issue. However, the pyb module doesn't appear to support an easy way to map a Pin object back to a timer and channel. Any thoughts? Pin objects have an AF list I could parse. However, the wrapper will then end up being somewhat complex.

Maybe it's better just to split the example into an STM32 version and non-STM32 version? I can just make two light shield examples under the IMXRT and STM32 boards folder.

Or do we want this new smarter PWM module that can work on STM32 boards?

kwagyeman:
This is finally fixed now on STM32 and IMXRT.

PR #2270, opened 2024-07-10
modules/py_ml: ndarrays only need to match the input size if 1D.

kwagyeman:
1D ndarrays only need to match the input size. Their shape doesn't have to match.

### DIFF
--- a/src/omv/modules/py_ml.c
+++ b/src/omv/modules/py_ml.c
@@ -69,17 +69,21 @@ static void py_ml_process_input(py_ml_model_obj_t *model, mp_obj_t arg) {
             // Input is an ndarry. The input is converted and copied to the tensor buffer.
             ndarray_obj_t *input_array = MP_OBJ_TO_PTR(input_arg);
 
-            if (input_array->ndim != input_shape->len) {
-                mp_raise_msg(&mp_type_ValueError,
-                             MP_ERROR_TEXT("Input shape does not match the model input shape"));
-            }
-
-            for (size_t i = 0; i < input_array->ndim; i++) {
-                size_t ulab_offset = ULAB_MAX_DIMS - input_array->ndim;
-                if (input_array->shape[ulab_offset + i] != mp_obj_get_int(input_shape->items[i])) {
+            if (input_array->ndim > 1) {
+                if (input_array->ndim != input_shape->len) {
                     mp_raise_msg(&mp_type_ValueError,
                                  MP_ERROR_TEXT("Input shape does not match the model input shape"));
                 }
+
+                for (size_t i = 0; i < input_array->ndim; i++) {
+                    size_t ulab_offset = ULAB_MAX_DIMS - input_array->ndim;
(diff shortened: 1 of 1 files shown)


iabdalkader:
Why don't we just skip the batch size for everything ?

iabdalkader:
Meaning when loading input/output shapes, skip the first dimension ?

kwagyeman:
There's no way to know it's the batch size since it could be missing or present.

iabdalkader:
As far as I know, models always have a batch size, even if not specified. Can you find/create a model without a batch size ?

kwagyeman:
Will do a search again. When I previously did this, the answer was inconclusive. Will spend more time on it.

kwagyeman:
Okay, I spent some more time reading up about this.

The batch size is always the first dimension for any model. When creating a model, you can specify it or leave it unspecified and have the library choose it. It defines how many samples are processed at once... and the point is to leverage your GPU's parallel nature so training goes faster. However, the larger it is, the larger the GPU memory resources need to be. Training updates are only applied per Epoch, so the batch size doesn't affect the training results.

After training is finished, the batch size for inference is 1, as you want the model to work on one item at a time. If It were two or more, this would mean you'd be asking the model to process two or more pieces of data and run the same model on them in parallel. This makes sense for GPUs with lots of resources. For us, I don't see a need for a batch size larger than 1 as you can get the same result by running the the model serially over many pieces of data.

**TLDR; we should assume the first dimension is the batch size and it should be compared to be 1 as it always should be 1.** Note that we definitely could handle a model where it's larger than 1... we'd effectively just be leaving the other batch-size inputs as garbage and ignoring their outputs. However, the user should reduce their model to using a batch size of 1 to run onboard as running a model with a larger batch size would just result in a lot of extra wasted computation if only feed a single batch even though it was processing multiple batches.

iabdalkader:
> Thoughts? Will update this PR and #2258 afterwards.

Yeah we should remove the batch size. It's just like you said we can't waste a whole ndarray dimension, they're too expensive. The code in the backend should just skip the first dimension, start looping from the second one. I'll send it later.

As for scripts/API compatibility, I thought about this, and you'll always be safe if you use this: `ndarray.reshape(model.input_shape)`. Image are handled different anyway, and batch is ignored. 

One concern though, in general, does reshape create a deep copy ? And more specifically, if the dimensions of input and target are the same ?

> Note that we definitely could handle a model where it's larger than 1...

I don't think we should even bother with this, as it will be the same as calling the model in a loop from the top script. That said, we can store the input/output batch size in the model object *in the future* if it's needed.

kwagyeman:
Reshape creates deep copies if the array is sparse. Otherwise, it just makes a memory view of the array. This allows the original array to be accessed in one way and the view object to access it in a different way.

iabdalkader:
> Otherwise, it just makes a memory view of the array. 

Great, then we should always recommend a reshape using the model's input shape.

PR #2274, opened 2024-07-12
modules/py_ml: ML updates and fix.

iabdalkader:


kwagyeman (approved):
Only comment would be to add some magic in the python wrapper to look for a file with the same name as the model called "model.txt" and load all strings in it into labels it exists. But, this isn't really needed at all.

iabdalkader:
> Only comment would be to add some magic in the python wrapper to look for a file with the same name as the model called "model.txt" and load all strings in it into labels it exists when loading from disk. But, this isn't really needed at all.

I like this. You can reuse the empty `model.labels`. Please send it you want. I guess classification example would make use of it.

kwagyeman:
It can be in a separate PR. Merge this one and I'll send it.

PR #2276, opened 2024-07-12
scripts/examples: Add regression example.

kwagyeman:
Added to the RT1062 only as the model is 11KB so I don't want to waste that for the cameras which have limited flash remaining.


iabdalkader:
We don't have to add it to any cam, but I guess it makes testing easier.

Issue #2226, opened 2024-05-31
HIMAX Sensors not working with Arduino Giga


kwagyeman:
https://forums.openmv.io/t/failed-to-detect-hm0360-vga-monochrome-dvp-camera/9638/5

iabdalkader:
This issue has been mentioned on **OpenMV Forums**. There might be relevant details there:

https://forums.openmv.io/t/failed-to-detect-hm0360-vga-monochrome-dvp-camera/9638/10

kwagyeman:
Still waiting on Arduino to send the modules.

PR #2282, opened 2024-07-15
sensors/hm0360: Enable HM0360 to work on GIGA.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2226

The Arducam module doesn't configure the sensor to use it's internal clock.

### DIFF
--- a/src/omv/boards/ARDUINO_GIGA/omv_boardconfig.h
+++ b/src/omv/boards/ARDUINO_GIGA/omv_boardconfig.h
@@ -40,6 +40,8 @@
 
 #define OMV_HM01B0_ENABLE                   (1)
 #define OMV_HM0360_ENABLE                   (1)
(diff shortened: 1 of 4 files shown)


src/omv/boards/ARDUINO_GIGA/omv_boardconfig.h
```diff
@@ -40,6 +40,8 @@
 
 #define OMV_HM01B0_ENABLE                   (1)
 #define OMV_HM0360_ENABLE                   (1)
+#define OMV_HM0360_XCLK_FREQ                (24000000)
+#define OMV_HM0360_PLL1_CONFIG              (0x04)
```
kwagyeman, line 44: This is half what it was before because the internal OSC clock is 48MHz. Then, it was divided by 4. So, now we use a 24 MHz clock and divide it by 2.

iabdalkader (approved):

PR #2291, opened 2024-07-16
imlib: Fix unpack sign conversion.

kwagyeman:
YOLO and FOMO test networks both work now as expected.


kwagyeman:
Haha, we don't need this anymore. Can replace with numpy. Magic coming.

PR #2292, opened 2024-07-17
rpc: use long timeout when waiting for response data packet

sam-golioth:


kwagyeman:
Need a reason for the change. Those timeouts were picked very carefully.

sam-golioth:
> Need a reason for the change. Those timeouts were picked very carefully.

Oops! I meant to open this on the fork just to get a build for testing.

Since we're here, I'm running into an issue where in order to get reliable comms over the jumper wires I'm using, I need to drop the speed of the UART pretty low. That (plus some latency in the DMA driven UART driver I'm using) is leading to more than 2 ms between the RESULT_HEADER and RESULT_DATA packets.
Screenshot 2024-07-16 at 9 04 37 PM

kwagyeman:
You should be using a fast uart speed. I've clocked this library via the UART up to 7.5Mb/s on the H7. Not sure what the RT1062 goes up to. Running with a slow speed will cause issues like you are seeing when the packets take so long to send.

The long timeout is only used for the last transaction because that's the only one with a large data transfer. The rest are small packets.

If you can, I would just scale the:

```
self._put_short_timeout_reset = 2
self._get_short_timeout_reset = 2
```

Values instead to match the slower uart speed you are using. You should be able to access them from your rpc_slave instance class.

PR #2311, opened 2024-07-22
tools/tflite2c: Support converting Vela models on the fly.

iabdalkader:


tools/vela.ini
kwagyeman, line 1: Should this be in a config folder? Or, I guess it will be moved later?

iabdalkader: No it's where it needs to be. You can add more configs to it, and then pass the config name from port Makefile or board Makefile. For example:
```diff
diff --git a/src/omv/boards/OPENMV4P/omv_boardconfig.mk b/src/omv/boards/OPENMV4P/omv_boardconfig.mk
index 6fe2be3af..f7848053f 100755
--- a/src/omv/boards/OPENMV4P/omv_boardconfig.mk
+++ b/src/omv/boards/OPENMV4P/omv_boardconfig.mk
@@ -19,3 +19,4 @@ MICROPY_PY_WINC1500 = 1
 MICROPY_PY_DISPLAY = 1
 MICROPY_PY_TV = 1
 MICROPY_PY_BUZZER = 0
+VELA_ARGS="--system-config NAME_DTCM_MRAM --more-args ..."
```

Issue #2295, opened 2024-07-18
RT1062 Crashing running YOLO Network


kwagyeman:
Running this on the RT1062 with all the 4.5.6 PRs merged crashes. Doesn't have a problem on the H7 Plus and it works without issues. Maybe it's a floating point ABI issue our code and tensorflow on the RT1062? Related to the float implementation issue?

```
import time
import ml
import image
from ml.utils import NMS

model = 'yolo/lpd-yolov5-int8-quantized.tflite'

# Alternatively, models can be loaded from the filesystem storage.
net = ml.Model(model)
labels = ["plate"]
print(net)

colors = [  # Add more colors if you are detecting more than 7 types of classes at once.
    (255, 0, 0),
    (0, 255, 0),
    (255, 255, 0),
    (0, 0, 255),
    (255, 0, 255),
    (0, 255, 255),
    (255, 255, 255),
]

@micropython.native
def yolov5(model, input, output):
    out = output[0]
    ib, ih, iw, ic = model.input_shape[0]
    ob, ow, oc = model.output_shape[0]
    if ob != 1:
        raise(ValueError("Expected model output batch to be 1!"))
    if oc < 6:
        raise(ValueError("Expected model output channels to be >= 6"))
    # cx, cy, cw, ch, score, classes[n]
    ol = ob * ow * oc
    nms = NMS(iw, ih, input[0].roi)
    m = time.ticks_ms()
    for i in range(ow):
        score = out[0, i, 4]
        if (score > 0.5):
            cx = out[0, i, 0]
            cy = out[0, i, 1]
            cw = out[0, i, 2] * 0.5
            ch = out[0, i, 3] * 0.5
            xmin = (cx - cw) * iw
            ymin = (cy - ch) * ih
            xmax = (cx + cw) * iw
            ymax = (cy + ch) * ih
            labels = out[0, i, 5:]
            label_index = max(enumerate(labels), key=lambda x: x[1])[0]
            nms.add_bounding_box(xmin, ymin, xmax, ymax, score, label_index)
    boxes = nms.get_bounding_boxes()
    print("time", time.ticks_diff(time.ticks_ms(), m))
    return boxes

clock = time.clock()
while True:
    clock.tick()

    img = image.Image("yolo/plate.jpg", copy_to_fb=True)
    img.to_rgb565()

    for i, detection_list in enumerate(
        net.predict([img], callback=yolov5)
    ):
        print("********** %s **********" % labels[i])
        for rect, score in detection_list:
[...]

PR #2315, opened 2024-07-24
Fix JPEG decoder boundary check

bitbank2:
This fix prevents odd-sized JPEG images from drawing beyond the pixel boundaries of the allocated buffer. For example, a 19x16 pixel image with 4:2:0 subsampling spans 2 16x16 MCUs, but only 3 columns of the second MCU must be written to the output framebuffer.


kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2295

kwagyeman:
Moved to: https://github.com/openmv/openmv/pull/2325

Issue #2269, opened 2024-07-09
Cleanup draw.c and move hardware-accelerated code paths to ports.


iabdalkader:
* Separate the hardware accelerated functions from the software code in `draw.c`.
* Define a basic API for the hardware accelerated functions.
* Provide default software implementations as weak functions in `draw.c`.
* Move the DMA2D implementations to ports/stm32/draw_dma2d.c (this file is enabled if DMA2D macro is enabled).

PR #2328, opened 2024-07-28
imlib/draw: Refactor draw_image to support gpu offload.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2269

I removed all the ping-pong buffer stuff going on in draw_image. This simplified a lot of code. It probably wasn't really even helping really. Also, DMA2D is used in it's fastest mode when it can offload a whole image copy. So, this likely speeds things up.

The GPU accelerated method is invoked if draw_image thinks it can use it. All checks for dealing with scaling, boundaries, centering, etc. are all handled already. The GPU is just given what it needs to do. Then, if it cannot support the operation, it should return false, and the software code will do it.

...

Regarding accelerating drawing methods like draw_lines and etc. New GPU stuff can accelerate these, too. However, this should be tackled later in another PR. This PR sets up our code to gain the best offload a GPU can support which is the giant image copy/scale operation.

### DIFF
--- a/src/omv/common/omv_gpu.h
+++ b/src/omv/common/omv_gpu.h
@@ -1,0 +1,39 @@
+/*
+ * This file is part of the OpenMV project.
+ *
+ * Copyright (c) 2013-2024 Ibrahim Abdelkader <iabdalkader@openmv.io>
+ * Copyright (c) 2013-2024 Kwabena W. Agyeman <kwagyeman@openmv.io>
+ *
+ * This work is licensed under the MIT license, see the file LICENSE for details.
+ *
+ * GPU driver.
+ */
+#ifndef __OMV_GPU_H__
+#define __OMV_GPU_H__
+#include "imlib.h"
+
+int omv_gpu_init();
+int omv_gpu_deinit();
+
+// Draws src_rect from src_img to dst_rect in dst_img.
+// If the sizes of src_rect and dst_rect are different then the image must be scaled.
+// Alpha is the alpha value (0-256) to use when blending the source image with the destination image.
+// If color_palette is not NULL, it is used to map the source image's 8-bit Y to an 16-bit RGB565 color.
+// If alpha_palette is not NULL, it is used to map the source image's 8-bit Y to an 8-bit alpha value.
+//
+// Valid Hint Flags:
+//
+// * IMAGE_HINT_BILINEAR: Use 2D bilinear interpolation when scaling the image versus nearest-neighbor.
+// * IMAGE_HINT_HMIRROR: The destination image result should be horizontally mirrored from the source image.
+// * IMAGE_HINT_VFLIP: The destination image result should be vertically flipped from the source image.
+// * IMAGE_HINT_BLACK_BACKGROUND: The destination image should be considered to be black (0) for alpha blending.
+
+int omv_gpu_draw_image(image_t *src_img,
+                       rectangle_t *src_rect,
+                       image_t *dst_img,
+                       rectangle_t *dst_rect,
+                       int alpha,
+                       const uint16_t *color_palette,
+                       const uint8_t *alpha_palette,
+                       image_hint_t hint);
+#endif // __OMV_GPU_H__
--- a/src/omv/common/omv_gpu.h
+++ b/src/omv/common/omv_gpu.h
@@ -13,7 +13,7 @@
 #include "imlib.h"
 
 int omv_gpu_init();
-int omv_gpu_deinit();
+void omv_gpu_deinit();
 
 // Draws src_rect from src_img to dst_rect in dst_img.
 // If the sizes of src_rect and dst_rect are different then the image must be scaled.
--- a/src/omv/imlib/draw.c
+++ b/src/omv/imlib/draw.c
@@ -10,13 +10,8 @@
  */
 #include "font.h"
 #include "imlib.h"
+#include "omv_gpu.h"
 #include "unaligned_memcpy.h"
-
-#ifdef IMLIB_ENABLE_DMA2D
-#include STM32_HAL_H
-#include "dma.h"
-#include "omv_common.h"
-#endif
 
 void *imlib_compute_row_ptr(const image_t *img, int y) {
     switch (img->pixfmt) {
@@ -711,133 +706,7 @@ void imlib_draw_row_setup(imlib_draw_row_data_t *data) {
     // but with the bpp of the source image.
     size_t image_row_size = image_size(&temp) / data->dst_img->h;
 
-    data->toggle = 0;
-    data->row_buffer[0] = fb_alloc(image_row_size, FB_ALLOC_CACHE_ALIGN);
-
-    #ifdef IMLIB_ENABLE_DMA2D
-    data->dma2d_enabled = false;
-    data->dma2d_initialized = false;
-
-    void *dst_buff = data->dst_row_override ? data->dst_row_override : data->dst_img->data;
-
-    if (data->dma2d_request && (data->dst_img->pixfmt == PIXFORMAT_RGB565) && DMA_BUFFER(dst_buff) &&
-        ((data->src_img_pixfmt == PIXFORMAT_GRAYSCALE) ||
-         ((data->src_img_pixfmt == PIXFORMAT_RGB565) && (data->rgb_channel < 0)
-          && (data->alpha != 256) && (!data->color_palette) && (!data->alpha_palette)))) {
-        data->row_buffer[1] = fb_alloc(image_row_size, FB_ALLOC_CACHE_ALIGN);
-        data->dma2d_enabled = true;
-        data->dma2d_initialized = true;
-
-        memset(&data->dma2d, 0, sizeof(data->dma2d));
-
-        data->dma2d.Instance = DMA2D;
-        data->dma2d.Init.Mode = DMA2D_M2M;
-        if (data->dst_img->pixfmt != data->src_img_pixfmt) {
-            data->dma2d.Init.Mode = DMA2D_M2M_PFC;
-        }
-        if ((data->alpha != 256) || data->alpha_palette) {
-            data->dma2d.Init.Mode = DMA2D_M2M_BLEND;
-        }
-        data->dma2d.Init.ColorMode = DMA2D_OUTPUT_RGB565;
-        data->dma2d.Init.OutputOffset = 0;
-        #if defined(MCU_SERIES_F7) || defined(MCU_SERIES_H7)
-        data->dma2d.Init.AlphaInverted = DMA2D_REGULAR_ALPHA;
-        data->dma2d.Init.RedBlueSwap = DMA2D_RB_REGULAR;
-        #endif
-        HAL_DMA2D_Init(&data->dma2d);
-
-        data->dma2d.LayerCfg[0].InputOffset = 0;
-        data->dma2d.LayerCfg[0].InputColorMode = DMA2D_INPUT_RGB565;
-        data->dma2d.LayerCfg[0].AlphaMode = DMA2D_REPLACE_ALPHA;
-        data->dma2d.LayerCfg[0].InputAlpha = data->black_background ? 0x00 : 0xff;
-        #if defined(MCU_SERIES_F7) || defined(MCU_SERIES_H7)
-        data->dma2d.LayerCfg[0].AlphaInverted = DMA2D_REGULAR_ALPHA;
-        data->dma2d.LayerCfg[0].RedBlueSwap = DMA2D_RB_REGULAR;
-        #endif
-        #if defined(MCU_SERIES_H7)
-        data->dma2d.LayerCfg[0].ChromaSubSampling = DMA2D_NO_CSS;
-        #endif
-        HAL_DMA2D_ConfigLayer(&data->dma2d, 0);
-
-        switch (data->src_img_pixfmt) {
-            case PIXFORMAT_GRAYSCALE: {
-                data->dma2d.LayerCfg[1].InputColorMode = DMA2D_INPUT_L8;
-                data->dma2d.LayerCfg[1].AlphaMode = DMA2D_COMBINE_ALPHA;
-                uint32_t *clut = fb_alloc(256 * sizeof(uint32_t), FB_ALLOC_CACHE_ALIGN);
-
-                if (!data->alpha_palette) {
-                    if (!data->color_palette) {
-                        for (int i = 0; i < 256; i++) {
-                            clut[i] = (0xff << 24) | COLOR_Y_TO_RGB888(i);
-                        }
-                    } else {
-                        for (int i = 0; i < 256; i++) {
-                            int pixel = data->color_palette[i];
-                            clut[i] =
-                                (0xff <<
-                                    24) |
-                                (COLOR_RGB565_TO_R8(pixel) << 16) | (COLOR_RGB565_TO_G8(pixel) << 8) | COLOR_RGB565_TO_B8(
-                                    pixel);
-                        }
-                    }
-                } else {
-                    if (!data->color_palette) {
-                        for (int i = 0; i < 256; i++) {
-                            clut[i] = (data->alpha_palette[i] << 24) | COLOR_Y_TO_RGB888(i);
-                        }
-                    } else {
-                        for (int i = 0; i < 256; i++) {
-                            int pixel = data->color_palette[i];
-                            clut[i] =
-                                (data->alpha_palette[i] <<
-                                    24) |
-                                (COLOR_RGB565_TO_R8(pixel) << 16) | (COLOR_RGB565_TO_G8(pixel) << 8) | COLOR_RGB565_TO_B8(
-                                    pixel);
-                        }
-                    }
-                }
-
-                DMA2D_CLUTCfgTypeDef cfg;
-                cfg.pCLUT = clut;
-                cfg.CLUTColorMode = DMA2D_CCM_ARGB8888;
-                cfg.Size = 255;
-                #if __DCACHE_PRESENT
-                SCB_CleanDCache_by_Addr(clut, 256 * sizeof(uint32_t));
-                #endif
-                HAL_DMA2D_CLUTLoad(&data->dma2d, cfg, 1);
-                HAL_DMA2D_PollForTransfer(&data->dma2d, 1000);
-                break;
-            }
-            case PIXFORMAT_RGB565: {
-                data->dma2d.LayerCfg[1].InputColorMode = DMA2D_INPUT_RGB565;
-                data->dma2d.LayerCfg[1].AlphaMode = DMA2D_REPLACE_ALPHA;
-                break;
-            }
-            default: {
-                break;
-            }
-        }
-
-        data->dma2d.LayerCfg[1].InputOffset = 0;
-        data->dma2d.LayerCfg[1].InputAlpha = fast_roundf((data->alpha * 255) / 256.f);
-        #if defined(MCU_SERIES_F7) || defined(MCU_SERIES_H7)
-        data->dma2d.LayerCfg[1].AlphaInverted = DMA2D_REGULAR_ALPHA;
-        data->dma2d.LayerCfg[1].RedBlueSwap = DMA2D_RB_REGULAR;
-        #endif
-        #if defined(MCU_SERIES_H7)
-        data->dma2d.LayerCfg[1].ChromaSubSampling = DMA2D_NO_CSS;
-        #endif
-        HAL_DMA2D_ConfigLayer(&data->dma2d, 1);
-        #if __DCACHE_PRESENT
-        // SCB_InvalidateDCache_by_Addr does nothing if dsize is 0.
-        data->dma2d_invalidate_dsize = 0;
-        #endif
-    } else {
-        data->row_buffer[1] = data->row_buffer[0];
-    }
-    #else
-    data->row_buffer[1] = data->row_buffer[0];
-    #endif
+    data->row_buffer = fb_alloc(image_row_size, FB_ALLOC_CACHE_ALIGN);
 
     int alpha = data->alpha, max = 256;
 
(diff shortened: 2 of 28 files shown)


iabdalkader:
This is really good, thanks! It's exactly what I need to implement hooks for the GPU. I'll post a review.

src/omv/imlib/imlib.h
```diff
@@ -1136,24 +1136,33 @@ typedef struct imlib_draw_row_data {
...
-    #endif
-    #endif
+    void *row_buffer; // private
     long smuad_alpha; // private
     uint32_t *smuad_alpha_palette; // private
 } imlib_draw_row_data_t;
 
 typedef void (*imlib_draw_row_callback_t) (int x_start, int x_end, int y_row, imlib_draw_row_data_t *data);
 
+typedef struct imlib_draw_gpu_data {
```
iabdalkader, line 1146: * I think all of the GPU functions should be moved to `omv/common/omv_gpu.h` and named accordingly.
* Do we need `dst_x_end`, `dst_y_end`, `src_x_end`, `src_x_end` ? Aren't those deducible from x,y and `w` and `h` ? If not then pass as rectangles.
* Should use OR'd flags. `OMV_GPU_BILINEAR`, `OMV_GPU_HMIRROR` etc.. The good news is I can implement most of those. Probably also flags to use color/alpha ?
* What should black_background do ?
* LUTs will need to be portable. For example I need them to be `unsigned long`. They'll need to be defined in `omv_portconfig.h` as `omv_clut_t`, we can then update the generated LUTs to use that.

The `omv_gpu.h` should look like:

```C
int omv_gpu_draw_image(image_t *src,
                       point_t src_xy,
                       image_t *dst,
                       point_t dst_xy,
                       const omv_clut_t *alpha_palette,
                       const omv_clut_t *color_palette,
                       uint32_t int alpha,
                       uint32_t flags);
```

src/omv/imlib/imlib.h
```diff
@@ -1316,12 +1325,12 @@ void imlib_find_hog(image_t *src, rectangle_t *roi, int cell_size);
...
-#ifdef IMLIB_ENABLE_DMA2D
-void imlib_draw_row_deinit_all();
+#if (OMV_GPU_ENABLE == 1)
+void imlib_hardware_gpu_init();
+void imlib_hardware_gpu_deinit();
 #endif
-void *imlib_draw_row_get_row_buffer(imlib_draw_row_data_t *data);
-void imlib_draw_row_put_row_buffer(imlib_draw_row_data_t *data, void *row_buffer);
 void imlib_draw_row(int x_start, int x_end, int y_row, imlib_draw_row_data_t *data);
+bool imlib_draw_image_gpu(imlib_draw_gpu_data_t *data);
```
iabdalkader, line 1333: These functions shouldn't be here.

src/omv/ports/stm32/omv_gpu.c
```diff
@@ -0,0 +1,223 @@
+/*
+ * This file is part of the OpenMV project.
+ *
+ * Copyright (c) 2013-2024 Ibrahim Abdelkader <iabdalkader@openmv.io>
+ * Copyright (c) 2013-2024 Kwabena W. Agyeman <kwagyeman@openmv.io>
+ *
+ * This work is licensed under the MIT license, see the file LICENSE for details.
+ *
+ * Hardware Accelerated Graphics Processing.
```
iabdalkader, line 9:
```suggestion
 * GPU driver for STM32 port.
```

src/omv/ports/stm32/omv_gpu.c
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    #if (TIME_GPU == 1)
+    printf("draw time: %u ms\n", mp_hal_ticks_ms() - start);
+    #endif
+
+    return true;
+}
+
+void imlib_hardware_gpu_init() {
```
iabdalkader, line 215: You could save some time if you init/deinit once on soft-reboot loop (if possible). Other than that, these should be.

```C
omv_gpu_init() {
// init should be called on soft-reboot
}

omv_gpu_deinit() {
// deinit should be called on soft-reboot
}

omv_gpu_draw_image() {
}
```

src/omv/ports/stm32/omv_portconfig.mk
```diff
@@ -623,6 +623,7 @@ UVC_OBJ += $(addprefix $(BUILD)/$(OMV_DIR)/imlib/,\
 	)
 
 UVC_OBJ += $(addprefix $(BUILD)/$(OMV_DIR)/ports/stm32/,\
+	omv_gpu.o                               \
```
iabdalkader, line 626:
```suggestion
	omv_gpu.o                               \
```
At the end not here.

iabdalkader (changes requested):

iabdalkader:
I have no other comments, great work! We can probably get this merged today then back on track.

src/omv/imlib/imlib.h
kwagyeman: > Do we need dst_x_end, dst_y_end, src_x_end, src_x_end ? Aren't those deducible from x,y and w and h ? If not then pass as rectangles.

Yeah, they can be rects. I didn't do that though since it produces an awkward amount of packing and unpacking.

> Actually no please pass the flags as those are useful. For example, I need to know if you want to scale or crop (crop when called from the sensor for set_window). The good news is I can support most of those flags, if not all of them.

Draw_image takes care of them already. The rects that are passed have already had all checks done on them. All the scaling and centring logic has been completely handled. I tried to write the logic so you could just directly pass the rects to the GPU blit method given it's docs.

> What should black_background do ?

In DMA2D is just sets the alpha for the background to 0 so that it acts as a black surface when blended.

> LUTs will need to be portable. For example I need them to be unsigned long. They'll need to be defined in omv_portconfig.h as omv_clut_t, we can then update the generated LUTs to use that.

The CLUTs are used directly. Code loads them into the hardware module. So, I don't think you need to make them portable?

> One more issue, it seems I'm allowed to do scaling in place, src==dst, it just works, so I don't need to allocate memory. Can we some how take advantage of that such that ? Maybe let gpu_draw_image allocate memory if it needs to ?

Yeah, I'll put some more cut-through paths in the draw image.

src/omv/imlib/imlib.h
kwagyeman: Okay

src/omv/ports/stm32/omv_portconfig.mk
kwagyeman: Done

src/omv/ports/stm32/omv_gpu.c
kwagyeman: Done

src/omv/ports/stm32/omv_gpu.c
kwagyeman: Done

src/omv/imlib/imlib.h
iabdalkader: * Yeah, they can be rects. I didn't do that though since it produces an awkward amount of packing and unpacking.

I think they'll just get passed on the stack, nothing extra, but will allow us to remove the extra gpu_data struct and just a call a function. Do we need the x/y ends ? Can't it infer those from the x/y start and w/h ?

> Draw_image takes care of them already

Understood, but I still need the flags because I need to know if you want to crop or scale. There's no way to know that other than passing yet another `bool`. At this point passing the flags is just easier, also covers future stuff.

> In DMA2D is just sets the alpha for the background to 0 so that it acts as a black surface when blended.

Can it be just an alpha arg then ?

> The CLUTs are used directly. Code loads them into the hardware module. So, I don't think you need to make them portable?

As I've said already I must load the CLUTs as ints, 32 bits per entry. It can't be uin8_t or 16 it must be 32 bits. Therefore it should be configured by each port.

kwagyeman: > As I've said already I must load the CLUTs as ints, 32 bits per entry. It can't be uin8_t or 16 it must be 32 bits. Therefore it should be configured by each port.

Yeah, so, for DMA2D there's a for loop that translates them into what the GPU needs:

```
if (!data->alpha_palette) {
            if (!data->color_palette) {
                for (int i = 0; i < 256; i++) {
                    clut[i] = (0xff << 24) | COLOR_Y_TO_RGB888(i);
                }
            } else {
                for (int i = 0; i < 256; i++) {
                    int pixel = data->color_palette[i];
                    clut[i] = (0xff << 24) |
                              (COLOR_RGB565_TO_R8(pixel) << 16) |
                              (COLOR_RGB565_TO_G8(pixel) << 8) |
                              COLOR_RGB565_TO_B8(pixel);
                }
            }
        } else {
            if (!data->color_palette) {
                for (int i = 0; i < 256; i++) {
                    clut[i] = (data->alpha_palette[i] << 24) | COLOR_Y_TO_RGB888(i);
                }
            } else {
                for (int i = 0; i < 256; i++) {
                    int pixel = data->color_palette[i];
                    clut[i] = (data->alpha_palette[i] << 24) |
                              (COLOR_RGB565_TO_R8(pixel) << 16) |
                              (COLOR_RGB565_TO_G8(pixel) << 8) |
                              COLOR_RGB565_TO_B8(pixel);
                }
            }
        }
```

This is hardware-specific. The LUTs passed in our library are designed for use in software. I can't change the type definition as that's the same as asking for all the software code to change that process them right now.

kwagyeman: I can, however, move this loop into draw_image(). I guess this would save it from being copied into each GPU driver.

I think it should be in the GPU driver, though, right?

src/omv/imlib/imlib.h
kwagyeman: Done

src/omv/imlib/imlib.h
iabdalkader: I see.. Well if nothing depends on their actual size, making them portable is just a `typedef` to make then as large as possible (i.e., 4 bytes).

iabdalkader: > I think it should be in the GPU driver, though, right as the LUT definition is hardware-specific.

Yes it is. The problem with doing them on the fly like that is that I have to generate them every frame which might be slow.

kwagyeman: Tested the driver and it takes it about 10 ms to blend an 720x480 image even with color table loading.

kwagyeman:
Finished testing everything. I found a bug with dma2d maybe? It appears the alpha value loaded from the CLUT in L8 mode is ignored. Color values from the CLUT are applied as expected. The foreground register's alpha value modulates the the foreground alpha as expected... but, the CLUT seems to always act like alpha is 0xff.

Anyway, couldn't figure it out. Removed this code path from the DMA2D GPU. Software already maxes out the frame rate in the same case and produced the expected image.

kwagyeman:
Done, tested the modes of DMA2D. All works.

iabdalkader (changes requested):

src/omv/imlib/imlib.c
```diff
@@ -17,17 +17,21 @@
 #include "file_utils.h"
 #include "imlib.h"
 #include "omv_common.h"
+#include "omv_gpu.h"
 #include "omv_boardconfig.h"
 
 void imlib_init_all() {
+    #if (OMV_GPU_ENABLE == 1)
+    omv_gpu_init();
+    #endif
```
iabdalkader, line 26:
```suggestion
```
This should be called from main(), as mentioned before. I may need to call it once on first soft-reset.

src/omv/ports/stm32/omv_gpu.c
```diff
@@ -0,0 +1,247 @@
...
+    if (hint & IMAGE_HINT_BLACK_BACKGROUND) {
+        return false;
+    }
+    #endif
+
+    #if (TIME_GPU == 1)
+    mp_uint_t start = mp_hal_ticks_ms();
+    #endif
+
+    DMA2D_HandleTypeDef dma2d = {};
```
iabdalkader, line 66: Unless you really want to zero-initialize the struct, please avoid doing that. The compiler will zero-initialize every member and by most likely calling `memset`. This does not help performance.

src/omv/imlib/draw.c
```diff
@@ -3491,16 +3226,37 @@ void imlib_draw_image(image_t *dst_img,
...
 
-        if (&new_src_img == src_img) {
-            fb_free();
-        }
+        goto exit_cleanup;
+    }
 
-        return;
+    #if (OMV_GPU_ENABLE == 1)
+    // Can we offload this to the GPU for processing?
```
iabdalkader, line 3233:
```suggestion
    // Try to offload this to the GPU for processing.
```

src/omv/ports/stm32/omv_gpu.c
```diff
@@ -0,0 +1,247 @@
...
+
+    if (src_img->pixfmt == PIXFORMAT_GRAYSCALE) {
+        fb_free(); // clut
+    }
+
+    #if (TIME_GPU == 1)
+    printf("draw time: %u ms\n", mp_hal_ticks_ms() - start);
+    #endif
+
+    return true;
```
iabdalkader, line 236:
```suggestion
    return true;
```
Normally we return 0 and -1. This could have saved me an hour or so :)

iabdalkader (changes requested):

src/omv/ports/stm32/omv_gpu.c
```diff
@@ -0,0 +1,247 @@
...
+    if (src_img->pixfmt == PIXFORMAT_GRAYSCALE) {
+        uint8_t *src8 = IMAGE_COMPUTE_GRAYSCALE_PIXEL_ROW_PTR(src_img, src_rect->y) + src_rect->x;
+        src = (uint32_t) src8;
+
+        #if __DCACHE_PRESENT
+        uint8_t *src8_tmp = src8;
+        for (int i = 0; i < src_rect->h; i++) {
+            SCB_CleanDCache_by_Addr(src8_tmp, src_rect->w);
+            src8_tmp += src_img->w;
+        }
```
iabdalkader, line 191:
```suggestion
        for (int i = 0; i < src_rect->h; i++) {
            SCB_CleanDCache_by_Addr(src8_tmp, src_rect->w);
            src8_tmp += src_img->w;
        }
```
Why flush line by line ?

iabdalkader:
The last MCU row seems to be broken:
image

I don't think it's related to this PR. It might be the cache fixes that was merged before the release.

```python
import sensor, image, time

sensor.reset()

clock = time.clock()
while(True):
    clock.tick()
    img = sensor.snapshot()
    img.scale(x_scale=0.5)
    img.flush()
    print(clock.fps())

File: src/omv/common/omv_gpu.h [hunk 11]
```diff
@@ -0,0 +1,39 @@
...
+//
+// * IMAGE_HINT_BILINEAR: Use 2D bilinear interpolation when scaling the image versus nearest-neighbor.
+// * IMAGE_HINT_HMIRROR: The destination image result should be horizontally mirrored from the source image.
+// * IMAGE_HINT_VFLIP: The destination image result should be vertically flipped from the source image.
+// * IMAGE_HINT_BLACK_BACKGROUND: The destination image should be considered to be black (0) for alpha blending.
+
+bool omv_gpu_draw_image(image_t *dst_img,
+                        rectangle_t *dst_rect,
+                        image_t *src_img,
+                        rectangle_t *src_rect,
```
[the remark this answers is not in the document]
kwagyeman: Are you sure?

The dst image is the image being operated on. All image functions have that image as the first argument.

This also minors how imlib_draw_image() argument order.

src/omv/imlib/draw.c
kwagyeman: Done

src/omv/ports/stm32/omv_gpu.c
kwagyeman: Done

src/omv/ports/stm32/omv_gpu.c
kwagyeman: You might only be drawing a sub-rectangle of the image. There's no need to clean everything.

src/omv/ports/stm32/omv_gpu.c
kwagyeman: The goal was to zero the struct here as the config is on the stack. I can move it to a static var.

src/omv/common/omv_gpu.h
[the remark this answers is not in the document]
iabdalkader: Yes it matches the GPU functions.. Only memcpy-like functions should use that convention.

src/omv/ports/stm32/omv_gpu.c
iabdalkader: If you set all member then you don't need to zero the struct. Stack is better, as it saves BSS.

src/omv/ports/stm32/omv_gpu.c
iabdalkader: But it will be extremely slower in normal use cases. I don't care much about the DMA2D just pointing that out.

src/omv/ports/stm32/omv_gpu.c
kwagyeman: Done

src/omv/common/omv_gpu.h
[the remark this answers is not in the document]
kwagyeman: Okay

src/omv/ports/stm32/omv_gpu.c
```diff
@@ -0,0 +1,247 @@
...
+#define TIME_GPU    (0)
+#if (TIME_GPU == 1)
+#include "py/mphal.h"
+#include <stdio.h>
+#endif
+
+bool omv_gpu_draw_image(image_t *dst_img,
+                        rectangle_t *dst_rect,
+                        image_t *src_img,
+                        rectangle_t *src_rect,
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/imlib/draw.c
```diff
@@ -3491,16 +3226,37 @@ void imlib_draw_image(image_t *dst_img,
...
+        src_rect.x = src_x_accum_reset >> 16;
+        src_rect.y = src_y_accum_reset >> 16;
+        src_rect.w = ((src_x_accum_reset + (src_x_frac * dst_rect.w)) >> 16) - src_rect.x;
+        src_rect.h = ((src_y_accum_reset + (src_y_frac * dst_rect.h)) >> 16) - src_rect.y;
+
+        image_hint_t gpu_hint = hint & (IMAGE_HINT_BILINEAR | IMAGE_HINT_BLACK_BACKGROUND);
+        gpu_hint |= (dst_delta_x < 0) ? IMAGE_HINT_HMIRROR : 0;
+        gpu_hint |= (dst_delta_y < 0) ? IMAGE_HINT_VFLIP : 0;
+
+        if (omv_gpu_draw_image(dst_img, &dst_rect, src_img, &src_rect, alpha, color_palette, alpha_palette, gpu_hint)) {
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/ports/stm32/omv_gpu.c
```diff
@@ -0,0 +1,247 @@
...
+}
+
+void omv_gpu_init() {
+}
+
+void omv_gpu_deinit() {
+    DMA2D_HandleTypeDef dma2d = {};
+    dma2d.Instance = DMA2D;
+    HAL_DMA2D_DeInit(&dma2d);
+}
```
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/common/omv_gpu.h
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/ports/stm32/omv_gpu.c
kwagyeman: Not all members are set. It's got to be zeroed. So, moving to bss.

kwagyeman:
Updated

src/omv/ports/stm32/omv_gpu.c
iabdalkader: Please leave to save the BSS, it doesn't need to be global.

kwagyeman: Put it back on the stack and zeroed it. However, I removed code that was assigning zeroed members constant values of 0. I feel more comfortable zeroing the whole thing since ST adds new fields each CPU.

src/omv/common/omv_gpu.h
```diff
@@ -0,0 +1,39 @@
...
+ * This work is licensed under the MIT license, see the file LICENSE for details.
+ *
+ * GPU driver.
+ */
+#ifndef __OMV_GPU_H__
+#define __OMV_GPU_H__
+#include "imlib.h"
+
+int omv_gpu_init();
+int omv_gpu_deinit();
```
iabdalkader, line 16:
```suggestion
void omv_gpu_deinit();
```

src/omv/ports/stm32/omv_gpu.c
```diff
@@ -0,0 +1,230 @@
...
+#if (TIME_GPU == 1)
+#include "py/mphal.h"
+#include <stdio.h>
+#endif
+
+int omv_gpu_init() {
+    return 0;
+}
+
+int omv_gpu_deinit() {
```
iabdalkader, line 28:
```suggestion
void omv_gpu_deinit() {
```

src/omv/ports/stm32/omv_gpu.c
```diff
@@ -0,0 +1,230 @@
...
+    DMA2D_HandleTypeDef dma2d = {};
+    dma2d.Instance = DMA2D;
+    HAL_DMA2D_DeInit(&dma2d);
+    return 0;
```
iabdalkader, line 32:
```suggestion
```

iabdalkader (changes requested):

iabdalkader (approved):

PR #2335, opened 2024-07-31
ports/stm32: Fix MPU region protection bug.

kwagyeman:
While the GC block allocation code worked correctly, the MPU protection code suffered from two bugs:

1. The MPU was being programmed to protect the dereferenced memory address of the address it should protect.
2. The MPU protection size was 1/4th of what it should be.

### DIFF
--- a/src/omv/common/mp_utils.c
+++ b/src/omv/common/mp_utils.c
@@ -140,7 +140,7 @@ void mp_init_gc_stack(void *sstack, void *estack, void *heap_start, void *heap_e
     // Add GC blocks (if enabled).
     #ifdef OMV_GC_BLOCK1_MEMORY
     typedef struct {
-        uint32_t *addr;
+        uint8_t *addr;
         uint32_t size;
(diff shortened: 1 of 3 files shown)


iabdalkader:

https://forums.openmv.io/t/fir-module-not-detected/9772/6

iabdalkader:

https://forums.openmv.io/t/preview-image-on-the-serial-terminal/9859/27

iabdalkader:
I have no idea how that happened. Thanks for finding and fixing this!

src/omv/common/mp_utils.c
```diff
@@ -140,7 +140,7 @@ void mp_init_gc_stack(void *sstack, void *estack, void *heap_start, void *heap_e
     // Add GC blocks (if enabled).
     #ifdef OMV_GC_BLOCK1_MEMORY
     typedef struct {
-        uint32_t *addr;
```
iabdalkader, line 143: Actually this doesn't need to be a pointer at all.

PR #2339, opened 2024-07-31
common: Change linker variable to uint32.

kwagyeman:
Tested

### DIFF
--- a/src/omv/ports/stm32/stm32fxxx_hal_msp.c
+++ b/src/omv/ports/stm32/stm32fxxx_hal_msp.c
@@ -53,7 +53,7 @@ void HAL_MspInit(void) {
     }
 
     typedef struct {
-        uint8_t *addr;
+        uint32_t addr;
         uint32_t size;
     } dma_memory_table_t;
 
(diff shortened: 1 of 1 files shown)


iabdalkader:
I actually realized that it's more work if it's uint32_t that's why I released without it.

kwagyeman:
Close?

iabdalkader:
Yup. Thanks though!

iabdalkader:
Maybe it makes sense just for the MPU since it expects uint32_t ?

src/omv/common/mp_utils.c
```diff
@@ -140,15 +140,15 @@ void mp_init_gc_stack(void *sstack, void *estack, void *heap_start, void *heap_e
     // Add GC blocks (if enabled).
     #ifdef OMV_GC_BLOCK1_MEMORY
     typedef struct {
-        uint8_t *addr;
+        uint32_t addr;
```
iabdalkader, line 143: We just revert this then we're good.

kwagyeman: k

kwagyeman: Done

iabdalkader (approved):

PR #2343, opened 2024-08-02
ports/mimxrt: Add PXP GPU driver.

kwagyeman:
This is a draft PR for the PXP driver. While I got the GPU working... it is not reliable.

For example, if I ask it to copy an XGA or higher grayscale image from one buffer to another, with no scaling or anything else going on, it will fail to write out the last remaining row of blocks and throw an AXI READ error. Once you go below the XGA resolution it's fine. However, if you ask it to scale VGA by 2X it's fine (which produces a larger resolution than XGA). Or if you ask it to scale down WQXGA (2560x1600) down by 2X it's fine. But, it cannot copy WQXGA from one buffer to another with no scaling without corrupting the last row of 8x8/16x16 output blocks in the image.

Here's an example image. The error is the same each time, the system encounters a bus error and stops while working on the last row:

image

Note that I've run the code, and sometimes it doesn't happen. It might also have to do with some registers, possibly in the PXP, that don't reset. I was running the code and not seeing any errors until I pushed it to too high a res, and then the error started happening again.

I cannot determine what rules I need to follow to avoid the unit throwing an AXI error and stopping. The error cases seem more or less random after you go beyond the SVGA resolution.

...

On the performance side of things, I didn't really notice any difference between the PXP and the CPU. The IMXRT is memory bandwidth limited on the SDRAM. So, the CPU is able to keep up with the PXP. It probably helps more with an RTOS and running the display in the background.

Anyway, this PR should stay a draft until bugs are fixed, even then, it may never make sense to use. The CPU is generally better and more flexible.

That said, using the PXP for acceleration on the IMXRT has now been explored.

### DIFF
--- a/src/omv/ports/mimxrt/omv_gpu.c
+++ b/src/omv/ports/mimxrt/omv_gpu.c
@@ -1,0 +1,235 @@
+/*
+ * This file is part of the OpenMV project.
+ *
+ * Copyright (c) 2013-2024 Ibrahim Abdelkader <iabdalkader@openmv.io>
+ * Copyright (c) 2013-2024 Kwabena W. Agyeman <kwagyeman@openmv.io>
+ *
+ * This work is licensed under the MIT license, see the file LICENSE for details.
+ *
+ * GPU driver for IMXRT port.
+ */
+#include "omv_boardconfig.h"
+#if (OMV_GPU_ENABLE == 1)
+#include "imlib.h"
+#include "fsl_pxp.h"
+
+int omv_gpu_init() {
+    return 0;
+}
+
+void omv_gpu_deinit() {
+    PXP_Deinit(PXP);
+}
+
+int omv_gpu_draw_image(image_t *src_img,
+                       rectangle_t *src_rect,
+                       image_t *dst_img,
+                       rectangle_t *dst_rect,
+                       int alpha,
+                       const uint16_t *color_palette,
+                       const uint8_t *alpha_palette,
+                       image_hint_t hint) {
+    OMV_PROFILE_START();
+
+    // PXP input image must be GRAYSCALE, RGB565, YUV422 or YVU422.
+    if ((src_img->pixfmt != PIXFORMAT_GRAYSCALE) && (src_img->pixfmt != PIXFORMAT_RGB565) &&
+        (src_img->pixfmt != PIXFORMAT_YUV422) && (src_img->pixfmt != PIXFORMAT_YVU422)) {
+        return -1;
+    }
+
+    // PXP output image must be GRAYSCALE or RGB565.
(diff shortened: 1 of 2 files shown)


iabdalkader:
Bus errors set some bits in fault registers you can likely find the exact cause. As for the performance, maybe it will be good for drawing lines etc.. if it supports that.

kwagyeman:
The PXP has a register that says the fault ID. But, it's always 0.

Anyway, on to SIMD work.

iabdalkader:
> The PXP has a register that says the fault ID. But, it's always 0.

If you go into a fault handler, you should read the fault registers.

src/omv/ports/mimxrt/omv_gpu.c
```diff
@@ -0,0 +1,247 @@
...
+    if ((hint & (IMAGE_HINT_HMIRROR | IMAGE_HINT_VFLIP)) &&
+        ((src_rect->w != dst_rect->w) || (src_rect->h != dst_rect->h))) {
+        return -1;
+    }
+
+    // PXP always applies bilinear scaling, which is always better than nearest-neighbor so ignore the hint.
+
+    #if (TIME_GPU == 1)
+    mp_uint_t start = mp_hal_ticks_ms();
+    #endif
```
iabdalkader, line 86: We have profiling macros now, see omv_gpu in stm32. The macro should be at the start & end of the function.

src/omv/ports/mimxrt/omv_gpu.c
```diff
@@ -0,0 +1,247 @@
...
+
+    PXP_ClearStatusFlags(PXP, kPXP_CompleteFlag);
+    PXP_Start(PXP);
+
+    while (!(kPXP_CompleteFlag & PXP_GetStatusFlags(PXP))) {
+        MICROPY_EVENT_POLL_HOOK
+    }
+
+    PXP_Reset(PXP);
+    PXP_Deinit(PXP);
```
iabdalkader, line 209: Maybe initialize once and just leave it initialized ?

iabdalkader:
Could you retest this and see if the draw_image bug fixed it ?

src/omv/ports/mimxrt/omv_gpu.c
kwagyeman: It crashes and locks up the CPU if you do.

src/omv/ports/mimxrt/omv_gpu.c
kwagyeman: Sure

kwagyeman:
Want to get bayer speed up first.

kwagyeman:
Copying and scaling down seem to work reliably now after a fix to draw image passing slightly offset scaling starting positions.

kwagyeman:
Note to update the license here to be non-commercial.

iabdalkader:
Don't want this feature.

PR #2352, opened 2024-08-04
imlib/jpeg: Minor clean-up.

iabdalkader:
Remove STORECODE macro and just use write_bits. This has absolutely no effect on performance, but makes the code easier read.

### DIFF
--- a/src/omv/imlib/jpege.c
+++ b/src/omv/imlib/jpege.c
@@ -974,7 +974,8 @@ typedef struct {
     int idx;
     int length;
     uint8_t *buf;
-    int bitc, bitb;
+    uint32_t bitb;
+    uint32_t bitc;
     bool realloc;
     bool overflow;
 } jpeg_buf_t;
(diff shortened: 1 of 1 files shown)


kwagyeman (approved):
Before the change:
* XGA GRAYSCALE jpeg compression at quality 90 on RT1062 is 58ms
* XGA RGB565 jpeg compression at quality 90 on RT1062 is 181ms

After the change:
* XGA GRAYSCALE jpeg compression at quality 90 on RT1062 is 61ms
* XGA RGB565 jpeg compression at quality 90 on RT1062 is 189ms

Change results in 4-5% slow down at large reses. I confirm image quality and output looks the same. Has no affect on performance for smaller reses.

Code is much more readable though with this change.

iabdalkader:
Super high resolutions are extremely slow anyway, but there's another reason, I don't actually remember why the macro had `if (iLen + iNewLen > 32) {` in there anymore.

iabdalkader:
BTW did you try disabling SIMD ? It actually sames about 1 or 2ms for small resolutions.

kwagyeman:
No, I just verified that the code still worked. Yeah, I noticed the SIMD improves performance on the STM32 but makes it slower on the RT1062 in my other PR.

PR #2356, opened 2024-08-05
imlib: Add support for sending raw preview frames.

iabdalkader:
This allows boards that don't have a hardware JPEG encoder, but do have fast scaling, to bypass JPEG encoding and send down-scaled raw frames. The raw preview frames are capped at a configurable max, but should not exceed ~60KBs.

### DIFF
--- a/src/omv/imlib/framebuffer.c
+++ b/src/omv/imlib/framebuffer.c
@@ -10,6 +10,7 @@
  */
 #include <stdio.h>
 #include "mpprint.h"
+#include "fmath.h"
 #include "framebuffer.h"
 #include "omv_boardconfig.h"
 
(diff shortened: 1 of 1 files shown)


src/omv/imlib/framebuffer.c
```diff
@@ -166,8 +167,32 @@ void framebuffer_update_jpeg_buffer() {
...
+                        memcpy(dst.pixels, src->pixels, image_size(src));
+                    } else {
+                        dst.w = OMV_RAW_PREVIEW_WIDTH;
+                        dst.h = OMV_RAW_PREVIEW_HEIGHT;
+                        if (src->w > src->h) {
+                            dst.h = fast_floorf(dst.h * (src->h / (float) src->w));
+                        } else if (src->h > src->w) {
+                            dst.w = fast_floorf(dst.w * (src->w / (float) src->h));
+                        }
+                        imlib_draw_image(&dst, src, 0, 0, 1.0f, 1.0f, NULL, -1, 255,
```
kwagyeman, line 188: Yeah, so, the 1.0f and 1.0f for x/y scale should definitely be the actual scaling values.

Not sure why this is passing. This invocation is telling the function to draw the image at the top left/right corner of the dst with no scaling.

I'll send a patch later. I can test with the RT1062 and switch it to raw image transport and update the IDE. I'll do that work next week maybe.

iabdalkader: The bilinear flag triggers the scaling, at least for the GPU. This works for me, feel free to update.

PR #2358, opened 2024-08-05
boards/OPENMV_RT1060: Add support for RAW streaming.

kwagyeman:
draw_image() was not being invoked correctly to cause it to scale. This has been fixed. Also, the offset for the bilinear scaler has been moved to affect only the software code. Finally, I switched the empty flag for jpeg streaming to -1 from 0 which fixes binary image support.

### DIFF
--- a/src/omv/imlib/framebuffer.c
+++ b/src/omv/imlib/framebuffer.c
@@ -86,6 +86,9 @@ void framebuffer_init0() {
     // Enable streaming.
     MAIN_FB()->streaming_enabled = true; // controlled by the OpenMV Cam.
 
+    // Set invalid size.
+    JPEG_FB()->size = -1;
+
     // Set default quality
     JPEG_FB()->quality = ((OMV_JPEG_QUALITY_HIGH - OMV_JPEG_QUALITY_LOW) / 2) + OMV_JPEG_QUALITY_LOW;
 
@@ -117,7 +120,7 @@ static void jpegbuffer_init_from_image(image_t *img) {
     if (img == NULL) {
         jpeg_framebuffer->w = 0;
         jpeg_framebuffer->h = 0;
-        jpeg_framebuffer->size = 0;
+        jpeg_framebuffer->size = -1;
     } else {
         jpeg_framebuffer->w = img->w;
         jpeg_framebuffer->h = img->h;
--- a/src/omv/imlib/framebuffer.c
+++ b/src/omv/imlib/framebuffer.c
@@ -167,29 +167,36 @@ void framebuffer_update_jpeg_buffer() {
                     .size = OMV_JPEG_BUFFER_SIZE_MAX,
                     .pixels = jpeg_framebuffer->pixels
                 };
+
+                bool compress = true;
                 bool overflow = false;
-                if (0) {
+
                 #if OMV_RAW_PREVIEW_ENABLE
-                } else if (src->is_mutable &&
-                           image_size(src) <= OMV_JPEG_BUFFER_SIZE_MAX) {
+                if (src->is_mutable) {
                     // Down-scale the frame (if necessary) and send the raw frame.
                     dst.size = src->bpp;
                     dst.pixfmt = src->pixfmt;
                     if (src->w <= OMV_RAW_PREVIEW_WIDTH && src->h <= OMV_RAW_PREVIEW_HEIGHT) {
-                        memcpy(dst.pixels, src->pixels, image_size(src));
-                    } else {
-                        dst.w = OMV_RAW_PREVIEW_WIDTH;
-                        dst.h = OMV_RAW_PREVIEW_HEIGHT;
-                        if (src->w > src->h) {
-                            dst.h = fast_floorf(dst.h * (src->h / (float) src->w));
-                        } else if (src->h > src->w) {
-                            dst.w = fast_floorf(dst.w * (src->w / (float) src->h));
+                        if (image_size(&dst) <= OMV_JPEG_BUFFER_SIZE_MAX) {
+                            memcpy(dst.pixels, src->pixels, image_size(src));
+                            compress = false;
                         }
-                        imlib_draw_image(&dst, src, 0, 0, 1.0f, 1.0f, NULL, -1, 255,
-                                         NULL, NULL, IMAGE_HINT_BILINEAR, NULL, NULL, NULL);
+                    } else {
+                        float x_scale = OMV_RAW_PREVIEW_WIDTH / (float) src->w;
+                        float y_scale = OMV_RAW_PREVIEW_HEIGHT / (float) src->h;
+                        float scale = IM_MIN(x_scale, y_scale);
+                        dst.w = fast_floorf(src->w * scale);
+                        dst.h = fast_floorf(src->h * scale);
+                        if (image_size(&dst) <= OMV_JPEG_BUFFER_SIZE_MAX) {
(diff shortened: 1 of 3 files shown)


src/omv/imlib/framebuffer.c
```diff
@@ -170,23 +170,20 @@ void framebuffer_update_jpeg_buffer() {
                 bool overflow = false;
                 if (0) {
                 #if OMV_RAW_PREVIEW_ENABLE
-                } else if (src->is_mutable &&
-                           image_size(src) <= OMV_JPEG_BUFFER_SIZE_MAX) {
```
kwagyeman, line 174: Since this is constant there isn't a need to compare it in run time. I tried adding a macro... but, the since the jpeg size is a linker variable it can't be compared by the pre-processor.

src/omv/imlib/framebuffer.c
```diff
@@ -170,23 +170,20 @@ void framebuffer_update_jpeg_buffer() {
...
+                } else if ((src->pixfmt == PIXFORMAT_GRAYSCALE) || (src->pixfmt == PIXFORMAT_RGB565)) {
```
iabdalkader, line 173: Please revert this and send a follow up fix for usbdbg if you like. Also you removed the check that raw image fits into the buffer.

src/omv/imlib/framebuffer.c
```diff
@@ -170,23 +170,20 @@ void framebuffer_update_jpeg_buffer() {
...
-                        dst.h = OMV_RAW_PREVIEW_HEIGHT;
-                        if (src->w > src->h) {
-                            dst.h = fast_floorf(dst.h * (src->h / (float) src->w));
-                        } else if (src->h > src->w) {
-                            dst.w = fast_floorf(dst.w * (src->w / (float) src->h));
-                        }
-                        imlib_draw_image(&dst, src, 0, 0, 1.0f, 1.0f, NULL, -1, 255,
-                                         NULL, NULL, IMAGE_HINT_BILINEAR, NULL, NULL, NULL);
+                        float x_scale = OMV_RAW_PREVIEW_WIDTH / (float) src->w;
+                        float y_scale = OMV_RAW_PREVIEW_HEIGHT / (float) src->h;
```
iabdalkader, line 181:
```suggestion
                        float x_scale = OMV_RAW_PREVIEW_WIDTH / (float) src->w;
                        float y_scale = OMV_RAW_PREVIEW_HEIGHT / (float) src->h;
```
What if the source image is smaller than the preview image ? Let's say the preview size is 240x240 and the source is 120x120, wouldn't this up scale the frame to 240x240 ?

src/omv/boards/OPENMV_RT1060/omv_boardconfig.h
```diff
@@ -18,6 +18,11 @@
 #define OMV_BOARD_UID_SIZE              3            // Unique ID size in words.
 #define OMV_BOARD_UID_OFFSET            12           // Bytes offset for multi-word UIDs.
 
+// RAW Preview configuration.
+#define OMV_RAW_PREVIEW_ENABLE          (1)
+#define OMV_RAW_PREVIEW_WIDTH           (800)
+#define OMV_RAW_PREVIEW_HEIGHT          (600)
```
iabdalkader, line 24:
```suggestion
#define OMV_RAW_PREVIEW_WIDTH           (800)
#define OMV_RAW_PREVIEW_HEIGHT          (600)
```
That's not really the purpose of this feature. Even if bandwidth allow it, you'd be spending a lot of time moving huge buffers to USB. The main purpose of this feature is to free CPU time. If you can't make this smaller you don't have to use it.

iabdalkader (changes requested):

src/omv/imlib/framebuffer.c
iabdalkader: I'm not following. So the solution is not to check at all to save 1 branch ?

iabdalkader: If a board is misconfigured this will save a lot of debugging time.

iabdalkader:
> draw_image() was not being invoked correctly to cause it to scale.

Yeah you're right, it is cropping and I just didn't notice that it's cropping. However I'm not sure either about the fix in this PR.

src/omv/imlib/framebuffer.c
kwagyeman: No, the first if statement catches that. So, this only applies if the image is larger than the buffer.

src/omv/boards/OPENMV_RT1060/omv_boardconfig.h
kwagyeman: If it's jpeg compressed at a larger resolution, it takes more CPU time than just memcpying. Same for scaling. You kinda want this size to be as large as possible to avoid doing anything other than a memcpy as this is the fast CPU operation.

Yes, this does make the image transfer size larger, but, you're getting a better trade over how long it would take the CPU to process the frame via scaling or JPEG.

src/omv/imlib/framebuffer.c
kwagyeman: Okay, will get binary working.

src/omv/imlib/framebuffer.c
kwagyeman: Yeah, will add back. It should be checked.

src/omv/imlib/framebuffer.c
iabdalkader: Actually may as well send it in the same PR.

src/omv/boards/OPENMV_RT1060/omv_boardconfig.h
iabdalkader: None of the boards use DMA to copy buffers to USB, not even the ST stack by default, but even if they did it will still consume a lot of the chip's bus bandwidth. So the CPU has to copy ~234 x 4KBs buffers per frame which can't be free. The trade-off is JPEG compression time for smaller preview image, not JPEG compression time for CPU time and bus bandwidth.

kwagyeman: The conditonals for this are somewhat complex on what is best. Will re-enable after PXP is working for the RT1062 which does hardware scaling.

An example, the camera FPS for turning this feature off or on is the same for SVGA and below. However, the IDE FPS changes and increases with this feature enabled.

Once you go above the RAW res limit and start scaling you get behavior that affects the camera FPS. Handling 1280x720 without this feature limits the camera FPS to 9.5, and with it, the system FPS goes up to 16. IDE FPS is then about half or each. So, in this case scaling is cheaper than JPEG compression which is so high it affects system performance while connected to the IDE.

Given that, the question then becomes, if I reduced the preview width/height, would this increase the camera FPS (not IDE FPS). Maxing out the scaling performance is required to find the answer to this as it is a cheaper operation than JPEG.

src/omv/imlib/framebuffer.c
kwagyeman: Done

src/omv/boards/OPENMV_RT1060/omv_boardconfig.h
kwagyeman: Dropped

src/omv/imlib/framebuffer.c
kwagyeman: Done, tested. Binary streaming works with the IDE.

image

kwagyeman: No updates to the IDE are required. It's working already.

iabdalkader (approved):

Issue #2361, opened 2024-08-06
Number of frame buffers is not set correctly.


iabdalkader:
I have a 1MB frame buffer and 512K FB alloc size and I'm still getting 1 vbuffer (by default) even for QQVGA, which causes capture to restart on every snapshot:

This might be just one of the reasons:
 https://github.com/openmv/openmv/blob/master/src/omv/imlib/framebuffer.c#L366

> // Find a buffering size automatically that doesn't use more than half.

There's already a guaranteed minimum for fb_alloc (which is 512KBs in this case), the code above makes it override that and take up to half of the frame buffer, for no good reason, especially now that TF memory has been moved to GC blocks.fb_alloc should Not be this greedy.

Expected behavior: When I define the FB size I expect to be able use it all for vbuffers, as many as the res allows, especially that I also reserve a minimum for fb_alloc.

iabdalkader:
Fixed

PR #2367, opened 2024-08-10
boards: Increase heap sizes on SDRAM boards.

kwagyeman:


iabdalkader:
I don't mind this change but it's not that easy. First of all issue #2361 must be fixed, as this change will make the situation worse. For example, after this change some scripts out there might suddenly get slower because FB may use less vbuffers.
Also GC doesn't scale well with bigger heaps, i.e., all allocs/frees in that block will be much slower. So we need to do some benchmarks. We may also need to use a bigger word size for the bigger heap, but I don't MP supports setting word size per GC dynamically.

kwagyeman:
Yeah, we shouldn't merge it until all issues are resolved.

iabdalkader:

https://forums.openmv.io/t/behavior-of-referenced-memory-sensor-dealloc-extra-fb/10125/6

PR #2368, opened 2024-08-10
imlib: Optimize debayering with Helium.

kwagyeman:
PR is ready.


src/omv/imlib/simd.h
```diff
@@ -0,0 +1,1455 @@
...
+    return r;
+    #endif
+}
+
+static inline v128_t vpkhtb(v128_t v0, v128_t v1) {
+    #if (__ARM_ARCH >= 8)
+    return (v128_t) vsriq_n_u32(v0.u32, v1.u32, 16);
+    #elif (__ARM_ARCH >= 7)
+    return (v128_t) {
+               .u32 = { __PKHTB(v0.u32[0], v1.u32[0], 16) }
```
iabdalkader, line 439:
```suggestion
        .u32 = { __PKHTB(v0.u32[0], v1.u32[0], 16) }
```
Seems like there's an extra 4 spaces in all these lines.

kwagyeman: Formatter requires that. If I remove it then it won't pass.

iabdalkader: This case is probably not handled correctly in the config, will see if I can fix it.

iabdalkader: I fixed this in the config, if you rebase and reformat it will use 4 spaces.

iabdalkader:
Building this with `DEBUG=1` with Helium fails due to some built-in expression's argument not being constant. Note you could use this:
```C
#ifndef NDEBUG
do_something_if_debug_enabled
#else
do_something_if_debug_disabled
#endif
```

src/omv/imlib/simd.h
kwagyeman: Updated... checking if passing now.

kwagyeman:
Use:

`void imlib_debayer_image_awb(image_t *dst, image_t *src, uint32_t r_out, uint32_t g_out, uint32_t b_out)`

For ISP-related stuff. It can debayer the image in-place and will optionally switch to a fast debayering mode when the dst image is 1/2 w & h of the src image.

r_out, g_out, b_out should be 1 for testing. These are meant to receive the results from the isp.c functions that generate AWB stats.

iabdalkader:
Could you always use the second half of the buffer for in-place ? It just makes the logic easier for the outside code. Also please add a mode arg, subsample, bilinear etc.. for the same reason: it just makes setting up things for bayer easier.

src/omv/imlib/bayer.c
```diff
@@ -1,937 +1,1647 @@
...
+            }
+        }
+    } else {
+        switch (dst->pixfmt) {
+            case PIXFORMAT_GRAYSCALE: {
+                image_t buf = {
+                    .w = dst->w,
+                    .h = VBAYER_BUF_BROWS,
+                    .pixfmt = PIXFORMAT_GRAYSCALE,
+                    .data = fb_alloc(dst->w * VBAYER_BUF_BROWS * sizeof(uint8_t), FB_ALLOC_NO_HINT),
```
iabdalkader, line 1565:
```suggestion
                    .data = fb_alloc(dst->w * VBAYER_BUF_BROWS * sizeof(uint8_t), FB_ALLOC_PREFER_SPEED

File: src/omv/imlib/bayer.c [hunk 3]
```diff
@@ -1,937 +1,1647 @@
...
+
+                fb_free(); // buf.data
+                break;
+            }
+            case PIXFORMAT_RGB565: {
+                image_t buf = {
+                    .w = dst->w,
+                    .h = VBAYER_BUF_BROWS,
+                    .pixfmt = PIXFORMAT_RGB565,
+                    .data = fb_alloc(dst->w * VBAYER_BUF_BROWS * sizeof(uint16_t), FB_ALLOC_NO_HINT),
```
iabdalkader, line 1605:
```suggestion
                    .data = fb_alloc(dst->w * VBAYER_BUF_BROWS * sizeof(uint16_t), FB_ALLOC_PREFER_SPEED),
```

iabdalkader:
There are a few issues here:
* RGB565 VGA works really well, but QVGA and QQVGA are broken.  Grayscale work well for all resolutions (except there's a grid, see the next issue).
* I think grayscale is not handled correctly for "GBRG". RGB565 looks right but with grayscale I get a grid. I tried all different formats, one produces valid grayscale, the other valid RGB565.
* The subsample mode is pretty much unusable. There's no default resolution that's a multiple of 32 bytes in width, and can't guarantee the height either. All I can try to ensure is that width and height are even so the output is w/2 h/2.

kwagyeman:
Also, change this:

image

The profiling macro changed. It doesn't need the function name anymore. Just:

OMV_PROFILE_START();
OMV_PROFILE_PRINT();

src/omv/imlib/bayer.c
kwagyeman: Done

src/omv/imlib/bayer.c
kwagyeman: Done

kwagyeman:
Had to deal with a lot of unrelated stuff today. I haven't finished everything up yet.

> While all resolutions work with Grayscale, RGB565 QVGA and QQVGA are broken. This might be the sensor shifting the pattern with different resolutions, so can you confirm that smaller reses are working ? If so, I'll just stick to VGA for now, this is not the actual sensor anyway.

I'm seeing this issue with fast mode. I'lll address it when refactoring the code to handle predicates. Not sure why it cares... It seems to work as long as you don't apply a crop to cause a bayer pattern shift and just use VGA/QVGA and nothing else.

> Grayscale shows a visible grid. RGB565 looks right, so the format GBRG is right one. However I tried all formats anyway, and one produces a valid grayscale, the other produces a valid RGB565, never both.

I don't see this on the OV5640 but I will test with the OV7725.

> The subsample mode is pretty much unusable. There's no default resolution that's a multiple of 32 bytes in width, and can't guarantee the height either. All I can try to ensure is that width and height are even so the output is w/2 h/2.

Will have this fixed tomorrow. Along with minor refactoring work to finish the PR.

> Could you use the second half of the buffer for in-place for grayscale too ?It just makes the logic easier for the outside code, unless it's significantly slower.

This just means its working out of place which it handles.

> Please add a mode arg, subsample, bilinear etc.. for the same reason: it just makes setting up things for bayer easier.

Please take a look at the example code for what I was thinking the API is first.

---

Here's how you can test it. This is not part of the PR and needs to be added to py_image.c:

```
static mp_obj_t py_image_debayer(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
    enum { ARG_max, ARG_fast, ARG_grayscale };
        { MP_QSTR_max, MP_ARG_BOOL | MP_ARG_KW_ONLY, {.u_bool = false} },
        { MP_QSTR_fast, MP_ARG_BOOL | MP_ARG_KW_ONLY, {.u_bool = false} },
        { MP_QSTR_grayscale, MP_ARG_BOOL | MP_ARG_KW_ONLY, {.u_bool = false} },
    };

    // Parse args.
    image_t *image = py_helper_arg_to_image(pos_args[0], ARG_IMAGE_UNCOMPRESSED);
    mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
    mp_arg_parse_all(n_args - 1, pos_args + 1, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);

    if (!image->is_bayer) {
        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected a BAYER image!"));
    }

    uint32_t r_out, g_out, b_out;

    // This shouldn't have an effect an an already white balanced image.
    // For a RAW bayer image where green is really strong it will fix the image.
    if (args[ARG_max].u_bool) {
        imlib_awb_rgb_max(image, &r_out, &g_out, &b_out); // white patch algorithm
    } else {
        imlib_awb_rgb_avg(image, &r_out, &g_out, &b_out); // gray world algorithm
    }

    fb_alloc_mark();

    image_t temp;
    temp.w = image->w / (args[ARG_fast].u_bool ? 2 : 1);
    temp.h = image->h / (args[ARG_fast].u_bool ? 2 : 1);
    temp.pixfmt = args[ARG_grayscale].u_bool ? PIXFORMAT_GRAYSCALE : PIXFORMAT_RGB565;

    // Convert in place.
    uint32_t size = image_size(&temp);
    bool fb = py_helper_is_equal_to_framebuffer(image);
    size_t buf_size = fb ? framebuffer_get_buffer_size() : image_size(image);
    PY_ASSERT_TRUE_MSG((size <= buf_size), "The image doesn't fit in the frame buffer!");
    temp.data = image->data;

    if ((!args[ARG_fast].u_bool) && (!args[ARG_grayscale].u_bool)) {
        // If RGB565 debayering with no 2x2 we need to move the image to the second half of the buffer.
        memcpy(image->data + image_size(image), image->data, image_size(image));
        image->data = image->data + image_size(image);
    }

    imlib_debayer_image_awb(&temp, image, r_out, g_out, b_out);
    fb_alloc_free_till_mark();
    py_helper_update_framebuffer(&temp);
    memcpy(image, &temp, sizeof(image_t));
    return py_image_from_struct(&temp);
}
static MP_DEFINE_CONST_FUN_OBJ_KW(py_image_debayer_obj, 1, py_image_debayer);
```

And then you can test with:

```
import sensor, image, time

sensor.reset()
sensor.set_pixformat(sensor.BAYER)
# sensor.__write_reg(0x3036, 0x20) # Needed to make the OV5640 work on the AE3
sensor.set_framebuffers(1)

while(True):
    img = sensor.snapshot()
    #img = img.crop(roi=[0, 0, 640-32, 478]) # Test BGGR
    #img = img.crop(roi=[0, 1, 640-32, 478]) # Test GRBG
    #img = img.crop(roi=[1, 0, 640-32, 478]) # Test GBRG
    #img = img.crop(roi=[1, 1, 640-32, 478]) # Test BGGR
    # Don't use the above with fast mode. Things break. Will debug.
    print(img)
    img.debayer(fast=True, grayscale=True)
```

iabdalkader:
Note about simd.h I'm trying to compile `bayer.c` with clang, but I get the following errors:

```
imlib/simd.h:514:12: error: argument to 'vgetq_lane_u8' must be a constant integer
  514 |     return vgetq_lane_u8(v0.u8, n);
```

About a 100 of them.. It seems that gcc somehow can inline it and then the immediate is a compile-time constant but clang couldn't. 

A simple fix for most of those functions is to use macros, for example:
```
#define vget_u8(v0, n) vgetq_lane_u8(v0.u8, n)
```

Would be nice to fix this as GCC's behavior is the wrong one so we can't count on it.

src/omv/imlib/simd.h
```diff
@@ -0,0 +1,1725 @@
...
+        .u32 = { start }
+    };
+    #endif
+}
+
+static inline v128_t vrol(v128_t v0, uint32_t n) {
+    #if (__ARM_ARCH >= 8)
+    #pragma GCC diagnostic push
+    #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+    uint32_t c;
```
iabdalkader, line 792:
```suggestion
    uint32_t c = 0;
```
Is `c` causing the issue ?

kwagyeman: Yeah, I was trying to avoid an ASM op assigning it to zero. However, the compiler assigns it to zero in the ASM. Will modify this and remove the diagnostic.

kwagyeman:
Updated test code:

```
static mp_obj_t py_image_debayer(uint n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
    enum { ARG_max, ARG_fast, ARG_grayscale };
        { MP_QSTR_max, MP_ARG_BOOL | MP_ARG_KW_ONLY, {.u_bool = false} },
        { MP_QSTR_fast, MP_ARG_BOOL | MP_ARG_KW_ONLY, {.u_bool = false} },
        { MP_QSTR_grayscale, MP_ARG_BOOL | MP_ARG_KW_ONLY, {.u_bool = false} },
    };

    // Parse args.
    image_t *image = py_helper_arg_to_image(pos_args[0], ARG_IMAGE_UNCOMPRESSED);
    mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
    mp_arg_parse_all(n_args - 1, pos_args + 1, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);

    if (!image->is_bayer) {
        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected a BAYER image!"));
    }

    uint32_t r_out, g_out, b_out;

    // This shouldn't have an effect an an already white balanced image.
    // For a RAW bayer image where green is really strong it will fix the image.
    if (args[ARG_max].u_bool) {
        imlib_awb_rgb_max(image, &r_out, &g_out, &b_out); // white patch algorithm
    } else {
        imlib_awb_rgb_avg(image, &r_out, &g_out, &b_out); // gray world algorithm
    }

    fb_alloc_mark();

    image_t temp;
    temp.w = image->w / (args[ARG_fast].u_bool ? 2 : 1);
    temp.h = image->h / (args[ARG_fast].u_bool ? 2 : 1);
    temp.pixfmt = args[ARG_grayscale].u_bool ? PIXFORMAT_GRAYSCALE : PIXFORMAT_RGB565;

    // Convert in place.
    uint32_t size = image_size(&temp);
    bool fb = py_helper_is_equal_to_framebuffer(image);
    size_t buf_size = fb ? framebuffer_get_buffer_size() : image_size(image);
    PY_ASSERT_TRUE_MSG((size <= buf_size), "The image doesn't fit in the frame buffer!");
    temp.data = image->data;

    if ((!args[ARG_fast].u_bool) && (!args[ARG_grayscale].u_bool)) {
        // If RGB565 debayering with no 2x2 we need to move the image to the second half of the buffer.
        memcpy(image->data + image_size(image), image->data, image_size(image));
        image->data = image->data + image_size(image);
    }

    imlib_debayer_image_awb(&temp, image, args[ARG_fast].u_bool, r_out, g_out, b_out);
    fb_alloc_free_till_mark();
    py_helper_update_framebuffer(&temp);
    memcpy(image, &temp, sizeof(image_t));
    return py_image_from_struct(&temp);
}
static MP_DEFINE_CONST_FUN_OBJ_KW(py_image_debayer_obj, 1, py_image_debayer);
```

```
import sensor, image, time

sensor.reset()
sensor.set_pixformat(sensor.BAYER)
#sensor.__write_reg(0x3036, 0x20) # Needed to make the OV5640 work on the AE3
sensor.set_framebuffers(1)

while(True):
    img = sensor.snapshot()
    #img = img.crop(roi=[0, 0, sensor.width(), sensor.height()]) # Test BGGR
    #img = img.crop(roi=[0, 1, sensor.width(), sensor.height()]) # Test GRBG
    #img = img.crop(roi=[1, 0, sensor.width(), sensor.height()]) # Test GBRG
    #img = img.crop(roi=[1, 1, sensor.width(), sensor.height()]) # Test BGGR
    print(img)
    img.debayer(fast=True, grayscale=True)
```

kwagyeman:
fast mode has been fixed. It's now controlled by a bool. It expects and image that's 1/2 width and height rounded down.

It also handles images of any res now.

iabdalkader:
I'm still getting a grid with Grayscale with OV7725, and fast mode still expects the image in the second half. To test with OV7725 set raw=1
```
sensor->hw_flags.raw = 1;
``
And in `set_pixformat` always set the format to bayer regardless of the args.

**Comment by kwagyeman** [comment]
The bayer flag was incorrectly set in this PR for the OV7725 sensor: https://github.com/openmv/openmv/pull/2360/files

If the bayer mode is not set its assumed to be BGGR.

**Comment by kwagyeman** [comment]
DEBUG mode was bugged from missing code on my part in imlib_bayer_shift. This has been fixed.

**Comment by kwagyeman** [comment]
Fixes: https://github.com/openmv/openmv/pull/2285

**Comment by kwagyeman** [comment]
Add this to imlib config to enable the max speed:

// Bayer
#define IMLIB_ENABLE_DEBAYER_MAX_SPEED
#define IMLIB_ENABLE_DEBAYER_YCBCR_MAX_SPEED
#define IMLIB_ENABLE_DEBAYER_AWB_MAX_SPEED

**Comment by iabdalkader** [comment]
> Add this to imlib config to enable the max speed:
> 
> // Bayer #define IMLIB_ENABLE_DEBAYER_MAX_SPEED #define IMLIB_ENABLE_DEBAYER_YCBCR_MAX_SPEED #define IMLIB_ENABLE_DEBAYER_AWB_MAX_SPEED

Can you reduce the size further for other boards ? Bayer is rarely used on other boards, so the old speed was fine for those boards. For boards that have the extra flash it doesn't matter. Also please combine all those flags in `IMLIB_ENABLE_DEBAYER_OPTIMIZATION`.

File: src/omv/sensors/ov7725.c [hunk 5]
```diff
@@ -683,7 +683,7 @@ int ov7725_init(sensor_t *sensor) {
     sensor->hw_flags.jpege = 0;
     sensor->hw_flags.gs_bpp = 2;
     sensor->hw_flags.rgb_swap = 1;
-    sensor->hw_flags.bayer = SENSOR_HW_FLAGS_BAYER_GBRG;
+    sensor->hw_flags.bayer = SENSOR_HW_FLAGS_BAYER_BGGR;
```
iabdalkader, line 686: This produces a broken RGB565 image, but a good Grayscale. The old one produces a good RGB565 and bad Grayscale.

iabdalkader:
With the OV7725, awb produces broken images for resolutions < VGA.
image

iabdalkader (approved):

PR #2372, opened 2024-08-12
Optimization of software JPEG encoder

bitbank2:
This change improves the speed of the software JPEG encoder by optimizing the output bitstream logic. It also changes the default color subsample option to 4:2:0 when not specified by the user.


kwagyeman:
@bitbank2 - Please fix the PR to pass the two failing checks.

I will review this PR after finishing with the SIMD bayer code changes.

kwagyeman:
For the commit messages, they should be:

"imlib/jpeg.c: Speed up jpeg compressor by doing xxxx."
"imlib/jpeg.c: Changed default sampling to xxx."

Then the PR title should be one of those commits.

Also, please do this to get rid of the merge with master, that should not be in the PR. I can provide help support for how to fix that via the slack.

While it's a bit of a learning curve doing clean commits, it ultimately makes us go faster as we can use the automated tools.

bitbank2:
closing to clean up the commit messages

PR #2374, opened 2024-08-14
imlib/framebuffer: Set new auto-adjustment policy.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2361

### DIFF
--- a/src/omv/imlib/framebuffer.c
+++ b/src/omv/imlib/framebuffer.c
@@ -367,14 +367,29 @@ void framebuffer_auto_adjust_buffers() {
         return;
     }
 
-    for (int i = 3; i > 0; i--) {
-        framebuffer_set_buffers(i);
+    #if (OMV_CSI_HW_CROP_ENABLE == 1)
+    if (framebuffer->u && framebuffer->v) {
+        // Typically a framebuffer will not need more than u*v*2 bytes.
+        uint32_t size_guess = sizeof(vbuffer_t) + (framebuffer->u * framebuffer->v * 2);
+        // Add in extra bytes to prevent round down from shrinking buffer too small.
+        size_guess += FRAMEBUFFER_ALIGNMENT - 1;
+        // Limit the frame buffer size guess.
+        size_guess = FB_ALIGN_SIZE_ROUND_DOWN(size_guess);
 
-        // Find a buffering size automatically that doesn't use more than half.
-        if (fb_avail() >= framebuffer_total_buffer_size()) {
-            return;
+        for (int i = 3; i > 0; i--) {
+            framebuffer_set_buffers(i);
+
+            // When we start to run out of memory this will stop being true.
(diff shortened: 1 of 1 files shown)


src/omv/imlib/framebuffer.c
```diff
@@ -367,14 +367,29 @@ void framebuffer_auto_adjust_buffers() {
...
+        for (int i = 3; i > 0; i--) {
+            framebuffer_set_buffers(i);
+
+            // When we start to run out of memory this will stop being true.
+            if ((size_guess * i) <= framebuffer_total_buffer_size()) {
+                return;
+            }
         }
+    } else {
+        framebuffer_set_buffers(3);
```
kwagyeman, line 388: Leaving this as a draft right now because it's not clear what to do when I have no size information on the contents of the frame buffer. I guess just set 3 buffers?

src/omv/imlib/framebuffer.c
```diff
@@ -367,14 +367,29 @@ void framebuffer_auto_adjust_buffers() {
...
     }
+    #else
+    framebuffer_set_buffers(3);
```
kwagyeman, line 391: Same issue here. Do you have any thoughts on the policy if there's no hardware crop? Am I allowed to use size information?

src/omv/imlib/framebuffer.c
```diff
@@ -367,14 +367,29 @@ void framebuffer_auto_adjust_buffers() {
...
+        size_guess = FB_ALIGN_SIZE_ROUND_DOWN(size_guess);
 
-        // Find a buffering size automatically that doesn't use more than half.
-        if (fb_avail() >= framebuffer_total_buffer_size()) {
-            return;
+        for (int i = 3; i > 0; i--) {
+            framebuffer_set_buffers(i);
+
+            // When we start to run out of memory this will stop being true.
+            if ((size_guess * i) <= framebuffer_total_buffer_size()) {
```
kwagyeman, line 383: The new policy is as you asked. I need to test this, but I wanted feedback on the other policy cases.

iabdalkader: No this is all bad logic we don't need more of it.

Let's ignore the HW cropping issue for a moment, because eventually I can override it and use the sensor's ROI feature to force the readout to match the window, which is what I want to do anyway to save the GPU cropping time.

Auto-adjust is called twice, once for pix format once for frame size, in any order. It should not do anything until it has both w/h and bpp. Once you have w,h,bpp you can find out the actual frame size then divide the `FB size / n_buffers--` until things fit, and that's it.

iabdalkader: I gave it a try: https://github.com/openmv/openmv/pull/2376

kwagyeman: Cool, closing this PR. Yeah, this logic was bad.

PR #2376, opened 2024-08-14
imlib: Fix automatic vbuffer count.

iabdalkader:
Before this patch, a 1.5MB framebuffer used a single vbuffer for VGA and smaller resolutions, causing the sensor driver to restart with every snapshot. With this patch, a 1.5MB frame buffer uses 2 vbuffers for VGA/RGB and 3 vbuffers for VGA/grayscale **by default**. In the case of 2 vbuffers, it should leave the rest for `fb_alloc` (not tested), so `fb_alloc` still gets some extra space, but only if there's any left. If you need more space for `fb_alloc`, reserve more if possible.

Additionally, `set_buffers` is now more efficient with buffer sizes. For example, if the source is 1bpp (Bayer) and the destination is 1bpp (grayscale or Bayer), the vbuffer size will be `w*h`, instead of assuming that everything is 2bpp, which allows for more vbuffers.

`set_buffers` now replaces `auto_adjust`. Passing `-1` to `set_buffers` will attempt to use 3 vbuffers, each with a size of `frame_size` if they fit; otherwise, the maximum possible buffers will be used. Passing `1` will use the whole framebuffer. In this case, `frame_size` is ignored.

**Note this definitely needs some more testing and I may have missed some corner cases.**

### DIFF
--- a/src/omv/imlib/framebuffer.c
+++ b/src/omv/imlib/framebuffer.c
@@ -262,45 +262,28 @@ int32_t framebuffer_get_depth() {
     return framebuffer->bpp;
 }
 
-// Returns the number of bytes the frame buffer could be at the current moment it time.
-static uint32_t framebuffer_raw_buffer_size() {
-    uint32_t size = (uint32_t) (fb_alloc_stack_pointer() - ((char *) framebuffer->data));
-    // We don't want to give all of the frame buffer RAM to the frame buffer. So, we will limit
-    // the maximum amount of RAM we return.
-    uint32_t raw_buf_size = (&_fb_memory_end - &_fb_memory_start - sizeof(framebuffer_t));
-    return IM_MIN(size, raw_buf_size);
+// Returns the current frame buffer size, factoring in the space taken by fb_alloc.
+static uint32_t framebuffer_max_buffer_size() {
+    uint32_t fb_total_size = FB_ALIGN_SIZE_ROUND_DOWN(&_fb_memory_end - (char *) framebuffer->data);
+    uint32_t fb_avail_size = FB_ALIGN_SIZE_ROUND_DOWN(fb_alloc_stack_pointer() - (char *) framebuffer->data);
+    return IM_MIN(fb_total_size, fb_avail_size);
 }
 
 uint32_t framebuffer_get_buffer_size() {
     uint32_t size;
 
     if (framebuffer->n_buffers == 1) {
-        // With only 1 vbuffer it's fine to allow the frame buffer size to change given fb_alloc().
-        size = framebuffer_raw_buffer_size();
+        // With only 1 vbuffer the frame buffer size can change given fb_alloc().
+        size = framebuffer_max_buffer_size();
     } else {
         // Whatever the raw size was when the number of buffers were set is locked in.
-        size = framebuffer->raw_buffer_size;
+        size = framebuffer->buff_size;
     }
 
     // Remove the size of the state header plus alignment padding.
     size -= sizeof(vbuffer_t);
 
-    #if (OMV_CSI_HW_CROP_ENABLE == 1)
-    // If the frame size is set, the memory for each buffer can be reduced,
-    // freeing up space for fb_alloc(). Note that this can only be done if
-    // the camera interface supports hardware cropping, i.e., the actual
-    // frame size will match the specified window size.
-    if ((framebuffer->n_buffers != 1) && framebuffer->u && framebuffer->v) {
-        // Typically a framebuffer will not need more than u*v*2 bytes.
-        uint32_t size_guess = framebuffer->u * framebuffer->v * 2;
-        // Add in extra bytes to prevent round down from shrinking buffer too small.
-        size_guess += FRAMEBUFFER_ALIGNMENT - 1;
-        // Limit the frame buffer size.
-        size = IM_MIN(size, size_guess);
-    }
-    #endif
-
-    // Needs to be a multiple of FRAMEBUFFER_ALIGNMENT for DMA transfers...
+    // Needs to be a multiple of FRAMEBUFFER_ALIGNMENT for DMA transfers.
     return FB_ALIGN_SIZE_ROUND_DOWN(size);
 }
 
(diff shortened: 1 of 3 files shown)


src/omv/imlib/framebuffer.h
```diff
@@ -34,6 +34,7 @@ typedef struct framebuffer {
     volatile int32_t tail;
     bool check_head;
     int32_t sampled_head;
+    uint32_t frame_size;
```
kwagyeman, line 37: Might want to name this something else since sensor.framesize means something completely different.

src/omv/imlib/framebuffer.c
```diff
@@ -327,19 +312,21 @@ void framebuffer_flush_buffers(bool fifo_flush) {
 
 int framebuffer_set_buffers(int32_t n_buffers) {
     uint32_t total_size = framebuffer_raw_buffer_size();
-    uint32_t size = total_size / n_buffers;
+    uint32_t vbuff_size = (n_buffers == 1) ?
+                          total_size : FB_ALIGN_SIZE_ROUND_UP(framebuffer->frame_size + sizeof(vbuffer_t));
+    int32_t max_buffers = (n_buffers == 1) ? n_buffers : (total_size / vbuff_size);
```
kwagyeman, line 317:
```suggestion
    uint32_t vbuff_size = (n_buffers == 1) ?
                          FB_ALIGN_SIZE_ROUND_DOWN(total_size -  sizeof(vbuffer_t)) :
                          FB_ALIGN_SIZE_ROUND_UP(framebuffer->frame_size + sizeof(vbuffer_t));
    int32_t max_buffers = total_size / vbuff_size;
```

I think you need to round down the total size as this doesn't include the vbuffer header which would use some space.

Then max_buffers when n_buffers is one should be the same calculation either way.

src/omv/imlib/framebuffer.c
```diff
@@ -327,19 +312,21 @@ void framebuffer_flush_buffers(bool fifo_flush) {
...
+    uint32_t vbuff_count = (n_buffers == -1) ? IM_MIN(max_buffers, 3) : IM_MIN(max_buffers, n_buffers);
 
-    // Error out if frame buffers are smaller than this...
-    if (size < (sizeof(vbuffer_t) + FRAMEBUFFER_ALIGNMENT)) {
+    if (vbuff_count == 0 || vbuff_size < sizeof(vbuffer_t)) {
```
kwagyeman, line 320:
```suggestion
    if (vbuff_count == 0 || vbuff_size < (sizeof(vbuffer_t) + FRAMEBUFFER_ALIGNMENT)) {
```

Should leave the alignment padding as part of this.

kwagyeman (approved):

src/omv/imlib/framebuffer.h
iabdalkader: `raw_buffer_size` is also very misleading. It seems to be the vbuffer buffer size + header but it's called raw.

src/omv/imlib/framebuffer.c
iabdalkader: > Then max_buffers when n_buffers is one should be the same calculation either way.

How can `total_size` equal `FB_ALIGN_SIZE_ROUND_DOWN(total_size -  sizeof(vbuffer_t))` for this to work:
```C
     uint32_t vbuff_size = (n_buffers == 1) ?
                          FB_ALIGN_SIZE_ROUND_DOWN(total_size -  sizeof(vbuffer_t)) :
    int32_t max_buffers = total_size / vbuff_size;
```

iabdalkader: I could rearrange things to make that work though. However, vbuff_size here already includes the header size.

src/omv/imlib/framebuffer.c
iabdalkader: It's already rounded down/up, right ?

iabdalkader:
```C
uint32_t framebuffer_get_buffer_size() {
    uint32_t size;

    if (framebuffer->n_buffers == 1) {
        // With only 1 vbuffer it's fine to allow the frame buffer size to change given fb_alloc().
        size = framebuffer_raw_buffer_size();
    } else {
        // Whatever the raw size was when the number of buffers were set is locked in.
        size = framebuffer->raw_buffer_size;
    }

```

Shouldn't those be the same now ?

src/omv/imlib/framebuffer.h
iabdalkader: `frame_size` doesn't need to be stored, it can be passed as an arg, but I want to keep the function accepting a single arg (count).

src/omv/imlib/framebuffer.c
kwagyeman: The vbuffer header yes, but, if the frame buffer storage area itself is 1 byte it will take up 32.

src/omv/imlib/framebuffer.c
kwagyeman: vbuff_size would always be slightly less than total_size so the result would be 1. Which was happening before with (n_buffers == 1) ? n_buffers.

src/omv/imlib/framebuffer.h
kwagyeman: Yeah, I was just suggesting giving it a better name.

src/omv/imlib/framebuffer.c
```diff
@@ -326,24 +309,21 @@ void framebuffer_flush_buffers(bool fifo_flush) {
 }
 
 int framebuffer_set_buffers(int32_t n_buffers) {
-    uint32_t total_size = framebuffer_raw_buffer_size();
-    uint32_t size = total_size / n_buffers;
+    uint32_t avail_size = FB_ALIGN_SIZE_ROUND_DOWN(framebuffer_max_buffer_size());
```
kwagyeman, line 312:
```suggestion
    uint32_t avail_size = FB_ALIGN_SIZE_ROUND_DOWN(framebuffer_max_buffer_size() - sizeof(vbuffer_t));
```

One frame buffer still has a vbuffer header that needs to be removed.

src/omv/imlib/framebuffer.c
kwagyeman: I guess the frame size could be 0. So... I suppose it's fine.

src/omv/imlib/framebuffer.c
iabdalkader: >vbuff_size would always be slightly less than total_size so the result would be 1

Yes that's right. Well now it's equal so will be 1.

src/omv/imlib/framebuffer.h
iabdalkader: I can't think of better names for this, it is a frame size. However, I'm still thinking about just passing it as a second arg to `set_buffers`. The reason for this is that if you ever call this function with anything other than `1` you must remember to set `frame_size` first, otherwise it will use whatever was there.

iabdalkader:
Assuming this is all good, merging.

PR #2381, opened 2024-08-19
imlib/jpege.c: Sped up the compressor by making the bitstream writer more efficient. Also changed the default color subsampling to 4:2:0

bitbank2:
This change speeds up color JPEG compression by about 9% compared to the previous code. It accomplishes this by making the bitstream writer more efficient through the use of local/register variables and not checking for output buffer overflow after each byte. The default color subsampling has also been changed from 4:4:4 to 4:2:0; this also speeds up compression quite a bit when the user doesn't explicitly choose a color subsampling scheme.


kwagyeman:
Make your commit message shorter and add a period at the end. Then make the PR title the same.

kwagyeman:
Also, change this:

image

The profiling macro changed. It doesn't need the function name anymore. Just:

OMV_PROFILE_START();
OMV_PROFILE_PRINT();

PR #2400, opened 2024-08-30
imlib/apriltag: Switch to libm's sin/cos.

iabdalkader:


kwagyeman (approved):
Why remove the BLE advertising?

PR #2401, opened 2024-08-30
imlib/apriltag: Disable unused TAG families by default.

kwagyeman:
Disable extra tag families that no one uses. This saves flash and makes the code run faster as you don't have to search through unused tag families. The IDE recommends TAG36H11 right now. The other families should generally be avoided anyway.


scripts/examples/06-April-Tags/find_apriltags_max_res.py
```diff
@@ -31,47 +30,15 @@
...
     clock.tick()
     img = sensor.snapshot()
-    for tag in img.find_apriltags(
-        families=tag_families
-    ):  # defaults to TAG36H11 without "families".
+    for tag in img.find_apriltags():
         img.draw_rectangle(tag.rect(), color=127)
         img.draw_cross(tag.cx(), tag.cy(), color=127)
-        print_args = (family_name(tag), tag.id(), (180 * tag.rotation()) / math.pi)
+        print_args = ("TAG36H11", tag.id(), (180 * tag.rotation()) / math.pi)
```
iabdalkader, line 42:
```suggestion
        print_args = ("TAG36H11", tag.id(), (180 * tag.rotation()) / math.pi)
```
But some boards will have other families enabled, maybe return the name with or `tag.family` ? Also converting all of those to attributes will probably save lots of space.

src/omv/boards/ARDUINO_GIGA/imlib_config.h
```diff
@@ -96,6 +96,12 @@
 
 // Enable find_apriltags() (64 KB)
 #define IMLIB_ENABLE_APRILTAGS
+//#define IMLIB_ENABLE_APRILTAGS_TAG16H5
+//#define IMLIB_ENABLE_APRILTAGS_TAG25H7
+//#define IMLIB_ENABLE_APRILTAGS_TAG25H9
+//#define IMLIB_ENABLE_APRILTAGS_TAG36H10
+#define IMLIB_ENABLE_APRILTAGS_TAG36H11
+//#define IMLIB_ENABLE_APRILTAGS_ARTOOLKIT
```
iabdalkader, line 104:
```suggestion
#define IMLIB_ENABLE_APRILTAGS_ARTOOLKIT
```
I'd leave this one too by default for all boards, as it's popular.

src/omv/boards/ARDUINO_GIGA/imlib_config.h
```diff
@@ -96,6 +96,12 @@
 
 // Enable find_apriltags() (64 KB)
 #define IMLIB_ENABLE_APRILTAGS
+//#define IMLIB_ENABLE_APRILTAGS_TAG16H5
+//#define IMLIB_ENABLE_APRILTAGS_TAG25H7
+//#define IMLIB_ENABLE_APRILTAGS_TAG25H9
+//#define IMLIB_ENABLE_APRILTAGS_TAG36H10
```
iabdalkader, line 102:
```suggestion
```
Let's just define the enabled ones, until we have a common `imlib_config.h`

iabdalkader (changes requested):

iabdalkader:
@kwagyeman Let's get this change merged and then do a release.

src/omv/boards/ARDUINO_GIGA/imlib_config.h
kwagyeman: I've never seen people to use it. Sure? This isn't the same as the ArUco codes that OpenCV works with. I think it's just a old tag formt that AprilTags was using. It's #2 of this image: 
image

scripts/examples/06-April-Tags/find_apriltags_max_res.py
kwagyeman: Should this be in another PR? Providing the family name as a function is a missing feature. I could make it as an attribute, but, converting all those over should be done at the same time in a release with other breaking changes I think.

src/omv/boards/ARDUINO_GIGA/imlib_config.h
kwagyeman: Okay

scripts/examples/06-April-Tags/find_apriltags_max_res.py
iabdalkader: Okay but the script can use that attribute now.

kwagyeman: Okay, will make an attribute with the family name string.

src/omv/boards/ARDUINO_GIGA/imlib_config.h
iabdalkader: No I'm not sure, it just sounded familiar. Okay I guess the default one is fine.

scripts/examples/06-April-Tags/find_apriltags_max_res.py
iabdalkader: Actually you should convert them now while we're at it, this way we can see the total size savings ?

kwagyeman: Done, added changed in another commit for this. But, I cannot test this while on the road.

kwagyeman: Looks like ~224 bytes or so. I deleted the subscr function though which probably accounts for those savings. Moving to attribute seems to be a minor savings.

src/omv/modules/py_image.c
```diff
@@ -5096,190 +5096,123 @@ static void py_apriltag_print(const mp_print_t *print, mp_obj_t self_in, mp_prin
...
+                dest[0] = self->h;
+                break;
+            case MP_QSTR_id:
+                dest[0] = self->id;
+                break;
+            case MP_QSTR_family:
+                dest[0] = self->family;
+                break;
+            case MP_QSTR_family_str:
+                #ifdef IMLIB_ENABLE_APRILTAGS_TAG16H5
```
iabdalkader, line 5129: There's no need for this `apriltag_families` are always defined, a switch statement here is fine even if not all of them are enabled.

src/omv/modules/py_image.c
```diff
@@ -5096,190 +5096,123 @@ static void py_apriltag_print(const mp_print_t *print, mp_obj_t self_in, mp_prin
...
+                if (mp_obj_get_int(self->family) == TAG16H5) {
+                    dest[0] = "TAG16H5";
```
iabdalkader, line 5131: You can't return a const chat here of course, you need an mp_obj there's a str from cstr function somewhere.

scripts/examples/06-April-Tags/find_apriltags_w_lens_zoom.py
```diff
@@ -32,8 +32,8 @@
     clock.tick()
     img = sensor.snapshot()
     for tag in img.find_apriltags():  # defaults to TAG36H11
-        img.draw_rectangle(tag.rect(), color=(255, 0, 0))
-        img.draw_cross(tag.cx(), tag.cy(), color=(0, 255, 0))
-        print_args = (tag.id(), (180 * tag.rotation()) / math.pi)
-        print("Tag Family TAG36H11, Tag ID %d, rotation %f (degrees)" % print_args)
+        img.draw_rectangle(tag.rect, color=(255, 0, 0))
+        img.draw_cross(tag.cx, tag.cy, color=(0, 255, 0))
+        print_args = (tag.family_str, tag.id, (180 * tag.rotation) / math.pi)
```
iabdalkader, line 37: How about `tag.name` ?

iabdalkader (changes requested):

scripts/examples/06-April-Tags/find_apriltags_max_res.py
iabdalkader: Yeah I noticed that too was wondering why its needed. Less code is better anyway.

src/omv/modules/py_image.c
kwagyeman: Doh! Yes, of course.

kwagyeman: Like

src/omv/modules/py_image.c
kwagyeman: Done

kwagyeman:
I went through and tested all the scripts using AprilTags.

iabdalkader (approved):

iabdalkader:
In the future, changes like this should have the breaking label. I'll then collect and add them the stable release log.

PR #2402, opened 2024-08-31
micropython: Update submodule.

iabdalkader:
* Revert optimization levels.
* Allow boards to update TinyUSB CDC buffer sizes.
* Fix libmetal's build rule.
* Fix gc's max_free size.


kwagyeman:
This may have a large impact on modules like py_ml and py_image. There are loops in these modules which copy a lot of data. Can we override the -Os with -O2 for the modules folder via a cflag here: https://github.com/openmv/openmv/blob/master/src/omv/modules/micropython.mk?

iabdalkader:
> This may have a large impact on modules like py_ml and py_image. There are loops in these modules which copy a lot of data. Can we override the -Os with -O2 for the modules folder via a cflag here: https://github.com/openmv/openmv/blob/master/src/omv/modules/micropython.mk?

It may not have that much of an impact given that they all eventually call other C functions that are optimized with the default level. However, ulab, which actually takes most of the space when using O2, may be affected.

iabdalkader:
I will send a follow-up PR to fix the level for C modules.

PR #2403, opened 2024-08-31
lib/tflm: Use GC's free memory for the temporary tensor arena.

iabdalkader:
On some boards, FB memory can become less than the available GC memory for example with 3 frame buffers or a big frame. In this case the first pass allocation (the one used to get the tensor arena's actual size) can fail, even though GC has enough memory to allocate the arena. This patch uses GC's free memory in the first pass to get the arena size.

### DIFF
--- a/src/lib/tflm/tflm_backend.cc
+++ b/src/lib/tflm/tflm_backend.cc
@@ -25,15 +25,16 @@ extern "C" {
 #include "py/objlist.h"
 #include "py/objtuple.h"
 #include "py/binary.h"
+#include "py/gc.h"
 #include "py_ml.h"
-#include "fb_alloc.h"
 
 using namespace tflite;
-#define TF_ARENA_ALIGNMENT  (16 - 1)
+#define TF_ARENA_ALIGN      (16 - 1)
+#define TF_ARENA_ROUND(x)   (((x) + TF_ARENA_ALIGN) & ~(TF_ARENA_ALIGN))
 typedef MicroMutableOpResolver<113> MicroOpsResolver;
 
 typedef struct ml_backend_state {
-    void *arena;
+    uint8_t *arena;
     const Model *model;
     MicroOpsResolver *resolver;
     MicroInterpreter *interpreter;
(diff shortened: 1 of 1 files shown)


src/lib/tflm/tflm_backend.cc
```diff
@@ -192,38 +193,58 @@ static void ml_backend_init_ops_resolver(MicroOpsResolver *resolver) {
...
-    MicroInterpreter interpreter(tflite_model,
-                                 resolver,
-                                 tensor_arena,
-                                 tensor_arena_size);
+    gc_info_t info;
+    gc_info(&info);
+    // Allocate a temporary interpreter to get the optimal arena size.
+    size_t arena_size = info.max_free - TF_ARENA_ALIGN;
+    uint8_t *arena_memory = m_new(uint8_t, arena_size);
+    MicroInterpreter interpreter(tflite_model, resolver, arena_memory, arena_size);
```
kwagyeman, line 211: Don't you need to align the initial tensor arena here before ou pass it?

src/lib/tflm/tflm_backend.cc
```diff
@@ -192,38 +193,58 @@ static void ml_backend_init_ops_resolver(MicroOpsResolver *resolver) {
...
+                                                                           aligned_arena,
+                                                                           arena_size);
+    if (state->interpreter->AllocateTensors() != kTfLiteOk) {
+        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Failed to allocate tensors"));
+    }
+
+    // Initialize the model's state.
+    model->state = state;
+    model->memory_addr = (uint32_t) state->arena;
+    model->memory_size = arena_size;
```
kwagyeman, line 237: The extra 1KB was added originally because the returned value of RAM required for the tensorarea was sometimes less than what was actually needed. This may not be a problem anymore though.

src/lib/tflm/tflm_backend.cc
iabdalkader: Either way works, you can align the pointer or round up the returned size.

src/lib/tflm/tflm_backend.cc
iabdalkader: It was never was needed.

src/lib/tflm/tflm_backend.cc
iabdalkader: I've updated the code to simplify this process. The library seems to align the pointer internally, if needed, so the arena size must be big enough for `size + alignment`. If not, the first pass will fail, as there's not enough GC memory for size + alignment. If it succeeds then the rounded up `used_bytes()` can be used for the second pass with or without an aligned pointer.

src/lib/tflm/tflm_backend.cc
```diff
@@ -192,38 +193,57 @@ static void ml_backend_init_ops_resolver(MicroOpsResolver *resolver) {
...
-    uint8_t *tensor_arena = (uint8_t *) fb_alloc_all(&tensor_arena_size, FB_ALLOC_PREFER_SIZE | FB_ALLOC_CACHE_ALIGN);
-
-    MicroInterpreter interpreter(tflite_model,
-                                 resolver,
-                                 tensor_arena,
-                                 tensor_arena_size);
+    gc_info_t info;
+    gc_info(&info);
+    // Allocate a temporary interpreter to get the optimal arena size.
+    size_t arena_size = info.max_free * MICROPY_BYTES_PER_GC_BLOCK;
```
kwagyeman, line 209: This is the max number of free blocks that can be found in a linear chuck? Or max free fragmented?

kwagyeman: info.max_free seems to be the length of the maximum free block on looking at gc_info().

src/lib/tflm/tflm_backend.cc
kwagyeman: Works great.

src/lib/tflm/tflm_backend.cc
iabdalkader: Currently it's the maximum run of free blocks in any area.

PR #2453, opened 2024-10-12
misc: refactor sensor drivers.

iabdalkader:


kwagyeman (approved):
Quite a refactor.

nathanntg:
Hello! Very appreciative of the OpenMV project, have been having a lot of fun using it.

I noticed recently that some of my old scripts are no longer working, and in searching, found this PR. I am not positive it is directly related to this PR, but I noticed that in the move from `py_sensor.c` to `py_csi.c`, a lot of the ioctl names were changed as well  (e.g., `sensor.IOCTL_LEPTON_SET_MEASUREMENT_MODE` to `sensor.IOCTL_LEPTON_SET_MODE`). Is there a good place to see a full mapping of those changes across releases? 

Thanks!

kwagyeman:
We've started tracking breaking changes in the release notes. These IOCTL names were not actually meant to be changed by Ibrahim. However, the shorter versions improve script readability as I noticed after we added pep8 limiting to the example scripts that many lines were breaking on how long maple of these ioctl names were. So, I decided to keep this change.

As for tracking them in the future. The best thing to do right now is to read the release notes per release. Any breaking changes will be listed. As OpenMV matures we should be able to improve this process.

nathanntg:
Thanks so much! Appreciate the effort and support. (I love the idea of highlighting breaking changes in the release notes, and I might even put in a plug for incorporating breaking API changes into versioning. I was also caught off guard by the rpc removal, so now just rolling back to the older firmware and IDE until I can regroup.)

kwagyeman:
API versioning definitely could be improved. This is something we can work to improve in the future.

As for the RPC removal, cool that you got that code to work for desktop comms. However, it was generating way too many bug reports and structurally not fixable without a large performance loss for PC comms. We recommend to use the debug protocol moving forwards via the scripts in tools.

iabdalkader:
This was supposed to be labeled as a breaking change but it wasn't. I've labeled it now, so it can be found in a search. As for versioning, moving forward I'll increment the major if the API changes.

PR #2465, opened 2024-10-24
sensors/ov5640: Fix PCLK calculation for setting correct exposure time.

RayCali:
MIPI bit mode division forgot that the camera has 2 MIPI lanes hence before dividing the PCLK with the bit mode the bit mode value should be divided by two. Secondly, the pll-predivider value extraction was done wrong, leading to wrong division of pll-predivider.

### DIFF
--- a/src/omv/sensors/ov5640.c
+++ b/src/omv/sensors/ov5640.c
@@ -1128,14 +1128,14 @@ static int calc_pclk_freq(uint8_t sc_pll_ctrl_0,
                           uint8_t sys_root_div) {
     uint32_t pclk_freq = omv_csi_get_xclk_frequency();
     pclk_freq /= ((sc_pll_ctrl_3 & 0x10) != 0x00) ? 2 : 1;
-    pclk_freq /= ((sc_pll_ctrl_0 & 0x0F) == 0x0A) ? 10 : 8;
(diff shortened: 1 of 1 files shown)


RayCali:
> Hi, thanks for the PR.
> 
> Both the title and the commit message need to be in the following format otherwise they won't pass CI tests:
> 
> ```
> sensors/ov5640: Fix something.
> ```

I have changed the title and made a new commit, does that solve it?

iabdalkader:
> does that solve it?

No. I suggest you squash the two commits into one commit, and use the second commit headline. This should fix it.

RayCali:
> > does that solve it?
> 
> No. I suggest you squash the two commits into one commit, and use the second commit headline. This should fix it.

alright, i think it should work now

kwagyeman:
@RayCali - Can you please confirm this fixes the issue that you posted on the forums and that all the exposure calculations make sense? If possible, can you post some before and after results? I just want to record that this fix makes the thing work correctly.

kwagyeman:
Found the info: https://forums.openmv.io/t/ov5640-exposure-time-not-making-sense-with-fps/10105/5?u=kwagyeman Approved

kwagyeman (approved):

Issue #2438, opened 2024-10-02
Wrong alpha passed to draw_image or from draw_image to gpu.


iabdalkader:
In so many places, alpha is passed as 256, when it should be 0...255:
```C
imlib/draw.c:            imlib_draw_image(&t_src_img, src_img, 0, 0, x_scale, y_scale, roi,
imlib/draw.c-                             -1, 256, NULL, NULL,
imlib/draw.c-                             hint & (IMAGE_HINT_AREA | IMAGE_HINT_BILINEAR | IMAGE_HINT_BICUBIC),
imlib/draw.c-                             NULL, NULL, NULL);
```

kwagyeman:
@iabdalkader - This is correct. The -1 is the RGB channel extraction. If it were 0 then that would cause red channel extraction.

iabdalkader:
> @iabdalkader - This is correct. The -1 is the RGB channel extraction. If it were 0 then that would cause red channel extraction.

`-1, 256,` 256 should be 255... Alpha is an int from 0-> 255 0%->100%

kwagyeman:
Yes, we should fix this to make it industry standard.

kwagyeman:
https://github.com/openmv/openmv/pull/2486

PR #2486, opened 2024-11-07
imlib: Update draw_image alpha to 0-255 from 0-256.

kwagyeman:
Note that this is a breaking change. Anyone passing an alpha of 256 will get an error message.

Fixes: https://github.com/openmv/openmv/issues/2438

### DIFF
--- a/src/omv/imlib/draw.c
+++ b/src/omv/imlib/draw.c
@@ -722,6 +722,9 @@ void imlib_draw_row_setup(imlib_draw_row_data_t *data) {
 
     int alpha = data->alpha, max = 256;
 
+    // To avoid having to divide by 255 scale alpha to 0-256 so we can right shift by 8.
+    alpha = fast_roundf((alpha * 256) / 255.0f);
+
     if (data->dst_img->pixfmt == PIXFORMAT_RGB565) {
         alpha >>= 3; // 5-bit alpha for RGB565
         max = 32;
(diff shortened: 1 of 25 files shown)


src/omv/imlib/draw.c
```diff
@@ -722,6 +722,9 @@ void imlib_draw_row_setup(imlib_draw_row_data_t *data) {
 
     int alpha = data->alpha, max = 256;
 
+    // To avoid having to divide by 255 scale alpha to 0-256 so we can right shift by 8.
```
iabdalkader, line 725: This doesn't look right. You're still dividing by 255, so where's the right shift by 8 ? Also `smuad_alpha` will be `0x1000000` if alpha == 255.

kwagyeman: It's used like this:

`__SMUAD(smuad_alpha, smuad_pixel) >> 8`

Where `smuad_alpha` is `((alpha << 16) | (256 - alpha))`

It performs `(alpha * src_pixel) + ((256 - alpha) * dst_pixel) / 256`.

Alpha of 255 needs to produce 100% of the source pixel. And alpha of 0 needs to be 100% of the destination pixel.

kwagyeman: It's possible to divide by 255 instead of 256, but, this can't be done in a single shift. But, it's more work per pixel. So, by doing `alpha = fast_roundf((alpha * 256) / 255.0f);` you save a lot of time per pixel.

iabdalkader (approved):

Issue #2469, opened 2024-10-30
Possible memory corruption/allocation error with image.Image(ndarray)


Jarvis-X:
**Describe the bug**
As described in [this post](https://forums.openmv.io/t/memory-corruption-with-image-image-ndarray/10146). 

**To Reproduce**
- The full program (first the desired behavior, then change one line of code to reproduce the errors)
  - Download and execute [this program](https://github.com/LehighBlimpGroup/Blob-detection-and-Tracking/blob/main/perception_subsystem.py) on either a NiclaVision (recommended for reproducing the errors sooner and more easily) or an OpenMV Cam RT1062. 
  - You might need to tweak the RGB gains on [line 49 to 52](https://github.com/LehighBlimpGroup/Blob-detection-and-Tracking/blob/main/perception_subsystem.py#L49-L52) a bit so that the image looks more naturally colored. 
    - To do that, on OpenMV Cam RT1062, point the camera in the environment and run the program, then copy the lone 3-element tuple from the console to [line 50](https://github.com/LehighBlimpGroup/Blob-detection-and-Tracking/blob/main/perception_subsystem.py#L50)
    - On NiclaVision, run [this program](https://github.com/LehighBlimpGroup/Blob-detection-and-Tracking/blob/main/get_gains.py) once and copy the values in the console after "RGB gain:" to [line 52](https://github.com/LehighBlimpGroup/Blob-detection-and-Tracking/blob/main/perception_subsystem.py#L52)
  - Point the camera towards a color wheel where purple or green or red colors are presented until cells of the corresponding colors show up in the image buffer. The LED should change colors/blink/dim. Keep the program running for as long as you want. It should not crash or throw any error.
  - *So far, we consider the behaviors "stable" as it is NOT using image.Image(ndarray) methods that I am suspecting to be buggy.*
  - Now, go to [line 449](https://github.com/LehighBlimpGroup/Blob-detection-and-Tracking/blob/main/perception_subsystem.py#L449) and replace `_, _, max_val, same_val_rc = self.optimized_loop(metric)` with `_, _, max_val, same_val_rc = self.kernel_count(metric)`
  - Rerun the program for ~2 mins max. We should see the program crash with the errors mentioned in the original post.
[...]

PR #2487, opened 2024-11-07
py_image: Fix corruption using to_ndarray() on grayscale images.

kwagyeman:
Maybe fixes: https://github.com/openmv/openmv/issues/2469

This would corrupt the trailing 1-3 bytes of an allocation. However, allocations are multiples of 4 bytes. So, while this is a fix, I am not sure if it fixes the user error as the corrupted bytes would be at the tail of a heap block not in use.


iabdalkader:
Do you need to process odd bytes or is it always a multiple of 4 ? Maybe round down ?

Issue #2357, opened 2024-08-05
get_state drops text on the STM32


kwagyeman:
I noticed that a lot of lines of text are missing on the STM32 with get_state.

Pretty easy to replicate with master:

```
import sensor, image, time

sensor.reset()
sensor.set_pixformat(sensor.RGB565)
sensor.set_framesize(sensor.QVGA)
sensor.skip_frames(time = 2000)

clock = time.clock()

while(True):
    clock.tick()
    img = sensor.snapshot()
    print("hello world")
    print(clock.fps())
```

Gives:

```
hello world
23.3051
hello world
23.301
hello world
23.2975
hello world
23.2945
hello world
23.3236

Traceback (most recent call last):
  File "<stdin>", line 16, in <module>
OpenMV v4.5.8-43.g0ad4d34a.dirty; MicroPython v1.22-omv.r17.434.g8f6a976de; OPENMV4P-STM32H743
Type "help()" for more information.
>>> 
```

With the old polling. With get_state() I get:

```
23.3029
hello world
23.301
hello world
23.2992
hello world
23.3161
Tracebll last):
  File "<stdin>",e>
MPYMV v4.5.8-43.g0ad4d34a.dirty-omv.r17.434.g8f6a976de; OPEpe "help()" for more informal
```

The same on a tiny USB board produces:

```
46.7706
hello world
46.70913
hello world
46.74797
hello world
46.69261
hello world
46.72897

Traceback (most recent call last):
  File "<stdin>", line 14, in <module>
OpenMV v4.5.8-43.g0ad4d34a.dirty; MicroPython v1.22-omv.r17.434.g8f6a976de; OpenMV IMXRT1060-MIMXRT1062DVJ6A
Type "help()" for more information.
```

Need to look into how to fix. The change related to breaking on interrupts being disabled shouldn't affect this as interrupts are on.

iabdalkader:
It's not a bug, it drops text on all cams. The TX buffer eventually overflows now that text is not polled separately. I can send more text in the same packet, up to 512 if the IDE can handle variable length string.

kwagyeman:
The text isn't destroyed with GetState using TinyUSB or not using GetState with the STM32. Seems like it's just related to the STM32 using GetState. Even with a polling rate of 1KHz very small strings are eaten. Something is off.

iabdalkader:
The STM32 clears CDC buffers on soft-reboot, for some reason which I forgot.

iabdalkader:
[...]

PR #2493, opened 2024-11-10
misc: Update get_state() to work with a variable packet size.

kwagyeman:
Updates get_state() return a packet equal to the EP size that the IDE knows. This significantly helps with text buffer transfer on HS cameras. With this fix, it's very hard to trivially overflow the text buffer and get chopped-up text unless you are printing very large data structures. Results show that 25ms get_state() polling is good and does not result in any dropped text on high-speed cameras.

Tested in IDE commits past v4.4.4 release.

I verified that the IDE works with cameras without this change, FS cameras with this change, and HS cameras with this change.

Fixes: https://github.com/openmv/openmv/issues/2357

### DIFF
--- a/src/omv/common/usbdbg.c
+++ b/src/omv/common/usbdbg.c
@@ -227,7 +227,12 @@ void usbdbg_data_in(uint32_t size, usbdbg_write_callback_t write_callback) {
         }
 
         case USBDBG_GET_STATE: {
-            uint32_t buffer[16] = { 0 };
+            // IDE will request 63/511 bytes of data to avoid ZLP packets.
+            // 64/512 packets still work too but generate ZLP packets.
+            size = OMV_MIN(size, 512);
+            uint8_t byte_buffer[size];
+            memset(byte_buffer, 0, size);
+            uint32_t *buffer = (uint32_t *) byte_buffer;
             static uint32_t last_update_ms = 0;
 
             // Set script running flag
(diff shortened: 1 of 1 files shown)


iabdalkader:

https://forums.openmv.io/t/openmv-rt1062-hangs-frequently-only-when-connected-to-ide/10416/3

iabdalkader:

https://forums.openmv.io/t/openmv-rt1062-hangs-frequently-only-when-connected-to-ide/10416/4

src/omv/common/usbdbg.c
```diff
@@ -227,7 +227,11 @@ void usbdbg_data_in(uint32_t size, usbdbg_write_callback_t write_callback) {
         }
 
         case USBDBG_GET_STATE: {
-            uint32_t buffer[16] = { 0 };
+            // IDE will request 63/511 bytes of data to avoid ZLP packets.
+            // 64/512 packets still work too but generate ZLP packets.
+            uint8_t byte_buffer[size];
```
kwagyeman, line 232: Make sure this doesn't overflow

kwagyeman: Done

kwagyeman:
@iabdalkader - Added something to prevent the stack from overflowing and retested with 63/511 byte packets.

iabdalkader (approved):

PR #2506, opened 2024-11-19
_boot.py: write unique_id to .openmv_disk.

kidswong999:
When two OpenMV Cams are connected to the computer, OpenMV IDE cannot know the exact path.

After adding uniqueid to .openmv_disk file.

OpenMV IDE first gets the uniqueid through the serial port, and then can determine whether the path matches the connected OpenMV Cam.

### DIFF
--- a/scripts/libraries/_boot.py
+++ b/scripts/libraries/_boot.py
@@ -130,4 +130,8 @@ try:
 except Exception:
     create_file(".openmv_disk")
 
-del os, sys, vfs, fat, bdev, sdcard
+with open(".openmv_disk", "w") as f:
+    import machine
+    f.write("".join(f'{byte:02X}' for byte in machine.unique_id()))
+
+del os, sys, vfs, fat, bdev, sdcard, machine


iabdalkader:
The new DFU bootloader uses the unique ID for the USB serial string. The IDE can use that to identify the board.

kwagyeman:
Hi, I'll see what I can do. Given Ibrahim's information above I should be able to fix this for linux. Not sure if I have the information available for windows/mac. I can knock this issue out quickly.

kidswong999:
This PR is useless, we found a better way. I will close this PR. 

This command works perfect in Powershell:
`Get-Disk | Where-Object {$_.SerialNumber -eq "396C39613532"} | Get-Partition | Select-Object -ExpandProperty DriveLetter`

kwagyeman:
I'll have the feature in a dev IDE before the end of the week.

kwagyeman:
This is now fixed on windows in the IDE. Adding linux and then mac support next.

PR #2508, opened 2024-11-22
imlib/draw: Fixed transpose on SDRAM-less boards.

kwagyeman:
When the omv_gpu driver was added for the STM32 it allocates a CLUT using fb_alloc. However, during transpose we allocate all RAM. So, the GPU driver call fails to allocate the RAM it needs for loading it's CLUT.

On SDRAM boards this error doesn't happen the fb_alloc_all call will only exhaust onchip SRAM.

### DIFF
--- a/src/omv/imlib/draw.c
+++ b/src/omv/imlib/draw.c
@@ -3144,8 +3144,14 @@ void imlib_draw_image(image_t *dst_img,
             t_src_img.data = src_img->data;
         }
 
+        // Query available on-chip RAM.
         uint32_t size;
         void *data = fb_alloc_all(&size, FB_ALLOC_PREFER_SPEED | FB_ALLOC_CACHE_ALIGN);
+        fb_free();
+
+        // Allocate a buffer to hold chunks of the transposed image while not using all of the on-chip RAM.
+        size = IM_MIN(size, image_size(&t_src_img));
+        data = fb_alloc(size, FB_ALLOC_PREFER_SPEED | FB_ALLOC_CACHE_ALIGN);
 
         // line_num stores how many lines we can do at a time with on-chip RAM.
         image_t temp = {.w = t_roi.w, .h = t_roi.h, .pixfmt = t_src_img.pixfmt};


iabdalkader:

https://forums.openmv.io/t/img-replace-out-of-fast-frame-buffer-stack-memory/10235/2

src/omv/imlib/draw.c
```diff
@@ -3144,8 +3144,17 @@ void imlib_draw_image(image_t *dst_img,
             t_src_img.data = src_img->data;
         }
 
+        // Query available on-chip RAM.
         uint32_t size;
         void *data = fb_alloc_all(&size, FB_ALLOC_PREFER_SPEED | FB_ALLOC_CACHE_ALIGN);
```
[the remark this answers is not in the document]
kwagyeman: Done

kwagyeman:
Fixed with a robust enough solution. The amount of RAM dedicated to transposing is limited to the transposed image size.

I tested transposing QVGA RGB565/GRAYSCALE images on the H7/F7, and things worked without issue.

src/omv/imlib/draw.c
```diff
@@ -3144,8 +3144,14 @@ void imlib_draw_image(image_t *dst_img,
...
+        fb_free();
+
+        // Allocate a buffer to hold chunks of the transposed image while not using all of the on-chip RAM.
+        size = IM_MIN(size, image_size(&t_src_img));
```
iabdalkader, line 3153: If the available memory is less than the image size, then this will fail any way right ? So why the query, free, alloc, only to fail ?

kwagyeman: It can work with less memory. It does the job faster the more memory it has.

src/omv/imlib/draw.c
```diff
@@ -3144,8 +3144,14 @@ void imlib_draw_image(image_t *dst_img,
             t_src_img.data = src_img->data;
         }
 
+        // Query available on-chip RAM.
         uint32_t size;
         void *data = fb_alloc_all(&size, FB_ALLOC_PREFER_SPEED | FB_ALLOC_CACHE_ALIGN);
```
iabdalkader, line 3149: Perhaps we should add an fb_alloc_available() function.

kwagyeman: https://github.com/openmv/openmv/issues/2522

iabdalkader: How does the above query on chip RAM ? If there isn't enough it will just allocate from DRAM and return the size.

kwagyeman: FB_ALLOC_PREFER_SPEED it will allocate to overlay stack.

PR #2518, opened 2024-11-26
sensors/genx320: Enable histogram mode.

kwagyeman:
Histogram mode is now working. Whatever FPS you want with low CPU utilization unlocked.


iabdalkader:
Please keep the examples as they were, we don't need a new directory per sensor.

kwagyeman:
This PR is ready. Sensor works as expected and flickering is now completely mitigated.

kwagyeman:
https://github.com/user-attachments/assets/ea4f8029-2dbb-4946-bff7-13ce7fac6e3b

PR #2521, opened 2024-11-26
scripts/libraries: Fix yolo breaking on one element.

kwagyeman:
Need to get support for keepdims added. Otherwise, we'll have to add stuff like this for any function that can output a scalar instead of an ndarray when it returns a single result.


kwagyeman:
Feature request for `keepdims` support: https://github.com/v923z/micropython-ulab/issues/697

iabdalkader:
Tested ?

kwagyeman:
Yeah, this is the only way to fix it without keepdims. But, it's super ugly.

PR #2523, opened 2024-11-28
scripts/examples: Fix new refactored IOCTL names.

kwagyeman:
Not sure when these were renamed. https://github.com/openmv/openmv/commit/62708eb38bd0aadc2851f1a3b51655ed612b24d1?


iabdalkader:
I didn't actually mean to rename the QSTRs, just the enums, but the shorter new ones are better.

Issue #2522, opened 2024-11-26
Add flag to fb_alloc_available()


kwagyeman:
Support querying onchip ram to clean this up: https://github.com/openmv/openmv/blob/master/src/omv/imlib/draw.c#L3149

kwagyeman:
Not relevant anymore.

PR #2534, opened 2024-11-30
alloc/fb_alloc: Add hint support to fb_avail.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2522


iabdalkader:
Actually I wanted to refactor this some more. Isn't `FB_ALLOC_PREFER_SIZE the same as`!FB_ALLOC_PREFER_SPEED` or no flags at all?  Also the flags should be `FB_ALLOC_FLAGS_FAST/ALIGNED`

kwagyeman:
> Isn't FB_ALLOC_PREFER_SIZE the same as !FB_ALLOC_PREFER_SPEED or no flags at all?

fb_alloc_all() defaults to preferring speed unless told. fb_alloc() will put the alloc in the overlay if it fits unless you specify prefer size... which there's no reason to do since it has to give you the amount you requested or fail. However, this technically allows you to steer allocs out of the overlay if you needed to. But, that's brittle API behavior. Should remove it. After doing this then I think your above statement it correct.

> Also the flags should be FB_ALLOC_FLAGS_FAST/ALIGN

Sure, will change along with simpyling the flag usage.

kwagyeman:
Slight change to the above, all allocs should just be cache aligned in general. There's no reason to pass that flag when it should be the default.

iabdalkader:
> all allocs should just be cache aligned in genera

This might be a bit wasteful for small allocs.

> Refactored it some more to reduce all the no hints stuff being passed everywhere

I don't think it's a good idea to remove the flags, they future-proof it.

kwagyeman:
> This might be a bit wasteful for small allocs.

Yeah, I have a better idea. I mainly only need this for images. Should just make sure those are covered.

> I don't think it's a good idea to remove the flags, they future-proof it.

I was thinking the same thing when removing them. Would be a pain to add back if we need them again.

...

Okay, updating this PR.

kwagyeman:
@iabdalkader - Removed all the excess commits from this PR except for the original reason.

Tested that TRANSPOSE still works fine on cameras with SDRAM and without.

src/omv/imlib/agast.c
```diff
@@ -215,7 +215,7 @@ static corner_t *agast58_detect(image_t *img, int b, int* num_corners, rectangle
 	width=s_width;
 
     // Try to alloc MAX_CORNERS or the actual max corners we can alloc.
-    int max_corners = IM_MIN(MAX_CORNERS, (fb_avail() / sizeof(corner_t)));
+    int max_corners = IM_MIN(MAX_CORNERS, (fb_avail(FB_ALLOC_NO_HINT) / sizeof(corner_t)));
```
iabdalkader, line 218:
```suggestion
    int max_corners = IM_MIN(MAX_CORNERS, (fb_avail(0) / sizeof(corner_t)));
```
No flags is just `0`.

src/omv/alloc/fb_alloc.h
```diff
@@ -33,7 +33,7 @@
 char *fb_alloc_stack_pointer();
 void fb_alloc_fail();
 void fb_alloc_init0();
-uint32_t fb_avail();
+uint32_t fb_avail(int hints);
```
iabdalkader, line 36:
```suggestion
uint32_t fb_avail(int flags);
```

Also would like to rename those:
```c
enum {
    FB_ALLOC_FLAGS_FASTMEM = (1 << 0),
    FB_ALLOC_FLAGS_ALIGNED = (1 << 1),
} fb_alloc_flags_t;
```
If it's not fast, then it's size.

iabdalkader (changes requested):

src/omv/alloc/fb_alloc.h
kwagyeman: Done

src/omv/imlib/agast.c
kwagyeman: Done

src/omv/alloc/fb_alloc.h
```diff
@@ -26,22 +26,28 @@
...
-#define FB_ALLOC_PREFER_SPEED    1
-#define FB_ALLOC_PREFER_SIZE     2
-#define FB_ALLOC_CACHE_ALIGN     4
+
+typedef enum fb_alloc_flags {
+    // Cache aligns the allocation so you can use cache maintenance instructions on it.
+    // Use this with DMA peripherals.
+    FB_ALLOC_FLAGS_ALIGNED = (1 << 0),
+    // Force the allocation to be external to save the faster internal memory.
+    FB_ALLOC_FLAGS_EXTERNAL = (1 << 1),
```
iabdalkader, line 35:
```suggestion
    // Allocates cache-aligned memory. 
    FB_ALLOC_FLAGS_ALIGNED = (1 << 0),  
    // Allocates external memory to preserve faster internal memory.  
    FB_ALLOC_FLAGS_EXTERNAL = (1 << 1),  
```

iabdalkader:
I'm reviewing this and all good so far. However, I think we actually need `FB_ALLOC_FLAGS_FASTMEM` instead, and we should reconsider the behavior of fb_alloc/all.

I think in most (if not all) cases more slow-memory is better than less fast-memory. If you need fast memory, even if it's less, then in this case you specify `FASTMEM`. So `fb_alloc_avail` with no speed flag should return the max available memory (minus alignment if any).

Let's consider all use cases:

| Size/Speed | Absolutely need fast memory      | Fast memory would be nice | Don't Care        |
|------------|----------------------------------|---------------------------|-------------------|
| Don't care | avail(FASTMEM)/alloc(FASTMEM)    | avail(0)/alloc(FASTMEM)   | avail(0)/alloc(0) |
| Exact size | avail(FASTMEM)/alloc(FASTMEM)[1] | avail(0)/alloc(FASTMEM)   | avail(0)/alloc(0) |

*[1] Should check size returned form `fb_avail()` and fail if it's not enough.*

The draw.c case. Let's assume less fast-memory here is more useful than more slow-memory *regardless of how much that fast memory is*:

```C
// Will return the available fast memory.
uint32_t size = fb_alloc_avail(FB_ALLOC_FLAGS_FASTMEM);
size = MIN(size, image_size(....
// Will allocate up to the max available fast memory ().
void *data = fb_alloc(size, FB_ALLOC_FLAGS_FASTMEM);
```

Compared this to the corners case. Here I would prefer to alloc all corners in fast memory, if possible, if not then all corners in DRAM ( I don't want less fast-memory):
```C
 // Returns the max available memory (say 500KBs DRAM)
int max_corners = IM_MIN(MAX_CORNERS, (fb_alloc_avail(0) / sizeof(corner_t)));
 // Try to allocate 150KBs, from fast memory, will fail then allocate from the 500KBs.
corner_t *corners = (corner_t*) fb_alloc(max_corners * sizeof(corner_t), FB_ALLOC_FLAGS_FASTMEM);
```

Does this make sense?

PR #2545, opened 2024-12-09
sensos/boson: Add initial support for the FLIR Boson.

kwagyeman:
This is basic driver for the FLIR Boson with some basic color tracking examples.

The sensor is operated in 8-bit AGC mode. More advanced functionality is currently not available until I can update my BOSON using the USB interface (parts on order now to do this). The FLIR Boson I received is currently running a firmware version older than the 4.X release on the FLIR Boson+. As such, I cannot set it into RGB565 muxed onto 8-bits yet and 16-bit RAW temp mode muxed onto 8-bits. It does currently support YUV422 output muxed onto 8-bits, but, the frame rate drops by half when this mode is used because the clock rate doesn't increase. The 4.X firmware release adds support for these modes and increasing the clock frequency which should allow me to improve support to the level of the FLIR Lepton Driver.

In any case, the driver is good enough right now though for demoing and having object tracking running on the data.

https://youtu.be/SbldYSyMPYE?si=LAhI9HPdiv2A3M4g

Driver and adapter module have been tested and work on the H7 Plus and RT1062. I only enabled it for the RT1062, though, given the dwindling FLASH on the H7 Plus.


src/Makefile
```diff
@@ -56,6 +56,7 @@ CUBEAI_DIR=stm32cubeai
 CMSIS_DIR=hal/cmsis
 MICROPY_DIR=micropython
 GENX320_DIR=drivers/genx320
+BOSON_DIR=drivers/boson
```
iabdalkader, line 59: You should also export this in `omv/common/export.mk`

iabdalkader (changes requested):

src/Makefile
kwagyeman: Done

iabdalkader (approved):

iabdalkader:

https://forums.openmv.io/t/openmv-h7-plus-adv7280a-video-decoder/10809/4

PR #2555, opened 2024-12-21
sensors/pag7936: Add missing control functions.

kwagyeman:
This PR adds a bunch of missing things to the PAG7936 driver including auto exposure and auto gain control support.

I tried changing around the default exposure/gain settings but got mixed results. Trying to increase the default brightness target just results in the gain going to the max and it blows out any bright area in the image.

I also tried enabling the denoise and edge enhancement feature of the camera sensor. However, this feature causes border artifacts as it doesn't appear to modify the pixels around the edge of the image. When enabled you can see a box border around the image appear when there are large brightness changes.

...

Regarding the triggering commit. I implemented all the steps per the datasheet. However, the steps the datasheet calls out result in the camera outputting frames without needing to receive a hardware or software trigger. So, it doesn't appear to actually work given our most recent information. That said, the cameras fps output behavior does update when triggered mode is enabled. So, it's just missing something unknown.

...

Regarding increasing the brightness for 480 FPS. There's not much that can be done here. At that frame rate the sensor will already max out the exposure and auto gain. If the image is still dark then the processor will have to increase the brightness of the image or the scene will need more illumination.

Exposure 2080us with max analog gain (simulating 480 FPS):

image

Not much that can be done on the sensor side here.


iabdalkader:
This is waiting on PixArt, right?

kwagyeman:
The autogain/exposure commit can be improved with some feedback from PixArt. But, it's functional.

The triggering commit doesn't appear to work even though it's coded per the datasheet.

Waiting on responses to these questions from PixArt. I think it makes more sense to merge now and then send updates when they reply as it could be a while.

kwagyeman:
@iabdalkader - This is ready to go per the email from Pixart. I've moved the triggering to another PR until they reply with how to fix it.

PR #2556, opened 2024-12-21
imlib/draw: Enable GPU in-place cut-through.

kwagyeman:
Tested on the STM32H7. In-place image cropping performance with RGB565 (only operation in-place DMA2D supports) sees a massive boost going from 21ms to 2ms as the copy on the image is removed.

Tested on Alif, all instances of scaling down and/or cropping at the same time work. The copy overhead has been removed.

Scaling up on Alif seems to have issues. Without this PR with and with out-of-place buffers you get a corrupted image.

### DIFF
--- a/src/omv/imlib/draw.c
+++ b/src/omv/imlib/draw.c
@@ -3040,9 +3040,6 @@ void imlib_draw_image(image_t *dst_img,
         color_palette = NULL;
     }
 
-    // Special destination?
-    bool is_jpeg = src_img->pixfmt == PIXFORMAT_JPEG;
-    bool is_png = src_img->pixfmt == PIXFORMAT_PNG;
     // Best format to convert yuv/bayer/jpeg image to.
     int new_not_mutable_pixfmt = (rgb_channel != -1) ? PIXFORMAT_RGB565 :
                                  (color_palette ? PIXFORMAT_GRAYSCALE :
@@ -3067,13 +3064,6 @@ void imlib_draw_image(image_t *dst_img,
 
     // Is the line length growing which will prevent us from working in-place?
     bool is_upscaling = src_img_row_bytes < dst_img_row_bytes;
-
-    // Force a deep copy if we cannot use the image in-place.
-    bool need_deep_copy = (dst_img->data == src_img->data)
-                          && (is_scaling || is_upscaling || is_bayer_yuv_conversion);
-
-    // Force a deep copy if we are scaling.
-    bool is_bayer_yuv_conversion_scaling = is_bayer_yuv_conversion && is_scaling;
 
     #if (OMV_GPU_ENABLE == 1)
     if (!callback &&
@@ -3109,7 +3099,10 @@ void imlib_draw_image(image_t *dst_img,
     #endif
 
     // Make a deep copy of the source image.
-    if (need_deep_copy || is_bayer_yuv_conversion_scaling || is_jpeg || is_png) {
+    if (((dst_img->data == src_img->data) &&
+         (is_scaling || is_upscaling || is_bayer_yuv_conversion)) ||
+        (is_bayer_yuv_conversion && is_scaling) ||
+        src_img->is_compressed) {
         new_src_img.w = src_img->w; // same width as source image
         new_src_img.h = src_img->h; // same height as source image
 
@@ -3126,9 +3119,9 @@ void imlib_draw_image(image_t *dst_img,
                         imlib_debayer_image(&new_src_img, src_img);
                     } else if (src_img->is_yuv) {
                         imlib_deyuv_image(&new_src_img, src_img);
-                    } else if (is_jpeg) {
+                    } else if (src_img->pixfmt == PIXFORMAT_JPEG) {
                         jpeg_decompress(&new_src_img, src_img);
-                    } else if (is_png) {
+                    } else if (src_img->pixfmt == PIXFORMAT_PNG) {
                         png_decompress(&new_src_img, src_img);
                     }
                     break;
@@ -3139,7 +3132,7 @@ void imlib_draw_image(image_t *dst_img,
                     break;
                 }
                 default: {
-                    if (is_png) {
+                    if (src_img->pixfmt == PIXFORMAT_PNG) {
                         png_decompress(&new_src_img, src_img);
                     }
                     break;
--- a/src/omv/imlib/draw.c
+++ b/src/omv/imlib/draw.c
@@ -3061,19 +3061,19 @@ void imlib_draw_image(image_t *dst_img,
     size_t dst_img_row_bytes = image_size(dst_img) / dst_img->h;
 
     // Do we need to convert the image?
-    bool is_bayer_color_conversion = src_img->is_bayer && !dst_img->is_bayer;
-    bool is_yuv_color_conversion = src_img->is_yuv && !dst_img->is_yuv;
-    bool is_color_conversion = is_bayer_color_conversion || is_yuv_color_conversion;
+    bool is_bayer_conversion = src_img->is_bayer && !dst_img->is_bayer;
+    bool is_yuv_conversion = src_img->is_yuv && !dst_img->is_yuv;
+    bool is_bayer_yuv_conversion = is_bayer_conversion || is_yuv_conversion;
 
     // Force a deep copy if we cannot use the image in-place.
     bool need_deep_copy = (dst_img->data == src_img->data)
-                          && (is_scaling || (src_img_row_bytes < dst_img_row_bytes) || is_color_conversion);
+                          && (is_scaling || (src_img_row_bytes < dst_img_row_bytes) || is_bayer_yuv_conversion);
 
     // Force a deep copy if we are scaling.
-    bool is_color_conversion_scaling = is_color_conversion && is_scaling;
+    bool is_bayer_yuv_conversion_scaling = is_bayer_yuv_conversion && is_scaling;
 
     // Make a deep copy of the source image.
-    if (need_deep_copy || is_color_conversion_scaling || is_jpeg || is_png) {
+    if (need_deep_copy || is_bayer_yuv_conversion_scaling || is_jpeg || is_png) {
         new_src_img.w = src_img->w; // same width as source image
         new_src_img.h = src_img->h; // same height as source image
 
--- a/src/omv/imlib/draw.c
+++ b/src/omv/imlib/draw.c
@@ -927,46 +927,6 @@ void imlib_draw_row(int x_start, int x_end, int y_row, imlib_draw_row_data_t *da
                                 }
                             }
                         }
-                        // More desirable results are produced by alpha blending with 8-bits.
-                        // if (!data->color_palette) {
-                        //     for (int x = x_start; x < x_end; x++) {
-                        //         int pixel = IMAGE_GET_BINARY_PIXEL_FAST(src32, x) & IMAGE_GET_BINARY_PIXEL_FAST(dst32, x);
-                        //         IMAGE_PUT_BINARY_PIXEL_FAST(dst32, x, pixel);
-                        //     }
-                        // } else {
-                        //     const uint16_t *color_palette = data->color_palette;
-                        //     uint16_t pal0 = color_palette[0], pal255 = color_palette[255];
-                        //     pal0 = COLOR_RGB565_TO_Y(pal0) > 127;
-                        //     pal255 = COLOR_RGB565_TO_Y(pal255) > 127;
-                        //     switch ((pal0 << 1) | (pal255 << 0)) {
-                        //         case 0: {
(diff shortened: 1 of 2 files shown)


src/omv/ports/stm32/omv_gpu.c
```diff
@@ -167,10 +167,14 @@ int omv_gpu_draw_image(image_t *src_img,
...
-        SCB_CleanInvalidateDCache_by_Addr(dst16_tmp, dst_rect->w * sizeof(uint16_t));
-        dst16_tmp += dst_img->w;
+    if (dst_img->w == dst_rect->w) {
+        SCB_CleanInvalidateDCache_by_Addr(dst16, dst_rect->w * dst_rect->h * sizeof(uint16_t));
+    } else {
+        uint16_t *dst16_tmp = dst16;
+        for (int i = 0; i < dst_rect->h; i++) {
+            SCB_CleanInvalidateDCache_by_Addr(dst16_tmp, dst_rect->w * sizeof(uint16_t));
+            dst16_tmp += dst_img->w;
+        }
```
iabdalkader, line 177: Why was line-by-line invalidate ever needed?

src/omv/imlib/draw.c
```diff
@@ -3072,6 +3072,36 @@ void imlib_draw_image(image_t *dst_img,
     // Force a deep copy if we are scaling.
     bool is_color_conversion_scaling = is_color_conversion && is_scaling;
 
+    #if (OMV_GPU_ENABLE == 1)
+    bool gpu_in_place_op = (dst_img->data == src_img->data)
+                           && (src_img_row_bytes >= dst_img_row_bytes) && (!is_color_conversion);
+
+    bool runnable_on_gpu = (rgb_channel < 0) && (!(hint & (IMAGE_HINT_AREA | IMAGE_HINT_BICUBIC))) && (!callback);
+
```
iabdalkader, line 3080:
```suggestion
```

src/omv/imlib/draw.c
```diff
@@ -3072,6 +3072,36 @@ void imlib_draw_image(image_t *dst_img,
...
+    gpu_hint |= (dst_delta_y < 0) ? IMAGE_HINT_VFLIP : 0;
+
+    // Try to offload this to the GPU for processing before allocating a copy of the image.
+    if (gpu_in_place_op && (!is_jpeg) && (!is_png) && (!(hint & IMAGE_HINT_TRANSPOSE)) &&
+        runnable_on_gpu && (!omv_gpu_draw_image(src_img, &src_rect, dst_img, &dst_rect,
+                                                alpha, color_palette, alpha_palette, gpu_hint))) {
+        goto exit_cleanup;
+    }
+    #endif
+
```
iabdalkader, line 3104:
```suggestion
    #if (OMV_GPU_ENABLE == 1)
    bool scale_on_gpu = !callback &&
                        !is_color_conversion &&
                        !src_img->is_compressed &&
                        (rgb_channel < 0) &&
                        (src_img_row_bytes >= dst_img_row_bytes) &&
                        (dst_img->data == src_img->data) && // TODO why not scale on GPU if src!=dst
                        !(hint & (IMAGE_HINT_AREA | IMAGE_HINT_BICUBIC | IMAGE_HINT_TRANSPOSE));
    if (scale_on_gpu) {
        rectangle_t dst_rect = {
            .x = dst_x_start,
            .y = dst_y_start,
            .w = dst_x_end - dst_rect.x,
            .h = dst_y_end - dst_rect.y,
        };

        rectangle_t src_rect = {
            .x = src_x_accum_reset >> 16,
            .y = src_y_accum_reset >> 16,
            .w = ((src_x_accum_reset + (src_x_frac * dst_rect.w)) >> 16) - src_rect.x,
            .h = ((src_y_accum_reset + (src_y_frac * dst_rect.h)) >> 16) - src_rect.y,
        };

        image_hint_t gpu_hints = ((dst_delta_x < 0) ? IMAGE_HINT_HMIRROR : 0) |
                                 ((dst_delta_y < 0) ? IMAGE_HINT_VFLIP : 0) |
                                 (hint & (IMAGE_HINT_BILINEAR | IMAGE_HINT_BLACK_BACKGROUND));

        // Try to offload this to the GPU for processing before allocating a copy of the image.
        if (omv_gpu_draw_image(src_img, &src_rect, dst_img, &dst_rect,
                               alpha, color_palette, alpha_palette, gpu_hints)) {
            goto exit_cleanup;
        }
    }
    #endif

```

iabdalkader (changes requested):

src/omv/ports/stm32/omv_gpu.c
kwagyeman: Yes, any operations involving cropping the middle of an image and/or drawing the middle of another image will be faster.

src/omv/imlib/draw.c
kwagyeman: I need the variables for the second gpu call which comes later in the code. The second call handles if src!=dst.

src/omv/ports/stm32/omv_gpu.c
iabdalkader: And every other normal use case will be slower.

src/omv/imlib/draw.c
iabdalkader: Why do you need a second call? What's special about `src!=dst`? The GPU can scale either way.

kwagyeman: The second one handles if you are scaling after any type of color conversion or etc. which requires allocating a second image.

Yeah, I I guess though you can make the first one intercept the trivial out of place case. You don't need to wait till the second one to catch that.

iabdalkader: > if you are scaling after any type of color conversion or etc. which requires allocating a second image.

But scaling with color conversion is supported on both GPUs. Like?
```
img.to_grayscale(x_scale=0.5, y_scale=0.5, hint=image.BILINEAR)
```
No debayer though of course.

iabdalkader: Also if this is `(src_img_row_bytes >= dst_img_row_bytes)` to ensure that it's down-scaling because of that bug, it should either be moved to the GPU driver (if it can't handle it) or fixed.

kwagyeman: Here's a better explanation of what I was thinking when making this change. I was just adding in the functionality to cover making it work to handle the in-place case. The second call later on was still handling all other situations.

However, on pondering this more there's no reason the first call can't handle almost everything and then the second call just needs to handle scaling after yuv/bayer/jpeg/png conversion. Given this, I can actually reduce the complexity of the first call's to just checking `runnable_on_gpu && (!(hint & IMAGE_HINT_TRANSPOSE))` I think. The GPU call itself will return -1 if a jpeg/png/bayer/yuv image is passed.

> You should rename this is_color_conversion to something else like is_bayer_yuv_conversion.
Sure

> Some of the bool flags are only used one, please double check. For example do you really need is_color_conversion_scaling, need_deep_copy, is_jpeg, is_png ?
That's done to avoid a giant if statement filled with boolean logic.

Given me a bit, will update it...

iabdalkader: After debayer, JPEG/PNG decompression or YUV conversion it's a bit too late to benefit from GPU scaling speed up, isn't it? I'd just let the software fallback handle all other unsupported cases.

- Note that GPU can transpose while scaling, not tested and unsure if it can be done in place.
- Please remove the flags I mentioned.

kwagyeman: Done

src/omv/imlib/draw.c
```diff
@@ -3061,19 +3020,50 @@ void imlib_draw_image(image_t *dst_img,
...
+            goto exit_cleanup;
+        }
+    }
+    #endif
 
     // Make a deep copy of the source image.
-    if (need_deep_copy || is_color_conversion_scaling || is_jpeg || is_png) {
+    if (((dst_img->data == src_img->data) && (is_scaling || is_line_length_growing || is_bayer_yuv_conversion))
+        || (is_bayer_yuv_conversion && is_scaling)
+        || src_img->is_compressed) {
```
iabdalkader, line 3066:
```suggestion
    if (((dst_img->data == src_img->data) &&
        (is_scaling || is_line_length_growing || is_bayer_yuv_conversion)) ||
        (is_bayer_yuv_conversion && is_scaling) ||
        src_img->is_compressed) {
```

iabdalkader (changes requested):

src/omv/imlib/draw.c
```diff
@@ -3061,19 +3020,50 @@ void imlib_draw_image(image_t *dst_img,
...
-    bool is_color_conversion = is_bayer_color_conversion || is_yuv_color_conversion;
+    bool is_bayer_conversion = src_img->is_bayer && !dst_img->is_bayer;
+    bool is_yuv_conversion = src_img->is_yuv && !dst_img->is_yuv;
+    bool is_bayer_yuv_conversion = is_bayer_conversion || is_yuv_conversion;
 
-    // Force a deep copy if we cannot use the image in-place.
-    bool need_deep_copy = (dst_img->data == src_img->data)
-                          && (is_scaling || (src_img_row_bytes < dst_img_row_bytes) || is_color_conversion);
+    // Is the line length growing which will prevent us from working in-place?
+    bool is_line_length_growing = src_img_row_bytes < dst_img_row_bytes;
```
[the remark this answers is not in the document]
iabdalkader: Or `is_downscaling` -> `is_scaling_down`.

src/omv/imlib/draw.c
```diff
@@ -3061,19 +3020,50 @@ void imlib_draw_image(image_t *dst_img,
...
 
-    // Force a deep copy if we are scaling.
-    bool is_color_conversion_scaling = is_color_conversion && is_scaling;
+    #if (OMV_GPU_ENABLE == 1)
+    if ((!callback) &&
+        (!is_bayer_yuv_conversion) &&
+        (!src_img->is_compressed) &&
+        (rgb_channel < 0) &&
+        ((dst_img->data != src_img->data) || (!is_line_length_growing)) &&
+        (!(hint & (IMAGE_HINT_AREA | IMAGE_HINT_BICUBIC | IMAGE_HINT_TRANSPOSE)))) {
```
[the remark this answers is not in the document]
kwagyeman: DMA2D completely breaks when ```dst_img->data == src_img->data && is_line_length_growing```.

For example, doing `.to_rgb565()` on an GRAYSCALE image.

`dst_img->data != src_img->data || !is_line_length_growing`  allows it to work in-place as long as the mutable format isn't growing in length or out of place. `!is_line_length_growing` by itself is unnecessarily restrictive as it prevents the GPU from being used during an out-of-place grayscale->rgb565 conversion or etc.

I could push the check into the GPU driver. But, it's somewhat universal. The GPU would need to cache the entire image or work on the pixel array backwards to avoid corrupting the image.

src/omv/imlib/draw.c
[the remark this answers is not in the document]
kwagyeman: The check isn't about detecting scaling up or down. It's just to detect when a change from binary->grayscale, binary->rgb565, and grayscale->rgb565. Seeing the destination size growing was just a straightforward way to determine this.

Maybe `line_bbp_increased`?

src/omv/imlib/draw.c
kwagyeman: Done

kwagyeman:
```
if (((dst_img->data == src_img->data) &&
         (is_scaling || is_upscaling || is_bayer_yuv_conversion)) ||
        (is_bayer_yuv_conversion && is_scaling) ||
        src_img->is_compressed) {
```

Note: Clang requires a space on the `(is_scaling || is_upscaling || is_bayer_yuv_conversion))` line since it's part of the compound statement of the first line.

src/omv/imlib/draw.c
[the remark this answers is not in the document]
iabdalkader: True, then just rename it please.

src/omv/imlib/draw.c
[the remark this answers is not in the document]
iabdalkader: `is_line_expanding` ?

src/omv/imlib/draw.c
iabdalkader: Sure, the point is the logical ops go on the end.

src/omv/imlib/draw.c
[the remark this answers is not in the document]
kwagyeman: Done

src/omv/imlib/draw.c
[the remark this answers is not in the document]
kwagyeman: Done

iabdalkader (approved):

PR #2562, opened 2024-12-27
imlib/draw: Simplify GPU source rect calculation.

kwagyeman:
Tested with scaling for for various network input sizes and it produces a clean result.

### DIFF
--- a/src/omv/imlib/draw.c
+++ b/src/omv/imlib/draw.c
@@ -3043,10 +3043,10 @@ void imlib_draw_image(image_t *dst_img,
         };
 
         rectangle_t src_rect = {
-            .x = src_x_accum_reset >> 16,
-            .y = src_y_accum_reset >> 16,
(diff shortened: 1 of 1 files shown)


src/omv/imlib/draw.c
```diff
@@ -3043,10 +3043,10 @@ void imlib_draw_image(image_t *dst_img,
         };
 
         rectangle_t src_rect = {
-            .x = src_x_accum_reset >> 16,
-            .y = src_y_accum_reset >> 16,
-            .w = ((src_x_accum_reset + (src_x_frac * dst_rect.w)) >> 16) - src_rect.x,
-            .h = ((src_y_accum_reset + (src_y_frac * dst_rect.h)) >> 16) - src_rect.y,
+            .x = fast_roundf(src_x_start / x_scale),
+            .y = fast_roundf(src_y_start / y_scale),
```
kwagyeman, line 3047: Note that src_x_start and src_y_start have already been scaled so this is unscaling.

PR #2564, opened 2024-12-28
sensors/pag7936: Add support for triggering.

kwagyeman:
Enables the global shutter triggering with the PAG7936. Tested on the N6 and AE3.

Implemented per SingTown's request since January.

**Dependent on: https://github.com/openmv/openmv/pull/2985 being merged first.**

### DIFF
--- a/drivers/sensors/pag7936.c
+++ b/drivers/sensors/pag7936.c
@@ -88,9 +88,16 @@
 #define SENSOR_OPMODE_SUSPEND       (0x86)
 #define SENSOR_TRIGGER_FRAMENUM     (0x002E)
 #define SENSOR_TRIGGER_EN           (0x002F)
+#define SENSOR_TRIGGER_EN_FLAG      (0x01)
+#define SENSOR_TRIGGER_MODE_SOFT    (0x00)
+#define SENSOR_TRIGGER_MODE_GPIO0   (0x10)
+#define SENSOR_TRIGGER_MODE_GPIO1   (0x20)
+#define SENSOR_TRIGGER_MODE_GPIO2   (0x30)
 #define SENSOR_TG_EN                (0x0030)
 #define SENSOR_TG_EN_FLAG           (0x01)
 #define SENSOR_TRIGGER_MODE         (0x0031)
+#define SENSOR_TRIGGER_MODE_0       (0x02)
+#define SENSOR_TRIGGER_MODE_1       (0x03)
 #define SENSOR_SOFTWARE_TRIGGER     (0x00EA)
 #define ISP_EN_H                    (0x0800)
 #define ISP_EN_H_EN                 (0x01)
@@ -843,6 +850,39 @@ static int ioctl(omv_csi_t *csi, int request, va_list ap) {
             *va_arg(ap, uint32_t *) = rgb_stats[2];
             *va_arg(ap, uint32_t *) = rgb_stats[0];
             *va_arg(ap, uint32_t *) = rgb_stats[3];
+            break;
+        }
+        case OMV_CSI_IOCTL_SET_TRIGGERED_MODE: {
+            int enable = va_arg(ap, int);
+            ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, SENSOR_OPMODE, 2, SENSOR_OPMODE_SUSPEND, 1);
+
+            if (enable) {
+                ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, SENSOR_TG_EN, 2, 0, 1);
+                ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, SENSOR_TRIGGER_EN, 2, SENSOR_TRIGGER_EN_FLAG, 1);
+                ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, SENSOR_TRIGGER_MODE, 2, SENSOR_TRIGGER_MODE_1, 1);
+                ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, SENSOR_TRIGGER_EN, 2,
+                                         SENSOR_TRIGGER_EN_FLAG | SENSOR_TRIGGER_MODE_GPIO0, 1);
+            } else {
+                ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, SENSOR_TG_EN, 2, SENSOR_TG_EN_FLAG, 1);
+                ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, SENSOR_TRIGGER_EN, 2, 0, 1);
+                ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, SENSOR_TRIGGER_MODE, 2, 0, 1);
+            }
+
+            ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, SENSOR_OPMODE, 2, SENSOR_OPMODE_RUN, 1);
+
+            // Skip past the first corrupt frames...
+            if (!csi->disable_delays) {
+                mp_hal_delay_ms(100);
+            }
+            break;
+        }
+        case OMV_CSI_IOCTL_GET_TRIGGERED_MODE: {
+            int *enable = va_arg(ap, int *);
+            uint8_t reg;
+            ret |= omv_i2c_read_reg(csi->i2c, csi->slv_addr, SENSOR_TG_EN, 2, &reg, 1);
+            if (ret >= 0) {
+                *enable = (reg & SENSOR_TG_EN_FLAG) ? 0 : 1;
+            }
             break;
         }
         default: {


willothewisp0331:
Hi, great to see the progress on PAG7936! We are currently working on a project that heavily relies on the trigger function for the PAG7936 sensor. We are wondering if the trigger feature will be available soon? 
If it's still in progress, would you recommend us to use a specific branch for testing in the meantime? 
Thanks for your hard work!

kwagyeman:
I'll get it working tomorrow. However, it's not going to be ready to merge for a while. There's some refactoring that needs to be done to make the trigger signal GPIO pin settable by the camera driver when bound. So, I can only just make it function right now. But, not merge.

willothewisp0331:
We'd be happy to use the 'functional but not yet merged' version for now.
Could you please let us know when the code is available (perhaps in a specific branch or a temporary commit)
We are also happy to help test it on our hardware and provide feedback if that helps with your refactoring later.
Thank you so much for the quick turnaround.

kwagyeman:
I'll just get this commit working and drop a firmware for the N6 and AE3 on it that you can load with OpenMV IDE.

kwagyeman:
Still working on this.

kwagyeman:
Got it working now.

kwagyeman:
[firmware.zip](https://github.com/user-attachments/files/25821034/firmware.zip)
Please find a tip of dev N6 firmware attached.

kwagyeman:
[firmware_M55_HP.zip](https://github.com/user-attachments/files/25821041/firmware_M55_HP.zip)
Please see the attached AE3 firmware.

willothewisp0331:
Hi @kwagyeman 
Thank you for the hard work on this.

We’ve tested `csi.IOCTL_SET_TRIGGERED_MODE` on the N6 sensor for hardware synchronization. 
While it works, we found a significant performance drop using the following test code:

```
import csi
import time

csi0 = csi.CSI()
csi0.reset()
csi0.pixformat(csi.GRAYSCALE)
csi0.framesize(csi.QVGA)
csi0.auto_exposure(False, exposure_us=2000)

# We found that setting it to 470 FPS works best; 
# setting it to 480 FPS seems to cause the FPS to drop to ~200.
csi0.framerate(470)
csi0.snapshot(time=2000)
clock = time.clock()

csi0.ioctl(csi.IOCTL_SET_TRIGGERED_MODE, True) # False for Continuous Mode

i = 0
while True:
    clock.tick()
    img = csi0.snapshot()
    i = i + 1
    if (i % 200 == 0):
        print(clock.fps())
```
       

Continuous Mode: ~470 FPS (QVGA)

Triggered Mode : ~180 FPS

Our target is 240+ FPS. It seems the current snapshot() implementation in triggered mode introduces a significant bottleneck. We expected some overhead, but not a drop of this magnitude.

We will open a dedicated topic on the OpenMV Forum to discuss our specific goals and see if there are alternative approaches (such as low-level optimizations or DMA tweaks) to reach our performance target.

Thanks!

iabdalkader:

https://forums.openmv.io/t/achieving-240-fps-with-frame-synchronized-multi-source-tracking-n6-pag7936/11542/1

kwagyeman:
@willothewisp0331 - Turning triggered mode on causes a performance drop. This is kinda expected. When you turn it on new frames aren't exposed until you call snapshot.

What are you trying to do? The current behavior that happens is consistent with what the feature is designed to do... There's another triggering mode that increases the frame readout delay (mode 0) but allows the camera to pipeline things. This may be what you want. Can you describe?

kwagyeman:
Hi OpenMV Team,

I am working on a high-speed spot tracking system using the OpenMV N6 with the PAG7936 sensor. I need to synchronize external light sources (via GPIO) with the camera’s exposure window to ensure each frame only captures specific light spots without cross-talk.

Continuous Mode: ~470 FPS (QVGA)
Triggered Mode (csi.IOCTL_SET_TRIGGERED_MODE): Drops to ~180 FPS (Raw capture)
[Add support for triggering.](https://github.com/openmv/openmv/pull/2564)
Actual System (with find_blobs & UART): ~110 FPS
Target: 240+ FPS
Proposed Schemes:

Scheme A: Triggered Mode Optimization The snapshot() call in triggered mode seems to have a large software overhead. Is there any low-level ioctl or driver tweak to reduce the trigger-to-capture latency to reach 240+ FPS?

Scheme B: External Sync with Frame Indexing If we run in Continuous Mode (400+ FPS) and use an external MCU to pulse lights via FSYNC (P10):

Is there a way for the Python script to read a Hardware Frame Count or DMA tag to ensure the snapshot() buffer aligns with the correct external light phase?
Scheme C: Hardware Windowing (ROI) Does the N6 driver support hardware-level cropping/windowing (e.g., to QQVGA 160x120) to increase the readout speed?

Scheme D: Asynchronous Processing & Task Pipelining

Background DMA: Does the CSI driver support Double Buffering / Background DMA? We want to expose the next frame while the CPU is processing the current one.

Task Pipelining (UART & CPU): Currently, find_blobs and UART communication are executed sequentially. Is there a way to offload the UART transmission to DMA or a separate thread (without blocking the CPU) so that the next find_blobs can start immediately?

N6 NPU Offloading: Can the N6’s internal NPU be leveraged to offload basic image tasks like blob detection? This would significantly free up the CPU for protocol handling and system logic.

We prioritize speed and deterministic frame-to-light alignment. Which of these approaches would you recommend for achieving 240+ FPS?

Thanks!

```
import csi, time
from machine import Pin

csi0 = csi.CSI()
csi0.reset()
csi0.pixformat(csi.GRAYSCALE)
csi0.framesize(csi.QVGA)
csi0.auto_exposure(False, exposure_us=2000)
csi0.framerate(470)
csi0.snapshot(time=2000)

csi0.ioctl(csi.IOCTL_SET_TRIGGERED_MODE, True)
pins = [Pin("P0", Pin.OUT_PP), Pin("P1", Pin.OUT_PP)]

frame_count = 0
clock = time.clock()

while True:
    clock.tick()
    target_idx = frame_count % 2

    pins[target_idx].high() # turn on LED

    img = csi0.snapshot()

    pins[target_idx].low() # turn off LED

    # Processing & Communication (Takes ~5ms)
    # blobs = img.find_blobs(...)
    # uart.write(...)

    frame_count += 1
    if frame_count % 100 == 0:
        print(clock.fps())
```

kwagyeman:
Hi, I'll try to answer with a comprehensive response in a day. What you want to do is possible via the hardware. But, you're going to have to do a lot of firmware edits as hitting the frequency you want means that you need to optimize a lot of code paths.

willothewisp0331:
Hi, we fully understand that the standard Python API may have reached its limits for this specific high-speed requirement. We are prepared to dive into firmware-level adjustments or hardware architecture tweaks to bridge the performance gap.

We are looking forward to your comprehensive response. Any guidance, low-level hints, or specific code paths you recommend optimizing would be immensely helpful for our development.

Thank you for your support!

kwagyeman:
Sorry for the delayed response, I was busy at a conference.

Okay, so when you put the camera into triggered mode, it takes a picture when it receives a trigger. Each snapshot call triggers the camera, then the image is received and processed before the CPU sends the next trigger. This approach aligns the image capture such that it happens when you call snapshot.

You can also put the camera into a mode where it outputs a signal each time it takes a picture itself. However, you will need to disable the OpenMV Cam from driving the trigger pin as it will also try to drive the trigger pin too. However, in this mode you can leave the camera free running and it will give you a trigger waveform to use.

Let me know if you want this and I can give you a python snippet. We may ad such functionality to the API if there's demand.

Now, on image reception, the OpenMV Cam already has a DMA system that receives frames in the background. So, that's done. It's triple-buffered by default, so there are two back buffers always receiving frames. So, as long as your processing loop doesn't take longer than the image capture, then you are good. You won't drop any frames.

Regarding processing, you should use the NPU. You will find the NPU outperforming the CPU. I would switch from using find_blobs() to an AI model like Edge Impulses' FOMO model for object tracking, or training your own color tracking CNN. The NPU is just so much faster than the CPU that it will outperform for any application where you can use it. We can do 240 FPS almost with FOMO face detection.

Finally, regarding the UART. Yes, that will block. There's no DMA buffer with that by default with MicroPython. However, if you set the baud rate very high then the blocking time should be minimal. You will need to create your own UART driver from scratch in C with a fifo and etc. if you want a deeper buffer than just the minimal fifo built into the STM32 peripheral. I'd recommend just using a very fast baud rate such that the maximum time a message blocks for is less than 2ms.

willothewisp0331:
Hi, I would like to try the mode where the camera outputs a signal each time it takes a picture. Regarding the trigger pin, I assume you are referring to P10 (FSYNC) on the N6 module? Currently, I am not driving any external trigger pins; my synchronization is purely handled by the software snapshot() call.

Please provide the Python snippet for this "free-running with trigger output" mode. It will save me a significant amount of development time, and I am very grateful for any resources you can share.

Regarding image processing, I am intrigued by the NPU's performance. I will conduct a benchmark comparing NPU-based AI models (like FOMO) against CPU-based find_blobs() and post the FPS results back to share with anyone needed.

Looking forward to the snippet and further guidance. Thank you!

kwagyeman:
Hi, just use the master branch for the N6 firmware. Tools->Install Latest Development release.

Then, try this snippet out:

```
csi.__write_reg(0x0077, 0xC2)
csi.__write_reg(0x0082, 0x01)
```

This should turn on the camera sensor to output a square wave pulse on P10 that is perfectly matched with the sensor exposure. The GPIO will be active low.

Do: `csi.__write_reg(0x1229, 0x11)` to make it active high (and `csi.__write_reg(0x1229, 0x01)` to active low.

...

I am traveling now and cannot test this as I do not have the tools. But, this is what the datasheet suggests.

willothewisp0331:
Hi. We’ve verified the FSYNC (P10) signal with an oscilloscope, and it works perfectly! Our PoC script is also functioning as expected. We are now trying to integrate this hardware-trigger architecture into our application.

We are also in the process of testing the NPU + FOMO model and will share the results once ready.

Thank you so much for the help. Safe travels and have a great trip!

kwagyeman:
Great! I'll add an IOCTL for this ability that's separate from the normal trigger.

TofATR:
Hi everyone, and thanks for your support.
Yesterday I tried the firmware for the N6 that supports triggered mode.

I used this code to take measurements:
`
import csi
import time

cam = csi.CSI()
cam.reset()
cam.pixformat(csi.RGB565)
cam.framesize(csi.QVGA)
cam.auto_exposure(False, exposure_us=300)
cam.auto_gain(False, gain_db=0)
cam.brightness(0)
cam.snapshot(time=2000)
cam.ioctl(csi.IOCTL_SET_TRIGGERED_MODE, True)
clock = time.clock()
maxi = 0
moy = 0

while (True):
    time.sleep_ms(10)
    trigg = time.ticks_us()
    img = cam.snapshot()
    trigg2 = time.ticks_us()-trigg
    if maxi <= trigg2:
        maxi = trigg2
    moy = moy*0.95 + trigg2*0.05
    print(trigg2, " maxi=", maxi, " moy=", moy)
`

I'm randomly observing image capture times of 3.5ms or 11ms. I have the impression that there's still some frame synchronization and that the integration isn't starting at the moment of the cam.snapshot() call.

In my application, I'm taking images of a moving object (approx. 800mm/s) triggered by an external trigger.

Is there a new firmware available for the N6/PAG7936, or another method to force the integration to start in the most synchronized time with my external trigger?

willothewisp0331:
Quick update and a follow-up question as we integrate this into our application.

We re-tested the NPU + FOMO model and realized why it didn't provide a speedup: the default example script actually runs the AI inference on top of find_blobs (which runs on the CPU). Since it adds extra AI overhead without reducing the CPU load, we decided not to pursue that direction.

Currently, our optimized image processing (using find_blobs) + UART overhead takes about 7ms. We need to alternate between 2 sequential IR LEDs (LED 1 -> LED 2) that must align perfectly with each frame. However, we are hitting a bottleneck trying to push for higher FPS:

1. Triggered Mode: Alignment is 100% stable with zero frame drops. However, due to the lack of pipelining, the total frame time is 9ms (2ms exposure + 7ms process), dropping the effective update rate per LED to only 55 FPS.
2. Continuous Mode (with FSYNC): Pipelining works, but because the 7ms processing time is longer than the sensor's frame period at high FPS, the Triple Buffer automatically drops older frames to maintain real-time tracking. This causes phase desynchronization, making it impossible to map which image belongs to which LED phase.

Since our application strictly requires processing every single frame sequentially without any frame drops, we wanted to ask:

1. Is there an API or method to disable "Auto-Frame-Drop" in Triple Buffer Mode, forcing csi.snapshot() to act as a strict FIFO queue?
2. Alternatively, does your recent PR (https://github.com/openmv/openmv/pull/2564) for PAG7936 triggering support allow a low-overhead/pipelined triggering mode where the next sensor exposure can overlap with the current Python processing loop?

Thanks again for your guidance!

kwagyeman:
Hi, if you want the camera output pipelined, you just need to not turn triggering on. Then the camera outputs frames as fast as possible. Then do this to get when the camera exposes:

```
csi.__write_reg(0x0077, 0xC2)
csi.__write_reg(0x0082, 0x01)
```

As for making an image FIFO. Yes, you just do `csi0.framebuffers(100)`. See https://docs.openmv.io/library/omv.csi.html#csi.CSI.framebuffers

You can increase the frame buffer count to as many that will fit. Also, make sure to turn off flushing when the frames are captured: `csi0 = csi.CSI(fflush=False)` https://docs.openmv.io/library/omv.csi.html#csi.CSI. Frames will then accumulate in the fifo and not be flushed when the fifo overflows. You will drop frames after that, but, this will give you atleast N frames of valid data to observe.

It is not possible to continually capture frames and not drop them if your processing time is longer than the frame capture speed. The above is the best you can do with a buffer that captures as much detail as possible until RAM is exhausted.

iabdalkader:

https://forums.openmv.io/t/imu-data-ready-on-ae3-lsm6dsm/11672/2

chobitsfan:
Hi @kwagyeman Thank you very much for adding trigger support

On AE3, I want to trigger PAG7936 when new imu data available (align camera with imu)
```
cam = csi.CSI()
cam.ioctl(csi.IOCTL_SET_TRIGGERED_MODE, True)
while True:
    # read STATUS_REG
    data_rdy = imu.__read_reg(0x1e)
    # check if new accel & gyro data available
    if (data_rdy & 3) == 3:
        imu.acceleration_mg()
        imu.angular_rate_mdps()
        cam.snapshot()
```
Because we are in trigger mode, ```snapshot()``` blocks util exposure and readout finished. Is it possible to solve it like following code?
```
while True:
    # read STATUS_REG
    data_rdy = imu.__read_reg(0x1e)
    # check if new accel & gyro data available
    if (data_rdy & 3) == 3:
        imu.acceleration_mg()
        imu.angular_rate_mdps()
        cam.trigger()
    if cam.readable():
        # snapshot read image without trigger exposure
        cam.snapshot()
```

Thank you very much

kwagyeman:
It's not added yet, waiting on PR review.

Question: What are you trying to do? Could you post on the forums about this versus the PR. These are not meant for help or support.

kwagyeman:
@iabdalkader - Rebased and ready to merge.

iabdalkader (approved):

sid115:
I cannot get triggering to work on N6 with PAG7936. I used the current dev release firmware (https://github.com/openmv/openmv/commit/e9eeef4e19926f6b0219993977226c5bf501906a which includes this PR as far as I can see):

```python
csi0 = csi.CSI()
csi0.ioctl(csi.IOCTL_SET_TRIGGERED_MODE, True) # RuntimeError: Sensor control failed.
```

OpenMV IDE v4.8.10 on Windows 11.

kwagyeman:
```
import csi
import time

csi0 = csi.CSI()
csi0.reset()
csi0.pixformat(csi.RGB565)
csi0.framesize(csi.VGA)
csi0.snapshot(time=2000)
csi0.ioctl(csi.IOCTL_SET_TRIGGERED_MODE, True)

clock = time.clock()

while True:
    clock.tick()
    img = csi0.snapshot()
    print(clock.fps())
```

Works as expected with the latest firmware. Init the sensor first.

willothewisp0331:
Hi @kwagyeman ,

Quick update from our side — it's been a few months since we last talked. We've been running continuous mode + FSYNC (P10) output at high FPS for multi-source IR tracking, and found some issues we'd like your input on.

**What we found:**

We cycle 2~4 IR sources using the FSYNC pulse (one source per frame, round-robin). To know which source was active for a given frame, we count FSYNC pulses in a hardware interrupt and check the count after `snapshot()` returns.

The problem: the offset between our pulse counter and the frame returned by `snapshot()` is not constant. Sometimes it's 1, sometimes it's 2. This causes the source assignment to randomly shift by one slot — source A's data gets written to source B, and vice versa.

Single source works fine. The issue only appears with 2+ sources because the shifting corrupts the round-robin mapping.

We also tried:
- Reducing exposure from 2000μs to 1000μs → helped reduce cross-frame light bleed, but didn't fix the shifting
- Lowering framerate from 200 to 150 FPS (via `set_framerate`) → same shifting problem still occurs

**Our situation:**

- Processing per frame: ~7ms (find_blobs + UART)
- Frame period at 200 FPS: ~5ms (at 150 FPS: ~6.7ms — still shorter than 7ms processing)
- We're fine with dropping frames
- But we need to know which FSYNC pulse each returned frame belongs to
- This runs continuously for hours (commercial product)

**Questions:**

1. Is there any way to read a frame index or counter after `snapshot()` that tells us which FSYNC pulse that specific frame was captured on?

2. You mentioned "mode 0 that allows pipelining" before — would that give us a fixed offset between FSYNC and `snapshot()`?

3. With `framebuffers(2)` + `fflush=False`, is the buffer depth between FSYNC pulse and `snapshot()` return supposed to be constant?

We've tried a lot of software workarounds over the past months but the shifting seems to be at the driver/buffer level. Any hints would be really helpful.

Thanks!

kwagyeman:
@willothewisp0331 - I think we should set up a meeting and get on a call. What you want to do seems like it will require deep firmware-level work. Like, how you are capturing pulses, etc., matters here. The issue you are likely having is interrupt latency. Using a pin interrupt won't work; you likely need to use a hardware unit like a counter on the N6 for this. 

Anyway, please email us, and then we can get this all set up.

PR #2565, opened 2024-12-28
misc: romfs support.

iabdalkader:


scripts/libraries/ml/ml/model.py
```diff
@@ -29,11 +29,18 @@
 import uml
 import image
 from ml.preprocessing import Normalization
+import os
 
 
 class Model(uml.Model):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, kwargs.get("load_to_fb", False))
+        try:
```
kwagyeman, line 38: Nice

iabdalkader:
I had support for embedding the romfs image, for boards with internal flash only, using a trick like this:
```
$(OBJCOPY) -I binary -O elf32-littlearm -B arm $(ROMFS_IMAGE) \
                  $(BUILD)/romfs.o --rename-section .data=.romfs

#if defined(OMV_FLASH_ROMFS_ORIGIN)
.romfs (READONLY) : ALIGN(4)
{
  KEEP(*(.romfs*));
} >FLASH_ROMFS
#endif
```

However I'm dropping support for this as it requires the sector to be right before/after the firmware, so you couldn't place romfs in dead-space like sector 1 for Arduino boards. Also it may get in the way when/if multiple partitions are supported.

PR #2567, opened 2024-12-30
imlib/imlib: xalloc image to be cache aligned.

kwagyeman:
Ensures all xalloc calls to allocate image data are cache aligned and rounded up to the nearest size of a cache line. Both of these conditions are required to ensure the allocated image data doesn't interact with any other cache lines outside of it's allocation.


kwagyeman:
Note that: https://github.com/openmv/openmv/blob/master/src/omv/modules/py_ml.c#L383

Only aligns the allocation. It doesn't ensure that it's a multiple of the cache line size. So, another allocation could be created right behind it, occupying the same cache line. As long as the model data is read-only, this should be okay. But, if it needs to be cache invalidated, then it could corrupt anything allocated right after it.

PR #2568, opened 2024-12-30
scripts/libraries: Add yolo v5 post processing support.

kwagyeman:


scripts/libraries/ml/ml/postprocessing.py
```diff
@@ -33,6 +33,9 @@
 from ulab import numpy as np
 
 
+_NO_DETECTION = const(())
```
kwagyeman, line 36: Had to be here otherwise I got a compile error: SyntaxError: can't assign to expression

PR #2595, opened 2025-02-05
sensors/GenX320: Adding ioctl for setting individual biases.

nlyubova-psee:
* Adding ioctl for setting individual biases proving bias name and value
* Updating the sensor biases values used on the camera start to make the sensor more sensitive


iabdalkader:
Why not just support passing the biases in a list? For example:

```Python
sensor.ioctl(sensor.IOCTL_GENX320_SET_BIASES, [51, 45, 55, 50, 127, 0])
```
And then something like:
```C
if (mp_obj_is_int(args[1])) {
    error = omv_csi_ioctl(request, mp_obj_get_int(args[1]));
} else {
    uint8_t bias[6];
    mp_obj_t *array;
    mp_obj_get_array_fixed_n(args[1], OMV_ARRAY_SIZE(bias), &array);
    for (int i = 0; i < OMV_ARRAY_SIZE(bias); i++) {
        bias[i] = mp_obj_get_(array[i]);
    }
    // Special type means the arg is a pointer to array.
    error = omv_csi_ioctl(request, OMV_CSI_GENX320_BIASES_RAW, bias);
}
```

Technically, this could replace all of the other ioctls, but if these number change with the firmware/revisions of the sensor, the example script will have to detect this and use the right numbers somehow.

kwagyeman:
Nice

nlyubova-psee:
in this case, you need to set all biases every time and you need to check what are they are and you need to remember the order. Moreover, often we need to change one particular bias (like sensitivity or low-pass filter) without touching others.
Also we did an internal survey and people prefer to set values one by one.

iabdalkader:
Okay no problem.

lbristiel-psee:
I see your point, @iabdalkader, setting all the biases at once might seem like a time saver. However, as Natalia mentioned, in practice, users typically adjust only one or a few biases at a time and prefer not to modify or even be concerned with the others.
Given this, I’d stick with Natalia’s proposal.

kwagyeman:
PR looks good to me.

iabdalkader:
> users typically adjust only one or a few biases at a time and prefer not to modify or even be concerned with the others.

Wll you could pass a dictionary instead. For example:
```Python
sensor.ioctl(sensor.IOCTL_GENX320_SET_BIASES, {"hpf" : 50, "diff_on": 100})
``` 
Though processing this might be tricky, as I'd prefer not to pass MP objects to `ioctls`.

iabdalkader (approved):

PR #2597, opened 2025-02-07
ports/stm32: Move CLUT to DMA memory section.

kwagyeman:
Removes the need for fb_alloc in the GPU driver.

### DIFF
--- a/src/omv/ports/stm32/omv_gpu.c
+++ b/src/omv/ports/stm32/omv_gpu.c
@@ -132,6 +132,8 @@ int omv_gpu_draw_image(image_t *src_img,
     return 0;
 }
 #else
+uint32_t OMV_ATTR_SECTION(OMV_ATTR_ALIGNED(CLUT_BUFFER[256], 32), ".dma_buffer");
+
 int omv_gpu_draw_image(image_t *src_img,
                        rectangle_t *src_rect,
                        image_t *dst_img,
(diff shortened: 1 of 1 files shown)


src/omv/imlib/draw.c
```diff
@@ -3140,6 +3140,15 @@ void imlib_draw_image(image_t *dst_img,
...
+        #if (OMV_GPU_ENABLE == 1 && OMV_GPU_RESERVED_FB_ALLOC_BYTES > 0)
+        if (size > OMV_GPU_RESERVED_FB_ALLOC_BYTES) {
+            size -= OMV_GPU_RESERVED_FB_ALLOC_BYTES;
+        } else {
+            fb_alloc_fail();
+        }
+        #endif
+
         // Allocate a buffer to hold chunks of the transposed image while not using all of the on-chip RAM.
         size = IM_MIN(size, image_size(&t_src_img));
```
kwagyeman, line 3153: This limits the RAM usage during transpose to what is actually needed. The more RAM transpose has to work with up to the image size the faster it goes. However, if there's less RAM available than the image size then this uses up all fb_alloc space available still so it's not enough to ensure the GPU has RAM for CLUT loading or etc.

kwagyeman: It's only 1KB if it was reserved in the DMA memory area that is algined. But, DMA RAM is somewhat tight on some of our models. Might be hard to fit 1KB on the older boards. I think this fix is better than reserving RAM.

src/omv/boards/ARDUINO_GIGA/omv_boardconfig.h
```diff
@@ -41,6 +41,7 @@
 
 // GPU Configuration
 #define OMV_GPU_ENABLE                      (1)
+#define OMV_GPU_RESERVED_FB_ALLOC_BYTES     (2048)
```
iabdalkader, line 44:
```suggestion
#define OMV_GPU_DMA2D_CLUT_SIZE (2048)
```

src/omv/imlib/draw.c
kwagyeman: Huh, it all fits in DMA memory. Okay, guess this is settled. Just need to retest.

iabdalkader: Yes, that's what it's for. Note OMV3 uses DTCM for this, so you may want to fix it.

kwagyeman: I get an instant reset on the Giga with this... debugging.

kwagyeman: OMV3 works with this since there's an AHB_S port on the CPU which allows DMA access to the DTCM.

Test the OMV3 and H7P and they seem to be fine.

iabdalkader:
I tested the DMA buffer on Giga and it works fine. There's an `fb_free()` at the end of the gpu_draw function that should be moved. However, even with this fix I'm still running out of fb_memory, probably in another place. If I use `sensor.set_framebuffers(1)` all works fine.

kwagyeman:
Updated with the fixed commit code.

Regarding the running out of memory. I don't see that with grayscale VGA cameras.

* Frame Buffer = 640x480x2x3 -> 1,843,200
* LCD Buffer = 800x480x2x3 -> 2,304,000
= 2,304,000 + 1,843,200 = 4,147,200
-> Remaining RAM = 47,104 -> The transpose operation uses this remaining RAM to draw on the display.

Doing .to_rgb565() definitely runs out of RAM. But, if the camera is RGB565 to start with then it should be fine.

src/omv/boards/ARDUINO_GIGA/omv_boardconfig.h
kwagyeman: Commit dropped.

src/omv/imlib/draw.c
kwagyeman: Resolved.

kwagyeman:
Tested this on the:
* M4
* M7
* H7
* H7P
* Giga
* Portenta
* Nicla
And it works.

iabdalkader:
> Regarding the running out of memory. I don't see that with grayscale VGA cameras.

@kwagyeman  I use GC21 (RGB), and it still fails with this fix here:

image

I think there might be a bug in transpose because if I build the exact same firmware but with `DEBUG=1` it doesn't fail. I think this PR is unrelated though.

iabdalkader (approved):

Issue #2594, opened 2025-02-05
Register_callback


Unscythe:
No successful return message from register_callback function

Slave(main.py):
```
import rpc
import pyb  # Import pyb module to control the LED

# Initialize the RPC interface (USB Virtual COM Port)
interface = rpc.rpc_usb_vcp_slave()  # For UART use rpc_uart_slave()

# Initialize the LED (typically LED(1) is the blue LED on OpenMV)
led = pyb.LED(1)

# Define a simple function to respond with a message and blink the LED
def hello_world(data):
    led.on()               # Turn on the LED
    pyb.delay(100)         # Keep it on for 100 ms
    led.off()              # Turn off the LED
    return "Hello from OpenMV!"

# Register the hello_world function to be called over RPC
interface.register_callback(hello_world)

# Main loop to keep the RPC server running
interface.loop()  # Continuously listen for RPC commands
```

Master:
```

import rpc
import serial

# Initialize the connection to OpenMV via USB (VCP)
# Change the port to match your OpenMV camera's USB serial port (e.g., /dev/ttyACM0)
usb_port = '/dev/ttyACM0'
ser = serial.Serial(usb_port, baudrate=115200, timeout=1)

# Create an RPC client over USB VCP
client = rpc.rpc_usb_vcp_master('/dev/ttyACM0')

# Call the 'hello_world' function on OpenMV and get the response
result = client.call("main.py")

# Print the response from OpenMV
if result is not None:
    print("Response from OpenMV:", result)
else:
    print("No response from OpenMV.")

```

Output from terminal:
```
Response from OpenMV: <memory at 0x7fad669f40>
```
The response is from result = client.call("main.py")
The hello_world() callback function is not getting a response as the LED is also not blinking and there isn't a "Hello from OpenMV!" output

kwagyeman:
Hi, it's because the library is failing to connect to the script running on the camera. This could be because the camera isn't running your script or because the host computer isn't responding to serial packets fast enough to keep up with the MCU's timeouts. The protocol is meant for SPI/I2C/UART transmission between MCUs and works well for real-time systems.

Anyway, a better method is to use the [pyboard](https://github.com/micropython/micropython/blob/master/tools/pyboard.py) script to control the OpenMV Cam. It can send a script to the camera, stop/stop it, and receive the results.

Alternatively, you can also use our debug script tools: https://github.com/openmv/openmv/blob/master/tools/pyopenmv.py

...

[...]

Issue #2433, opened 2024-10-01
RPC Library not working over USB using VCP with another computer


Unscythe:
When trying to connect a raspberry Pi to a OpenMV H7 Hand cam plus over USB using VCP I am unable to receive any data from OpenMV.
Master:
```
import rpc
import serial

usb_port = "/dev/ttyACM0"

client = rpc.rpc_usb_vcp_master(port=usb_port)

while True:
    result = client.call("hello_world")

    if result is not None:
        print("Response from OpenMV:", result)
    else:
        print("No response from OpenMV.")
```
Slave:
```
import rpc
import pyb

red_led = pyb.LED(1)
# Initialize the RPC interface (USB Virtual COM Port)
pyb.LED.on(red_led)
pyb.delay(500)
pyb.LED.off(red_led)
pyb.delay(500)
interface = rpc.rpc_usb_vcp_slave()  # For UART use rpc_uart_slave()
pyb.LED.on(red_led)
pyb.delay(500)
pyb.LED.off(red_led)
pyb.delay(500)
## Define a simple function to respond with a message
def hello_world(data):
    return "Hello from OpenMV!"

## Register the hello_world function to be called over RPC
interface.register_callback(hello_world)

## Main loop to keep the RPC server running
pyb.LED.on(red_led)
pyb.delay(500)
pyb.LED.off(red_led)
pyb.delay(500)
interface.loop()  # Continuously listen for RPC commands

```
Using this pair of codes give me the output of 
```
No response from OpenMV.
No response from OpenMV.
No response from OpenMV.
```

kwagyeman:
H, I'll be able to tackle doing some upgrades on the lib next week.

Issue #2353, opened 2024-08-04
Test RPC Library again


kwagyeman:
https://forums.openmv.io/t/upgrade-to-v4-5-3-rpc-spi-no-longer-working/9306/32?u=kwagyeman

Issue #1672, opened 2022-06-24
Ethernet problems with Arduino H7 and Vision shield


RobiTobi:
If I run the example **image_transfer_jpg_as_the_remote_device_for_your_computer.py** on the camera and **rpc_image_tansfer_jpg_as_the_controller_device.py** on PC, with ethernet connectivity, I get just one or none Image in the pygame window. One image if I run call back `img = get_frame_buffer_call_back("sensor.GRAYSCALE", "sensor.QVGA", cutthrough=False, silent=True)` then the pygame window freezes with no response. I get none picture (black image and 0 FPS) if I run call back `img = get_frame_buffer_call_back("sensor.GRAYSCALE", "sensor.QVGA", cutthrough=True, silent=True)`

1. Configure IPv4 to ethernet adapter (192.168.1.100 / 255.255.255.0)
2. Connect patch cable from ethernet adapter to the Vision Board
3. Connect camera via USB C Cable with the computer
4. Double press hardware reset on H7
5. Connect to camera via OpenMV IDE (in my case COM4)
6. Flash firmware
7. Wait a few secounds after success Messagebox
8. Paste camera script into OpenMV IDE and save the script
9. Upload the Camera script as main.py, without converting (MessageBox -> No)
10. Eject drive from PC
11. Disconnect and reconnect USB
12. Download master branch from git
13. Create new Py Charm project here D:\openmv-master\tools\rpc
14. Overwrite the code of **rpc_image_tansfer_jpg_as_the_controller_device.py** with the PC script and save
15. Run RPC script on PC

I expected to get image by image with every call, which is then shown in the pygame window.

After some testing, it turned out, that putting the interface init from top to the while loop, it gives better result. I get multiple images in low speed.
But just with this call back:
`img = get_frame_buffer_call_back("sensor.GRAYSCALE", "sensor.QVGA", cutthrough=False, silent=True)`

If I try `cutthrough=True` I get no image, as I mentioned before. After switching back to `cutthrough=False` there is the same behavior (no image). If you unplug and reconnect the H7 via USB `cutthrough=False` it is working again.

![2022-06-24 15_34_23-posicom ist egal py – rpc_image_transfer_jpg_as_the_controller_device py](https://user-images.githubusercontent.com/90756169/175548680-53f5cc7b-01bb-4183-9dd0-e225006c7e76.png)

PC script:

```
# Image Transfer - As The Controller Device
#
[...]

PR #2598, opened 2025-02-07
tools/rpc: Remove the desktop RPC library.

kwagyeman:
This code doesn't work well on non-real time systems. Additionally, given USB UART is the connection of choice it makes scripts hard to use. It's better to use the debug protocol scripts or the pyboard controller script.

...

Still deciding on to remove this or not. But, leaning on deleting it. The purpose of the library was to solve microcontroller <-> microcontroller comms and it does that very well. Adding desktop comms to it was a mistake. Leaving this in draft mode until I've thought about it more.

Fixes: https://github.com/openmv/openmv/issues/2594
Fixes: https://github.com/openmv/openmv/issues/2433
Fixes: https://github.com/openmv/openmv/issues/1672
Fixes: https://github.com/openmv/openmv/issues/2353


iabdalkader:
I'd remove it, maybe focus on improving `pyopenmv` publish it to pypi

kwagyeman:
I'm going to give myself a couple of days to think about it and update this during the weekend.

kwagyeman:
On the topic of the desktop RPC library:

* VCP mode gets in the way of debugging and doesn't work reliably on PCs. Reliability could be improved for lower bandwidth by increasing the timeouts in the protocol. But, it getting in the way of debugging and programming makes it a dead-end ease-of-use wise.
*  Hardware UART - People don't have hardware UARTs anymore. They use USB to serial. This has a latency timer which again kills bandwidth of the protocol because timeouts have to be massively increased to be reliable.
* Network mode - Better protocols exist.
* CAN mode - No one ever used this.

As such, there's no need to continue to maintain this library code. Improving pyopenmv, as Ibrahim has mentioned, is more valuable.

PR #2614, opened 2025-02-18
sensors/genx320: Update ISSD sequence to increase frame rate.

kwagyeman:
Changes (from Prophesee):
* New ISSD Sequence increases 24MHz PCLK to 48MHz
* Time base is now setup on the sensor so you can pass arguments in 1MHz units versus needing to be scaled by the clock.

Changes (from OpenMV):
* Fixed the CPI Output control register setup so that defining HSYNC/VSYNC times work.

This PR is needed to increase the RT1062 frame rate: https://github.com/openmv/openmv/pull/2142. While debugging I noticed that the RT1062 drops 1-2 lines randomly every couple of frames leading to frame drops. As you decrease HSYNC lower than 240 the number of lines being dropped per frame starts to grow. By the time you get down to an HSYNC of 32, it's pretty much every other frame is missing lines, resulting in the frame rate being cut in half as our driver drops corrupt frames.

That said, all these frame rates are pointlessly high as you can't actually process 320x320 at above 50 FPS or so.

@nlyubova-psee

### DIFF
--- a/src/omv/sensors/genx320.c
+++ b/src/omv/sensors/genx320.c
@@ -49,8 +49,12 @@
 #define ACTIVE_SENSOR_HEIGHT            (SENSOR_HEIGHT - BLANK_LINES)
 #define ACTIVE_SENSOR_SIZE              (ACTIVE_SENSOR_WIDTH * ACTIVE_SENSOR_HEIGHT)
 
-#define HSYNC_CLOCK_CYCLES              32
-#define VSYNC_CLOCK_CYCLES              32
+#ifdef OMV_GENX320_EHC_ENABLE
+#define HSYNC_CLOCK_CYCLES              880 // 320 + 880 = 1200 cycles -> ~122 FPS
+#else
+#define HSYNC_CLOCK_CYCLES              280 // 320 + 280 = 600 cycles
+#endif
+#define VSYNC_CLOCK_CYCLES              8
 
 #define CONTRAST_DEFAULT                16
 #define BRIGHTNESS_DEFAULT              128
@@ -66,8 +70,8 @@
 #define EVENT_THRESHOLD_TO_CALIBRATE    100000
 #define EVENT_THRESHOLD_SIGMA           10
 
-#define EVT_CLK_FREQ                    ((omv_csi_get_xclk_frequency() + 500000) / 1000000)
-#define EVT_SCALE_PERIOD(period)        (((period) * EVT_CLK_FREQ) / 10)
+#define EVT_CLK_MULTIPLIER              (2)
+#define EVT_CLK_FREQ                    (((omv_csi_get_xclk_frequency() * EVT_CLK_MULTIPLIER) + 500000) / 1000000)
 
 #define AFK_50_HZ                       (50)
 #define AFK_60_HZ                       (60)
@@ -115,6 +119,7 @@ static int reset(omv_csi_t *csi) {
 
     // Set EVT20 mode
     psee_sensor_write(EDF_CONTROL, 0);
+    #endif // (OMV_GENX320_EHC_ENABLE == 1)
 
     // Configure Packet and Frame sizes
     psee_sensor_write(CPI_PACKET_SIZE_CONTROL, ACTIVE_SENSOR_WIDTH);
@@ -122,7 +127,6 @@ static int reset(omv_csi_t *csi) {
                       HSYNC_CLOCK_CYCLES << CPI_PACKET_TIME_CONTROL_BLANKING_Pos);
     psee_sensor_write(CPI_FRAME_SIZE_CONTROL, ACTIVE_SENSOR_HEIGHT);
     psee_sensor_write(CPI_FRAME_TIME_CONTROL, VSYNC_CLOCK_CYCLES);
-    #endif // (OMV_GENX320_EHC_ENABLE == 1)
 
     // Enable dropping
     psee_sensor_write(RO_READOUT_CTRL, RO_READOUT_CTRL_DIGITAL_PIPE_EN |
@@ -166,7 +170,7 @@ static int reset(omv_csi_t *csi) {
     }
 
     if (psee_ehc_activate(&psee_ehc, EHC_ALGO_DIFF3D, 0, EHC_DIFF3D_N_BITS_SIZE,
-                          EVT_SCALE_PERIOD(INTEGRATION_DEF_PREIOD), EHC_WITHOUT_PADDING) != EHC_OK) {
+                          INTEGRATION_DEF_PREIOD, EHC_WITHOUT_PADDING) != EHC_OK) {
         return -1;
     }
     #else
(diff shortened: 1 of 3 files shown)


iabdalkader (changes requested):

nlyubova-psee:
@kwagyeman thank you for debugging

src/omv/sensors/genx320.c
```diff
@@ -49,8 +49,15 @@
...
 
-#define HSYNC_CLOCK_CYCLES              32
-#define VSYNC_CLOCK_CYCLES              32
+#ifndef OMV_GENX320_HSYNC_VALUE
+#define OMV_GENX320_HSYNC_VALUE         280
+#endif
+
+#ifndef OMV_GENX320_VSYNC_VALUE
+#define OMV_GENX320_VSYNC_VALUE         10
+#endif
```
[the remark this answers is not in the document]
kwagyeman: Done

kwagyeman:
@nlyubova-psee - I updated the PR so that the hsync_clocks are now just calculated when you set the frame rate. I found a safe default value for the H7 and RT1062 that works for normal and transposed modes. Transposing the frame tops out at about 120 FPS before falling over. As for non-transposed modes, there's no cap, you can just keep raising the FPS until it fails to capture the data. For the H7 you can go up to about 375 FPS. The RT1062 has a lower cap right now as full hardware offload isn't enabled.

src/omv/sensors/genx320.c
```diff
@@ -49,8 +49,12 @@
...
 #define ACTIVE_SENSOR_SIZE              (ACTIVE_SENSOR_WIDTH * ACTIVE_SENSOR_HEIGHT)
 
-#define HSYNC_CLOCK_CYCLES              32
-#define VSYNC_CLOCK_CYCLES              32
+#ifdef OMV_GENX320_EHC_ENABLE
+#define HSYNC_CLOCK_CYCLES              880 // 320 + 880 = 1200 cycles -> ~122 FPS
+#else
+#define HSYNC_CLOCK_CYCLES              280 // 320 + 280 = 600 cycles
+#endif
+#define VSYNC_CLOCK_CYCLES              8
```
kwagyeman, line 57: Settings per camera type aren't required anymore. These defaults work for the H7 and RT1062 and are loose. set_framerate() then can be called to speed up the FPS and push towards the max of what the system can do.

iabdalkader:
> Transposing the frame tops out at about 120 FPS

The last commit fails on transpose, is this okay?

src/omv/sensors/genx320.c
iabdalkader: But will they be okay for N6?

src/omv/sensors/genx320.c
```diff
@@ -66,8 +70,8 @@
 #define EVENT_THRESHOLD_TO_CALIBRATE    100000
 #define EVENT_THRESHOLD_SIGMA           10
 
-#define EVT_CLK_FREQ                    ((omv_csi_get_xclk_frequency() + 500000) / 1000000)
-#define EVT_SCALE_PERIOD(period)        (((period) * EVT_CLK_FREQ) / 10)
+#define EVT_CLK_MULTIPLIER              (2)
+#define EVT_CLK_FREQ                    (((omv_csi_get_xclk_frequency() * EVT_CLK_MULTIPLIER) + 500000) / 1000000)
```
kwagyeman, line 74: This is doubled since the camera is running at 2x the speed now.

src/omv/sensors/genx320.c
```diff
@@ -166,7 +170,7 @@ static int reset(omv_csi_t *csi) {
     }
 
     if (psee_ehc_activate(&psee_ehc, EHC_ALGO_DIFF3D, 0, EHC_DIFF3D_N_BITS_SIZE,
-                          EVT_SCALE_PERIOD(INTEGRATION_DEF_PREIOD), EHC_WITHOUT_PADDING) != EHC_OK) {
+                          INTEGRATION_DEF_PREIOD, EHC_WITHOUT_PADDING) != EHC_OK) {
```
kwagyeman, line 173: The new ISSD internally makes the camera timebase not need scaling by the clock anymore.

src/omv/sensors/genx320.c
```diff
@@ -231,7 +235,18 @@ static int set_framerate(omv_csi_t *csi, int framerate) {
         return -1;
     }
 
-    psee_sensor_write(EHC_INTEGRATION_PERIOD, EVT_SCALE_PERIOD(us));
+    psee_sensor_write(EHC_INTEGRATION_PERIOD, us);
```
kwagyeman, line 238: The new ISSD internally makes the camera timebase not need scaling by the clock anymore.

src/omv/sensors/genx320.c
```diff
@@ -231,7 +235,18 @@ static int set_framerate(omv_csi_t *csi, int framerate) {
...
+    int lines = ACTIVE_SENSOR_HEIGHT + VSYNC_CLOCK_CYCLES;
+    int clocks_per_frame = (omv_csi_get_xclk_frequency() * EVT_CLK_MULTIPLIER) / framerate;
+    int hsync_clocks = (clocks_per_frame / lines) - ACTIVE_SENSOR_WIDTH;
+
+    if (hsync_clocks <= 0) {
+        return -1;
+    }
+
+    psee_sensor_write(CPI_PACKET_TIME_CONTROL, ACTIVE_SENSOR_WIDTH << CPI_PACKET_TIME_CONTROL_PERIOD_Pos |
+                      hsync_clocks << CPI_PACKET_TIME_CONTROL_BLANKING_Pos);
```
kwagyeman, line 249: Speeds up or slows down the FPS by adjusting the hsync time. As this is pushed higher frame capture will fall over. However, there is no hard and fast limit. When transposing the limit will be lower.

src/omv/sensors/genx320.c
kwagyeman: Should be, the N6 can handle the previous hsync of 32 which was much tighter than 880.

kwagyeman:
> The last commit fails on transpose, is this okay?

For event mode the data is 32-byte packets. So, it's not possible to handle the sensor driver transposing the image. Event mode is not enabled in our firmware by default though. We are in histogram mode which outputs a standard 8-bit image which can be transposed by the sensor driver.

src/drivers/genx320/src/psee_genx320.c
```diff
@@ -3175,7 +3175,7 @@ STC_StatusTypeDef psee_stc_trail_activate(STC_HandleTypeDef *stc, uint32_t stc_t
     }
 
     /* Assert Clock Param */
-    if ((evt_clk_freq != 10) && (evt_clk_freq != 24) && (evt_clk_freq != 25) && (evt_clk_freq != 50)) {
+    if ((evt_clk_freq < 10) || (evt_clk_freq > 50)) {
```
nlyubova-psee, line 3178: same here. This change will not work for STC

src/omv/sensors/genx320.c
nlyubova-psee: Sorry for bothering you. Usually, we configure CPI in the init phase.
I am afraid that we will get some corrupted data while doing it here.

src/drivers/genx320/src/psee_genx320.c
```diff
@@ -2992,7 +2992,7 @@ STC_StatusTypeDef psee_stc_only_activate(STC_HandleTypeDef *stc, uint32_t stc_th
...
```
nlyubova-psee, line 2995: same here

src/drivers/genx320/src/psee_genx320.c
```diff
@@ -3083,7 +3083,7 @@ STC_StatusTypeDef psee_trail_only_activate(STC_HandleTypeDef *stc, uint32_t trai
...
```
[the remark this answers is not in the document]
kwagyeman: Removed

src/drivers/genx320/src/psee_genx320.c
kwagyeman: Removed

src/drivers/genx320/src/psee_genx320.c
kwagyeman: Removed

kwagyeman:
@nlyubova-psee - I dropped the commit you were concerned about. However, this PR must be merged first: https://github.com/openmv/openmv/pull/2617

After that I can rebase this PR on master and it will be ready again.

src/omv/sensors/genx320.c
kwagyeman: I've added in code to flush the DMA buffers and wait until the image settles before restarting them.

nlyubova-psee:
> Please address Ibrahim's feedback.
> 

Do you mean about transpose ?

If yes, then sorry I don't know the solution. Do you ?

kwagyeman:
@nlyubova-psee - I've updated this PR and tested that it's still fin on the H7 Plus and RT1062. 

You or someone from your team should have the final review on this PR before it's merged. It can wait until you get back if no one else from Prophesee is available.

nlyubova-psee:
> @nlyubova-psee - I've updated this PR and tested that it's still fine on the H7 Plus and RT1062.
> 
> You or someone from your team should have the final review on this PR before it's merged. It can wait until you get back if no one else from Prophesee is available.

@kwagyeman I have tested this PR on both RT1062 and H7Plus and it works fine! thank you for your work !

nlyubova-psee (approved):

iabdalkader (approved):

PR #2617, opened 2025-02-20
sensors/GenX320: Adding IOCTL to control Anti-Flicker filter (AFK).

nlyubova-psee:
This PR adds IOCTL for AFK (GenX320 AntiFlicker filter) allowing
* enable AFK with a new range of frequencies (useful for frequency filtering)
* disable AFK (as it is enabled by default and some users needs those filtered out frequencies)
* fixing typos and small cleaning

Tested on a LED (on top of https://github.com/openmv/openmv/pull/2614 ) and works well for filtering out given frequencies.

### DIFF
--- a/src/omv/modules/py_csi.c
+++ b/src/omv/modules/py_csi.c
@@ -1028,6 +1028,15 @@ static mp_obj_t py_omv_csi_ioctl(size_t n_args, const mp_obj_t *args) {
             }
             break;
         }
+        case OMV_CSI_IOCTL_GENX320_SET_AFK: {
+            if (n_args == 2) {
+                error = omv_csi_ioctl(request, mp_obj_get_int(args[1]));
+            } else if (n_args == 4) {
+                error = omv_csi_ioctl(request, mp_obj_get_int(args[1]), mp_obj_get_int(args[2]),
+                                      mp_obj_get_int(args[3]));
+            }
+            break;
+        }
         #endif // (OMV_GENX320_ENABLE == 1)
 
         default: {
@@ -1241,6 +1250,7 @@ static const mp_rom_map_elem_t globals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_GENX320_BIAS_FO),              MP_ROM_INT(OMV_CSI_GENX320_BIAS_FO)},
     { MP_ROM_QSTR(MP_QSTR_GENX320_BIAS_HPF),             MP_ROM_INT(OMV_CSI_GENX320_BIAS_HPF)},
     { MP_ROM_QSTR(MP_QSTR_GENX320_BIAS_REFR),            MP_ROM_INT(OMV_CSI_GENX320_BIAS_REFR)},
+    { MP_ROM_QSTR(MP_QSTR_IOCTL_GENX320_SET_AFK),        MP_ROM_INT(OMV_CSI_IOCTL_GENX320_SET_AFK)},
     #endif
 
     // Sensor functions
--- a/src/omv/sensors/genx320.c
+++ b/src/omv/sensors/genx320.c
@@ -84,7 +84,7 @@ static bool hot_pixels_disabled = false;
 static int32_t contrast = CONTRAST_DEFAULT;
 static int32_t brightness = BRIGHTNESS_DEFAULT;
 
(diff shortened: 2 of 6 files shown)


src/omv/modules/py_csi.c
```diff
@@ -1028,6 +1028,15 @@ static mp_obj_t py_omv_csi_ioctl(size_t n_args, const mp_obj_t *args) {
             }
             break;
         }
+        case OMV_CSI_IOCTL_GENX320_SET_AFK: {
+            if (n_args == 2) {
+                error = omv_csi_ioctl(request, mp_obj_get_int(args[1]));
+            } else if (n_args == 4) {
+                error = omv_csi_ioctl(request, mp_obj_get_int(args[1]), mp_obj_get_int(args[2]),
+                                      mp_obj_get_int(args[3]));
+            }
```
kwagyeman, line 1037: Add error check on args not being 2 or 4.

src/omv/sensors/genx320.c
```diff
@@ -528,6 +528,38 @@ static int ioctl(omv_csi_t *csi, int request, va_list ap) {
             }
             break;
         }
+        // Controlling AFK filter
+        case OMV_CSI_IOCTL_GENX320_SET_AFK: {
+            int mode = va_arg(ap, int);
+            switch (mode) {
```
kwagyeman, line 534: If this is a bool then use an if statement versus case.

src/omv/modules/py_csi.c
iabdalkader: No need, error is set to invalid args.

src/omv/sensors/genx320.c
```diff
@@ -528,6 +528,38 @@ static int ioctl(omv_csi_t *csi, int request, va_list ap) {
...
+                    if (psee_afk_get_state(&psee_afk) != AFK_STATE_RESET) {
+                        if (psee_afk_deactivate(&psee_afk) != AFK_OK) {
+                            ret = -1;
+                        }
+                    }
+                    break;
+                }
+                // Enable AFK
+                case 1: {
+                    int freq_min = va_arg(ap, int);
```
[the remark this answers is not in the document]
nlyubova-psee: why do we need re-initialization ? looks like it works fine without re-initialization.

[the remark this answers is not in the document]
kwagyeman: Does it handle two calls to the IOCTL in a row that change the freq settings? I noticed `(psee_afk_get_state(&psee_afk) != AFK_STATE_RESET)` in the disable method. So, is there any state check that needs to be made or etc.

src/omv/sensors/genx320.c
nlyubova-psee: done

nlyubova-psee:
> Would be good to add an example script in a new commit.

  added

src/omv/sensors/genx320.c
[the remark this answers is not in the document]
nlyubova-psee: it can handle two calls with different frequencies for me one after another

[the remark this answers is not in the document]
nlyubova-psee: I have added the state check to save time and not disable it if it is already disabled.

kwagyeman (approved):

src/omv/sensors/genx320.c
```diff
@@ -84,6 +84,8 @@ static bool hot_pixels_disabled = false;
 static int32_t contrast = CONTRAST_DEFAULT;
 static int32_t brightness = BRIGHTNESS_DEFAULT;
 
+AFK_HandleTypeDef psee_afk;
```
iabdalkader, line 87:
```suggestion
static AFK_HandleTypeDef psee_afk;
```
Is this state really needed? It seems transient and can probably be on the stack instead of reserving memory.

nlyubova-psee: > ```suggestion
> 
> static AFK_HandleTypeDef psee_afk;
> 
> ```
> 
> Is this state really needed? It seems transient and can probably be on the stack instead of reserving memory. Maybe it is to check if it's active. In this case, it should be static.

Yes, I used it to check the state and not re-initialize it several times.
Sorry I cannot test it this week (on vacation without a laptop)

PR #2632, opened 2025-03-24
psee_ehc_activate_override: use the correct accumulation time

skal-hs:
'0' was written as EHC_INTEGRATION_PERIOD

### DIFF
--- a/src/drivers/genx320/src/psee_genx320.c
+++ b/src/drivers/genx320/src/psee_genx320.c
@@ -2517,7 +2517,7 @@ EHC_StatusTypeDef psee_ehc_activate_override(EHC_HandleTypeDef *ehc, EHC_AlgoTyp
         );
 
         /* EHC Accumulation time */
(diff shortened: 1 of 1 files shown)


kwagyeman:
@skal-hs - This is in the driver. So, it's your code and happy to merge. But, could you describe what this does?

kwagyeman:
@nlyubova-psee - Is this a correct fix? If so, please verify it. Thank you,

nlyubova-psee:
It makes sense to me, as this change is made in psee_ehc_activate_**override**() function. I think it is good to keep it with the same config as in psee_ehc_activate(). So, ok for me.

kwagyeman:
Just reviewed the code here and psee_ehc_activate_override is not used at all. This fix doesn't affect  our logic. So, approved. Thanks for sending the fix.

kwagyeman (approved):

PR #2648, opened 2025-04-07
misc/csi: Don't fail if sleep is not implemented.

iabdalkader:
At the very least, sleep will abort the transfer on low-power entry or exit, even if it's not fully supported.


iabdalkader:

https://forums.openmv.io/t/corrupt-images-are-saved-when-lightsleep-is-used/10815/16

kwagyeman (approved):
Looks good to me!

PR #2650, opened 2025-04-08
csi: remove enum to support preprocessor directives

timdechant:
While adding a new port, I found that my `OMV_CSI_CLK_SOURCE` was not effective.  This was due to the value being defined as an `enum`, which is undefined during the preprocessor run.  This means that `OMV_CSI_CLK_SOURCE == OMV_CSI_CLK_SOURCE_TIM` was evaluating to `UNDEFINED == UNDEFINED`, always true.

Changing the `enum` to `#define`s to resolve this.

### DIFF
--- a/src/omv/common/omv_csi.h
+++ b/src/omv/common/omv_csi.h
@@ -95,16 +95,14 @@
 
 #define OMV_CSI_TIMEOUT_MS      (3000)
 
+#define OMV_CSI_CLK_SOURCE_MCO  (0U)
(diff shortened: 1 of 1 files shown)


iabdalkader:
> While adding a new port

Just a heads up I'll be refactoring this whole repo soon, maybe after the next release, to get rid of `src/` dir and have a more flat structure, so many things will move around. It will look something like:
```
boot
lib
modules
drivers
ports # moved from src/omv/ports
etc..
```
While at it, or following that, I will replace recursive use of make with more included `.mk` files.

PR #2651, opened 2025-04-08
ports/alif: Change HP U55 config to SRAM_OPSI.

kwagyeman:
The correct performance information should be reported now when compiling the HP ROMFS img.

Grabbed new ini file from here: https://github.com/alifsemi/alif_ml-embedded-evaluation-kit/blob/main/scripts/vela/ensemble_vela.ini.

...

This is a draft because I need to test the resulting firmware. However, this mirrors the updated files for the IDE.

### DIFF
--- a/ports/alif/alif.mk
+++ b/ports/alif/alif.mk
@@ -144,7 +144,7 @@ $(error Invalid MCU core specified))
 endif
 
 ifeq ($(MCU_CORE),M55_HP)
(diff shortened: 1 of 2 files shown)


kwagyeman:
The resulting romfs file has a different hash than before so the resulting networks are different. However, I tested them and they continue to work. Ethos scheduling of ops was probably adjusted. Performance appears to be the same.

Issue #2652, opened 2025-04-09
New way to prevent I2C fast mode


timdechant:
STM32 HAL defines `I2C_FASTMODEPLUS_I2Cx` even when Fast Mode is not supported:
https://github.com/openmv/openmv/blob/22cbc5d6bbc286f23b0ea98fc66daf79190e27a4/src/hal/stm32/f7/include/stm32f7xx_hal_i2c_ex.h#L73-L77

...which causes an error here:
https://github.com/openmv/openmv/blob/22cbc5d6bbc286f23b0ea98fc66daf79190e27a4/src/omv/ports/stm32/omv_i2c.c#L190-L194

The macro `IS_I2C_FASTMODEPLUS()` or something similar might be helpful here, except that the HAL includes a cast in defining `I2C_FASTMODEPLUS_I2Cx`.  So the preprocessor cannot handle this in any directive!

I suggest adding an option like `OMV_I2C_DISABLE_FASTMODE` which can be defined to entirely bypass this code.

iabdalkader:
> ...which causes an error here:

I think this is correct. Don't you need FMP to achieve the highest speed? If it's not supported it should fail. Note that nothing actually uses `OMV_I2C_SPEED_FAST` except for MLX90x FIR sensors.

timdechant:
It fails to compile.

iabdalkader:
> It fails to compile, so there should be a compile-time method to disable it.

How does it fail to compile? Can you post the error or let me know how to reproduce it? It's very easy to add options, just need to be sure we really need it first.

timdechant:
Right, I skipped a step.

`HAL_I2CEx_EnableFastModePlus()` is called always, but is only defined if FastMode is supported:
https://github.com/openmv/openmv/blob/22cbc5d6bbc286f23b0ea98fc66daf79190e27a4/src/hal/stm32/f7/include/stm32f7xx_hal_i2c_ex.h#L116-L119

iabdalkader:
> `HAL_I2CEx_EnableFastModePlus()` is called always, but is only defined if FastMode is supported:

Same as `IS_I2C_FASTMODEPLUS` (at least for F7, H7 and N6). What if we just test for that instead?

timdechant:
`#if IS_I2C_FASTMODEPLUS(I2C_FASTMODEPLUS_I2Cx)` fails because it includes a cast.

iabdalkader:
[...]

PR #2653, opened 2025-04-09
i2c: disable fastmode at compile

timdechant:
Implements #2652

### DIFF
--- a/src/omv/ports/stm32/omv_i2c.c
+++ b/src/omv/ports/stm32/omv_i2c.c
@@ -184,7 +184,7 @@ int omv_i2c_init(omv_i2c_t *i2c, uint32_t bus_id, uint32_t speed) {
         return -1;
     }
 
-    #if !defined(OMV_I2C_DISABLE_FASTMODE)
+    #if defined(IS_I2C_FASTMODEPLUS)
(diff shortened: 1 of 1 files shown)


src/omv/ports/stm32/omv_i2c.c
```diff
@@ -184,6 +184,7 @@ int omv_i2c_init(omv_i2c_t *i2c, uint32_t bus_id, uint32_t speed) {
         return -1;
     }
 
+    #if !defined(OMV_I2C_DISABLE_FASTMODE)
```
iabdalkader, line 187:
```suggestion
    #if defined(IS_I2C_FASTMODEPLUS)
```

This should work, and you won't need to define new options.

iabdalkader: > It seems like we could still run into the poisonous cast in the preprocessor; should that be a runtime check?

This only tests if the macro is defined, which is true if FMP is supported, otherwise it's not defined at all.
```
#if defined(IS_I2C_FASTMODEPLUS)
if (speed == OMV_I2C_SPEED_FAST) {
....
```

iabdalkader (approved):

PR #2654, opened 2025-04-09
ports/stm32: assert i2c not SPEED_FAST if not supported

timdechant:
Complements #2653. 

One more detail for #2652.


iabdalkader:
Looks good. However, note the scope in your PRs should be `ports/stm32: commit title`. This helps it gets categorized correctly in the change log.

iabdalkader (approved):

PR #2669, opened 2025-04-21
imlib: Refactor framebuffer API to accept a context.

iabdalkader:


modules/py_csi.c
```diff
@@ -131,7 +131,7 @@ static mp_obj_t py_omv_csi_shutdown(mp_obj_t enable) {
 static MP_DEFINE_CONST_FUN_OBJ_1(py_omv_csi_shutdown_obj, py_omv_csi_shutdown);
 
 static mp_obj_t py_omv_csi_flush() {
-    framebuffer_update_jpeg_buffer();
+    framebuffer_update_jpeg_buffer(csi.fb);
```
kwagyeman, line 134: Shouldn't this use `framebuffer_t *fb = framebuffer_get(0);` so you can specify the frame buffer to flush later on?

modules/py_csi.c
```diff
@@ -194,12 +194,13 @@ static mp_obj_t py_omv_csi_height() {
 static MP_DEFINE_CONST_FUN_OBJ_0(py_omv_csi_height_obj, py_omv_csi_height);
 
 static mp_obj_t py_omv_csi_get_fb() {
-    if (framebuffer_get_depth() < 0) {
+    image_t image;
+
+    if (framebuffer_get_depth(csi.fb) < 0) {
```
kwagyeman, line 199: Shouldn't this use `framebuffer_t *fb = framebuffer_get(0);` so you can specify the frame buffer to get later on?

kwagyeman (approved):

kwagyeman:
Everything else looks good.

modules/py_csi.c
iabdalkader: No. `framebuffer_get(0);` is for when you need a frame buffer and don't have access to a CSI instance, for example `py_fir`, `py_tof`. This function will eventually receive a CSI ID.

PR #2685, opened 2025-05-07
boot: Add DFU partition delay setting.

iabdalkader:
Allows setting a custom delay for DFU partition. This should be set to ensure control transfers don't timeout if erasing the partition takes too long. By default, this is 0 for all partition types.

I did not check every possible flash type (we have boards with QSPI, OSPI, NOR and MRAM), but I think a default of 0 might be okay, as the slowest would be NOR (at around 2.5 seconds). For all SPI flashes, we do block erases (not sector not full erase). The recovery type/process for Everspin may need a little longer, though I don't remember it timing out

@kwagyeman Please check the boards and let me know which partitions need to have more delay, or just send a follow-up PR.

For all ST chips we use Program/erase parallelism = x32
STM32F427xx/STM32F429xx
image

STM32F765xx STM32F767xx STM32F768Ax STM32F769xx:
image

STM32H742xI/G STM32H743xI/G
image

W25Q256JV (4P):

image

Macronix OSPI (N6/AE3):
image

MRAM (AE3): Almost instantaneous.

### DIFF
--- a/boards/OPENMV_AE3/romfs.json
+++ b/boards/OPENMV_AE3/romfs.json
@@ -22,13 +22,7 @@
     },
     {
       "type": "tflite",
-      "path": "{TOP}/lib/models/yolo_v5_224_nano.tflite",
-      "alignment": 16,
-      "optimize": "Performance"
-    },
-    {
-      "type": "tflite",
-      "path": "{TOP}/lib/models/person_detect.tflite",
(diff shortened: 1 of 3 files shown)


kwagyeman (approved):

kwagyeman:
I also noticed the M4 firmware was definitely heavily broken and needed that previous fix that was merged.

Issue #2676, opened 2025-04-28
Thoughts on dual CSI support.


iabdalkader:
This is not meant to support two actual main sensors, I don't know if it can. It's intended to support an auxiliary or secondary sensor with a different interface (like the Lepton).

Possible solutions:

1- Duplicate most of the csi struct. To avoid duplicating `i2c_bus` and port bits (such as dma, dcmi etc..), which will be unset/unused for `csi1`, use pointers set to common static instances when initialized.

```C
// Port bits (such as dcmi, mdma etc..).
typedef struct _omv_csi_priv {
#ifdef OMV_CSI_PORT_BITS
    OMV_CSI_PORT_BITS
#endif
} omv_csi_priv_t;

static omv_csi_priv_t omv_csi_priv;

typedef struct _omv_csi_dev {
    bool primary;     // Not really needed could test priv==NULL

    uint32_t chip_id;
    uint8_t slv_addr;

    omv_i2c_t *i2c_bus;

    omv_csi_priv_t *priv;  // Can be NULL for csi1.

    // Sensor function pointers
    int (*reset) (omv_csi_t *csi);
    int (*sleep) (omv_csi_t *csi, int enable);
    int (*snapshot) (omv_csi_t *csi, image_t *image, uint32_t flags);
    ....
} omv_csi_t;

omv_csi_t csi[OMV_CSI_DEV_COUNT] = { 0 };

void omv_csi_init_default() {
    for (i=0; i<csi_count; i++) {
        csi[i].i2c_bus = &csi_i2c;
    }
    csi[0].priv = &omv_csi_priv;
}
```

This can be used like:

```C
for (i=0; i<csi_count; i++) {
    csi[i].fb = framebuffer_get(i);
    csi[i].snapshot = omv_csi_snapshot; // will be changed for dev[1] in lepton for example.
}

void DCMI_IRQHandler(void) {
    HAL_DCMI_IRQHandler(&csi[0].priv->dcmi);    // or csi[0].dcmi if duplicated.
}

// id passed from `py_csi_sleep(id=0)`
__weak int omv_csi_sleep(size_t id, int enable) {
    omv_csi_abort(&csi[id], true, false);
    if (csi[id].sleep != NULL && csi[id].sleep(&csi[id], enable) != 0) {
        return OMV_CSI_ERROR_CTL_FAILED;
    }
    return 0;
}
```

==========================================

2- Using a csi struct and sub-devices:

```C
typedef struct _omv_csi_dev {
[...]

PR #2691, opened 2025-05-16
ports/all: Dual CSI support.

iabdalkader:
A first draft of dual-CSI support (see https://github.com/openmv/openmv/issues/2676). This is based on N6 support, see commits starting at `imlib/framebuffer: Add support for a second frame buffer.`

The main changes are:

- Support for defining a second framebuffer. The main/default FB is used for drawing, fb_alloc etc...
- All CSI functions now accept an instance with the exception of clock config, which can be made an op and passed an instance too.
-  A new CSI API (highly subject to change) is introduced. The API allows using the main CSI, secondary CSI or both at the same time, with the ability to control which one updates the JPEG buffer.
- The old `sensor` module is still supported and both the new and old APIs are enabled.
- Note to minimize changes, only the enable sensors were fixed/updated

I've enabled Lepton, both PixArt sensors and Soft-CSI. Lepton + PAG should work, but it's untested. If a single sensor is connected, soft-csi gets added automatically, so you could run the following example as is:

```Python
import time
import csi

csi0 = csi.CSI()
csi0.reset(hard=True)
csi0.pixformat(csi.RGB565)
csi0.framesize(csi.QVGA)
csi0.snapshot(time=100)

csi1 = csi.CSI(cid=csi.SOFTCSI)
csi1.reset(hard=False)
csi1.pixformat(csi.RGB565)
csi1.framesize(csi.QVGA)
csi0.snapshot(time=100)

clock = time.clock()
while True:
    clock.tick()
    img0 = csi0.snapshot()
    img1 = csi1.snapshot(update=False)
    img0.draw_image(img1, 0, 0, x_scale=0.5, y_scale=0.5)
    print(clock.fps())
```

Technically, you could also enable 3 CSIs, but you'll need a third, statically-defined FB. So I'm considering making FBs dynamically allocated, possibly only for secondary CSIs. This will also fix the FB size issue. Thoughts?

Also, I'm considering replacing most of the functions with attributes. For example:

```Python
csi0 = csi.CSI()
csi0.reset(hard=True)
csi0.pixformat = csi.RGB565
csi0.framesize = csi.QVGA
csi0.window = (0, 0, 100, 100)
csi0.delays = False
csi0.snapshot(time=100)
```

They're smaller, faster, easier to maintain. Thoughts?

---

A note about the ops: More functions have been made ops (such as, abort, config and snapshot).

1- In port's main CSI driver (for example: `ports/omv_csi.c`), the main driver's ops are initially set as defaults:

```C
for (size_t i=0; i<OMV_CSI_MAX_DEVICES; i++) {
    omv_csi_t *csi = &csi_all[i];

    memset(csi, 0, sizeof(omv_csi_t));
    csi->fb = framebuffer_get(i);
    csi->i2c_bus = &i2c_bus;

    csi->abort = stm_csi_abort;         // main driver's abort (disables DCMI, DMA etc..)
    csi->config = stm_csi_config;       // main driver's config 
    csi->snapshot = stm_csi_snapshot;   // main driver's snapshot.

    csi->color_palette = rainbow_table;
}
```

This allows something like Lepton and Soft-CSI to set abort/config to NULL, which will not configure DCMI/DMA and will do the general `abort()` part but Not disable DMA/DCMI. These driver were never supposed to call stm32's config/abort anyway, because they don't use DCMI/DMA, but now with dual-csi config, these structs will not even be initialized for the secondary/aux sensors.

While we're at it, made main driver's snapshot an op too. Drivers like boson that need to both override the main snapshot, and still call it later, can get the main snapshot by stashing the pointer in their init functions:

```C
typedef int (*omv_csi_snapshot_t)(omv_csi_t *csi, image_t *image, uint32_t flags);

static int snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
    // Call `csi->priv()` to get an image using the main driver's snapshot.
    int ret = ((omv_csi_snapshot_t) csi->priv)(csi, image, flags);
    ....
}

int boson_init(omv_csi_t *csi) {
    // Initialize csi structure
    csi->reset = reset;
    csi->set_pixformat = set_pixformat;
    //etc...

    csi->priv = csi->snapshot
    csi->snapshot = snapshot;
```


kwagyeman:
> Technically, you could also enable 3 CSIs, but you'll need a third, statically-defined FB. So I'm considering making FBs dynamically allocated, possibly only for secondary CSIs. This will also fix the FB size issue. Thoughts?

Yes, that's perfect! Also, with the new information we received from ST about the AI hyperRAM not needed to be pegged we can just make the heap even bigger.

Note, to support 5MP camera sensors we want the framebuffer+fb_alloc to hold 20MB+.

> Also, I'm considering replacing most of the functions with attributes. For example:

Prefer functions. The new pythonic methods are perfect.

> While we're at it, made main driver's snapshot an op too. Drivers like boson that need to both override the main snapshot, and still call it later, can get the main snapshot by stashing the pointer in their init functions:

Nice.

common/omv_csi.c
```diff
@@ -129,197 +139,274 @@ uint16_t resolution[][2] = {
...
-    csi.vflip = false;
-    csi.transpose = false;
+    csi->sde = 0;
+    csi->pixformat = 0;
+    csi->framesize = 0;
+    csi->framerate = 0;
+    csi->first_line = false;
+    csi->drop_frame = false;
+    csi->last_frame_ms = 0;
+    csi->last_frame_ms_valid = false;
```
kwagyeman, line 233: Already zeroed. Doesn't need to be changed but it's somewhat of an issue to know which variables to clear in the driver at which times. I like how abort clears it's related variables and etc.

common/omv_csi.c
```diff
@@ -651,141 +785,135 @@ __weak int omv_csi_shutdown(int enable) {
...
+    if (((pixformat == PIXFORMAT_YUV422) && (csi->transpose || csi->auto_rotation)) ||
+        ((pixformat == PIXFORMAT_JPEG) && (omv_csi_get_cropped(csi) || csi->transpose || csi->auto_rotation))) {
         return OMV_CSI_ERROR_PIXFORMAT_UNSUPPORTED;
     }
 
     // Disable any ongoing frame capture.
-    omv_csi_abort(&csi, true, false);
-
-    // Flush previous frame.
-    framebuffer_update_jpeg_buffer(csi.fb);
```
kwagyeman, line 711: This seems to be missing. Previous behavior allows you to snapshot(), change the resolution/etc. and then snapshot() again and you see all the frames. Since frame buffer update is async, though, it's not a dependable mechanism to see the last frame.

common/omv_csi.c
```diff
@@ -651,141 +785,135 @@ __weak int omv_csi_shutdown(int enable) {
...
+    if (csi->framesize == framesize) {
         // No change
         return 0;
     }
 
     // Disable any ongoing frame capture.
-    omv_csi_abort(&csi, true, false);
-
-    // Flush previous frame.
-    framebuffer_update_jpeg_buffer(csi.fb);
```
kwagyeman, line 750: Also here.

common/omv_csi.c
```diff
@@ -865,493 +993,487 @@ __weak uint32_t omv_csi_get_dst_bpp() {
...
-    if (csi.pixformat == PIXFORMAT_JPEG) {
+    if (csi->pixformat == PIXFORMAT_JPEG) {
         return OMV_CSI_ERROR_PIXFORMAT_UNSUPPORTED;
     }
 
     // Disable any ongoing frame capture.
-    omv_csi_abort(&csi, true, false);
-
-    // Flush previous frame.
-    framebuffer_update_jpeg_buffer(csi.fb);
```
kwagyeman, line 883: And here.

common/omv_csi.c
```diff
@@ -865,493 +993,487 @@ __weak uint32_t omv_csi_get_dst_bpp() {
...
 }
 
-__weak int omv_csi_set_framebuffers(int count) {
+__weak int omv_csi_set_framebuffers(omv_csi_t *csi, int count) {
     // Disable any ongoing frame capture.
-    omv_csi_abort(&csi, true, false);
+    omv_csi_abort(csi, true, false);
 
-    // Flush previous frame.
-    framebuffer_update_jpeg_buffer(csi.fb);
```
kwagyeman, line 1227: And here.

drivers/sensors/lepton.c
```diff
@@ -495,12 +509,16 @@ int lepton_init(omv_csi_t *csi) {
...
+    csi->auxiliary = 1;
     csi->vsync_pol = 1;
     csi->hsync_pol = 0;
     csi->pixck_pol = 0;
     csi->frame_sync = 0;
     csi->mono_bpp = 1;
 
+    // Extra delay after power-on
+    mp_hal_delay_ms(1000);
+
```
kwagyeman, line 534: Already in reset. Not needed anymore?

common/omv_csi.c
iabdalkader: So what is requested here? Could you be more specific?

drivers/sensors/lepton.c
iabdalkader: It seems to get stuck if you attempt to use I2C before a reasonable delay. For example if you remove the delay after `LEP_RunOemReboot` it fails to boot, even though I2C is polled. I don't understand why this happens. You could test and let me know what timeouts are needed. This is with Lepton 2.5 btw.

common/omv_csi.c
iabdalkader: I'm not sure, but I think I removed that before adding the `update` flag. I think I can safely revert it.

iabdalkader: Okay now I remember why this had to be removed, it will force the second sensor to flush an image to the JPEG buffer, if it calls any of those functions. I don't know how you can snapshot, get the frame, switch res, and snapshot again, maybe flush manually?

modules/py_csi_ng.c
```diff
@@ -0,0 +1,1511 @@
...
+            if (error == OMV_CSI_ERROR_WOULD_BLOCK &&
+                (flags & OMV_CSI_CAPTURE_FLAGS_NBLOCK)) {
+                return mp_const_none;
+            }
+            omv_csi_raise_error(error);
+        }
+
+        // If an image is provided update it and return.
+        if (args[ARG_image].u_obj != mp_const_none) {
+            image_copy(py_helper_arg_to_image(args[ARG_image].u_obj, ARG_IMAGE_ANY), py_image_cobj(image));
```
kwagyeman: Mmm, I thought the copy was a deep copy. Shallow makes sense for performance reasons but we don't really have an API for shallow copies. New image objects in python are full buffers. So, I guess after making an image object to hold the new image it's data would be free'd by GC? Or was this meant to be a deep copy?

common/omv_csi.c
kwagyeman: I see, yes, it would cause bad flushes to happen. I guess it could be dropped. Technically, any flush without a timeout is a race condition anyway.

On that note, something to do is to add flushing with a timeout that would delay until the IDE grabs the frame or timeout. This could be in another PR.

drivers/sensors/lepton.c
kwagyeman: Will check. Not sure if I have the 2.5 still...

modules/py_csi_ng.c
iabdalkader: No, this was meant to be a deep copy, I thought `image_copy` did that. Do we have a function for deep copy that will also check if dst size is enough? Also, I realized this code is creating Python image for no reason when it times out or when an image is provided... Let me optimize that.

ports/stm32/omv_csi.c
```diff
@@ -324,73 +312,74 @@ int omv_csi_config(omv_csi_config_t config) {
...
+    }
+    for (size_t i=0; i<DCMIPP_NUM_OF_PIPES; i++) {
+        csi->dcmi.PipeState[i] = HAL_DCMIPP_PIPE_STATE_RESET;
+    }
+    #endif
 
     return 0;
 }
 
 uint32_t omv_csi_get_clk_frequency() {
```
kwagyeman, line 359: Should get/set clk freq have a csi object? I noticed they seem to be excluded. I guess they are always the primary? It may be better to have overrides like everything else. Like, for example, with the Lepton+PAG7936 board the lepton has it's own 25MHz clock source. That freq could be hardcoded into the firmware to be returned when it's an aux sensor. That said, I don't think anyone will actually care about this ever... so, the current approach is probably fine until there's an issue requiring it to be changed.

modules/py_csi_ng.c
kwagyeman: No, not in the imlib directly. It's always implemented in py_image. No function generically does the operation without being built into something else.

iabdalkader: I've optimized it to only create Python images when needed and I added a flag to image_copy to do deep copies, can you implement it in another PR? It's safer to deep-copy the image than to keep a reference to an FB like that. Note that `image_copy` doesn't seem to be used anywhere, why would it it's just `*img=*image`.

kwagyeman (commented):
As a follow on PR I think re-targeting the original csi code to target csi ng. Would be useful to avoid code duplication.

modules/py_csi_ng.c
kwagyeman: Sure, after the PR is merged or now?

ports/stm32/omv_csi.c
iabdalkader: I've thought about that, but quickly realized it will only lead to set_clk getting called more times than it is right now. Also, the only reason to override those is if a sensor requires a different clock source, but you won't know that until the sensor is clocked and detected.

iabdalkader:
> As a follow on PR I think re-targeting the original csi code to target csi ng. Would be useful to avoid code duplication.

It will be removed after a release or two. New examples should use the new API.

PR #2692, opened 2025-05-20
boards/OPENMV_AE3: Update ROMFS size.

iabdalkader:


kwagyeman (approved):
Works great! Was able to load 20 MB of models, tiny yolo and the face attribute network and both ran without issues. Make 24MB ROMFS so for the N6 too! :)

PR #2693, opened 2025-05-24
modules/py_ml: predict() speedup.

kwagyeman:
WIP


kwagyeman:
This is basically done. Just need to do some more testing on it.

PR #2695, opened 2025-05-30
scripts/examples: Fix Haar Cascade Paths.

kwagyeman:
People be complaining.


iabdalkader:
> Wait, there's more paths...

Yep.

PR #2713, opened 2025-06-16
scripts/libraries: Add support for the SSD1351 OLED display.

kwagyeman:
Adds support for the OLED breakout for the OpenMV AE3.


modules/py_spi_display.c
```diff
@@ -102,6 +102,20 @@ static void spi_display_command(py_display_obj_t *self, uint8_t cmd, uint8_t arg
...
+static void spi_display_command_overridable(py_display_obj_t *self, uint8_t cmd, uint8_t arg, qstr attr) {
+    if (self->controller != mp_const_none) {
+        mp_obj_t dest[3];
+        mp_load_method_maybe(self->controller, attr, dest);
+        if (dest[0] != MP_OBJ_NULL) {
+            dest[2] = MP_OBJ_FROM_PTR(self);
+            mp_call_method_n_kw(1, 0, dest);
+            return;
+        }
+    }
```
iabdalkader: Just move this to `spi_display_command`

ports/alif/omv_spi.c
```diff
@@ -92,11 +92,33 @@ int omv_spi_transfer_start(omv_spi_t *spi, omv_spi_transfer_t *xfer) {
...
         if (!spi_descr->is_lp) {
-            spi_transfer_blocking(spi->inst, &spi_xfer);
+            // TODO: Debug why spi_send_blocking/spi_receive_blocking don't work.
+            // if (spi->bus_mode == OMV_SPI_BUS_TX || xfer->rxbuf == NULL) {
+            //     spi_xfer.mode = SPI_TMOD_TX;
+            //     spi_send_blocking(spi->inst, &spi_xfer);
+            // } else if (spi->bus_mode == OMV_SPI_BUS_RX || xfer->txbuf == NULL) {
+            //     spi_xfer.mode = SPI_TMOD_RX;
+            //     spi_receive_blocking(spi->inst, &spi_xfer);
+            // } else if (spi->bus_mode == OMV_SPI_BUS_TX_RX) {
```
iabdalkader:
```suggestion
```
Please remove these.

ports/alif/omv_spi.c
```diff
@@ -92,11 +92,33 @@ int omv_spi_transfer_start(omv_spi_t *spi, omv_spi_transfer_t *xfer) {
...
         } else {
-            lpspi_transfer_blocking(spi->inst, &spi_xfer);
+            // TODO: Debug why spi_send_blocking/spi_receive_blocking don't work.
+            // if (spi->bus_mode == OMV_SPI_BUS_TX || xfer->rxbuf == NULL) {
+            //     spi_xfer.mode = SPI_TMOD_TX;
+            //     lpspi_send_blocking(spi->inst, &spi_xfer);
+            // } else if (spi->bus_mode == OMV_SPI_BUS_RX || xfer->txbuf == NULL) {
+            //     spi_xfer.mode = SPI_TMOD_RX;
+            //     lpspi_receive_blocking(spi->inst, &spi_xfer);
+            // } else if (spi->bus_mode == OMV_SPI_BUS_TX_RX) {
```
iabdalkader:
```suggestion
```

Remove.

iabdalkader (changes requested):
We should not patch the SPI driver. I was planning on reimplementing spi_transfer_blocking at some point, to poll events but also handle timeout. We should also send add it to machine_spi.c. See comment here:

https://github.com/micropython/micropython/blob/master/ports/alif/machine_spi.c#L308

scripts/examples/50-OpenMV-Boards/52-Alif-Boards/51-OLED-Shield/oled_display_with_joystick.py
```diff
@@ -0,0 +1,71 @@
...
+
+def read_expander():
+    data = i2c.readfrom(I2C_EXPANDER_ADDR, 1)[0] ^ 255
+    right = (data & 1)
+    up = ((data >> 1) & 1)
+    left = ((data >> 2) & 1)
+    down = ((data >> 3) & 1)
+    center = ((data >> 4) & 1)
+    lcd_on = ((data >> 5) & 1)
+    return right, up, left, down, center, lcd_on
```
iabdalkader: This should be a frozen module and it should use a real IRQ pin. Best to remove this for now if not ready.

iabdalkader (changes requested):

iabdalkader:
Please drop the first 3 commits and rebase on this: https://github.com/openmv/openmv/pull/2714. Also, you should use a reasonable timeout not the max.

scripts/examples/50-OpenMV-Boards/52-Alif-Boards/51-OLED-Shield/oled_display_with_joystick.py
kwagyeman: Dropping example.

kwagyeman:
Leaving this here for safe keeping:

```
# This work is licensed under the MIT license.
# Copyright (c) 2013-2025 OpenMV LLC. All rights reserved.
# https://github.com/openmv/openmv/blob/master/LICENSE
#
# OLED Display with Joystick Example
#
# Note: To run this example you will need a OLED Breakout Board for your OpenMV AE3
#
# The OLED Breakout Board allows you to view your OpenMV AE3's frame buffer on the go.

import sensor
import time
import display
from machine import I2C
from machine import Pin

I2C_EXPANDER_ADDR = 63

sensor.reset()
sensor.set_windowing((400, 400))

i2c = I2C(1)
exp_int_pin = Pin("P9", Pin.IN)
lcd = display.SPIDisplay(width=128, height=128, controller=display.SSD1351())
clock = time.clock()


def expander_has_event():
    return exp_int_pin.value() ^ 1


def read_expander():
    data = i2c.readfrom(I2C_EXPANDER_ADDR, 1)[0] ^ 255
    right = (data & 1)
    up = ((data >> 1) & 1)
    left = ((data >> 2) & 1)
    down = ((data >> 3) & 1)
    center = ((data >> 4) & 1)
    lcd_on = ((data >> 5) & 1)
    return right, up, left, down, center, lcd_on

x_scale_def = 128 / 400
y_scale_def = 128 / 400
cursor_x = 0
cursor_y = 0

while True:
    clock.tick()

    if expander_has_event():
        right, up, left, down, center, lcd_on = read_expander()

        if right:
            cursor_x += 2
        if left:
            cursor_x -= 2
        if up:
            cursor_y -= 2
        if down:
            cursor_y += 2
        if center:
            cursor_x = 0
            cursor_y = 0

    lcd.write(sensor.snapshot(), x=cursor_x, y=cursor_y,
              x_scale=x_scale_def, y_scale=y_scale_def)

    print(clock.fps())
```

ports/alif/omv_spi.c
kwagyeman: Removed

ports/alif/omv_spi.c
kwagyeman: Removed

modules/py_spi_display.c
kwagyeman: Done

kwagyeman:
Removing timeout and then retesting...

kwagyeman:
@iabdalkader 

My display seems quite stable. It's been running without crashing for 10 minutes now. I think the failing displays you were dealing with are just the connector. This will be fixed on the second rev.

Anyway, the PR is ready. 

Do you have time to tackle adding the SPI event polling to MicroPython and adding Pin interrupt support for the AE3? Moving on to finishing the PR for performance optimizations, then AE3 Image Quality, followed by GENX320 event mode support next.

iabdalkader:
> Do you have time to tackle adding the SPI event polling to MicroPython and adding Pin interrupt support for the AE3?

Both are done, please fix the joystick driver and add it here. It should accept a user callback, could use hard or soft IRQ probably doesn't matter.

> Moving on to finishing the PR for performance optimizations, then AE3 Image Quality, followed by GENX320 event mode support next.

Before you do, just one comment on this PR. Could you please switch on the command instead of passing the QSTR? Also, why not enable the display controller for RT1060 too? Works great with it and it has the flash space (possibly also 4P, etc.. will leave this to your discretion).

kwagyeman:
Done. However, I noticed something with the IRQ. An uncaught exception during it did not stop the script from running. In either soft or hard interrupt mode. Might be a bug.

iabdalkader:
> Done. However, I noticed something with the IRQ. An uncaught exception during it did not stop the script from running. In either soft or hard interrupt mode. Might be a bug.

Does it stop the script on other boards? Unhandled exceptions in IRQ callbacks just disable the callback, see `mpirq.c`

modules/py_spi_display.c
```diff
@@ -102,8 +104,18 @@ static int spi_write(py_display_obj_t *self, uint8_t cmd, uint8_t *args, size_t
     return 0;
 }
 
-static void spi_display_command(py_display_obj_t *self, uint8_t cmd, uint8_t arg) {
-    spi_write(self, cmd, &arg, (arg > 0) ? 1 : 0, false);
+static void spi_display_command(py_display_obj_t *self, uint8_t cmd, uint8_t arg, qstr attr) {
```
iabdalkader: Instead of passing the qstr, you could just map the command to qstr here:
```C
switch (cmd) {
    case LCD_COMMAND_RAMWR:
        attr = MP_QSTR_ram_write;
        break;
....
````
This way you don't have to pass `0` qstrs.

File: scripts/libraries/pca9674a.py [hunk 6]
```diff
@@ -0,0 +1,49 @@
...
+        self.bus = bus
+        self.address = address
+        self.callback = callback
+        self.irq_pin = None
+        self.irq_pin_label = irq_pin
+        self.state = bytearray(1)
+        self.reset()
+
+    def write(self, value):
+        self.bus.writeto(self.address, bytes[(value)])
```
iabdalkader:
```suggestion
        self.state[0] = value
        self.bus.writeto(self.address, self.state)
```
Saves many small allocs.

modules/py_spi_display.c
```diff
@@ -37,9 +37,11 @@
 #include "omv_spi.h"
 #include "py_display.h"
 
-#define LCD_COMMAND_DISPOFF         (0x28)
-#define LCD_COMMAND_DISPON          (0x29)
-#define LCD_COMMAND_RAMWR           (0x2C)
+#define LCD_COMMAND_OVERRIDABLE     (0x80)
```
iabdalkader: Is this necessary? See the other comment, if you switch on command and qstr is set, then the controller should provide it, if not it should raise an exception (so just use load instead of load_maybe).

```C
if (self->controller != mp_const_none) {
    qstr attr = 0; // is 0 here valid ?
    switch (cmd) {
        case LCD_COMMAND_RAMWR:
            attr = MP_QSTR_ram_write;
            break;
        default:
            attr = 0;
    }

    if (qstr) {
        mp_load_method(....)
    }
}
```

modules/py_spi_display.c
```diff
@@ -54,7 +56,7 @@ static void spi_transmit(py_display_obj_t *self, uint8_t *txdata, uint16_t size)
     omv_spi_transfer_t spi_xfer = {
         .txbuf = txdata,
         .size = size,
-        .timeout = OMV_SPI_MAX_TIMEOUT,
+        .timeout = 5000,
```
iabdalkader:
```suggestion
#ifndef OMV_SPI_DISPLAY_TIMEOUT
#define OMV_SPI_DISPLAY_TIMEOUT (1000)
#endif
        .timeout = OMV_SPI_DISPLAY_TIMEOUT,
```
I'd use 1 sec.

modules/py_spi_display.c
kwagyeman: Done

scripts/libraries/pca9674a.py
kwagyeman: Done

modules/py_spi_display.c
kwagyeman: load_maybe is correct, as overriding is optional. Otherwise, LCD_COMMAND_OVERRIDABLE removed.

modules/py_spi_display.c
kwagyeman: Done. However, I put it at 5 seconds so you could slow the clock freq down to like 200Khz via a refresh of 1 hz on the display for debugging.

modules/py_spi_display.c
```diff
@@ -103,6 +103,33 @@ static int spi_write(py_display_obj_t *self, uint8_t cmd, uint8_t *args, size_t
...
+            case LCD_COMMAND_RAMWR:
+                attr = MP_QSTR_ram_write;
+                break;
+            default:
+                break;
+        }
+
+        if (attr) {
+            mp_obj_t dest[3];
+            mp_load_method_maybe(self->controller, attr, dest);
```
iabdalkader, line 124: This will not fail if the attr doesn't exit, should use `mp_load_method`. If we get here it should fail, right?

modules/py_spi_display.c
iabdalkader: Ah okay, so those commands are still implemented in the fallback in C? Or are they all just optional?

modules/py_spi_display.c
iabdalkader: At 1Hz, will it take 5 seconds to send a frame?

modules/py_spi_display.c
kwagyeman: If the attribute doesn't exist, we just fall back to the C code. Some of our customers have already written display controllers. So, I'm trying not to break their code unless they want to override the default commands.

```
// no attribute found, returns:     dest[0] == MP_OBJ_NULL, dest[1] == MP_OBJ_NULL
```

So the `if (dest[0] != MP_OBJ_NULL) {` will fail and you'll fall back to the C code.

modules/py_spi_display.c
kwagyeman: No, like 2 at 128x128, at 320x240 would be more like 5. Was just setting it higher than normal to avoid it not working for whatever reason in the future. 1 second timeout is fine though, it's unlikely anyone will lower the fresh rate to 1 hz.

iabdalkader: Maybe we could add a timeout arg, but right now the timeout is the only to detect a stuck SPI transfer.

iabdalkader (approved):

PR #2714, opened 2025-06-16
ports/alif: Implement spi_transfer with poll and timeout.

iabdalkader:


iabdalkader:
Ideally we should try DMA and see if it works.

PR #2603, opened 2025-02-09
ports/stm32: Add N6 support.

iabdalkader:

PR #2718, opened 2025-06-20
boards/OPENMV_N6: Update MSC_BUF size to 4KB.

kwagyeman:
The first commit fixes a critical bug. The second commits don't appear to fix anything but maybe should be added.


iabdalkader:
* The SPI issue was indeed due to the GPDMA channels being disabled by `spi_deinit_all` as I've pointed out. Now FLIR is working again.
* The RW filesystem and ROMFS offsets and sizes were wrong, which probably corrupted ROMFS. I actually fixed that in older branches, but that changes must have been dropped when doing the last update.
* The secure address is not needed, it can access secure/ns address all the same. I prefer to leave the 
familiar address but it doesn't really make any difference.
* I enabled CLK sleep for GPIOs that are actually used, the rest are enabled by MicroPython, but also doesn't make any difference.

* The board crash on NPU is unrelated to NPU, it turns out it's due to your last minute change request to update the FB and GC block sizes: https://github.com/openmv/openmv/pull/2603#discussion_r2139407064. It seems GC can't handle 32MBs blocks, probably too many blocks that overflow some undersized type or something like that, I'd ask Damien about it. I think it's a bad idea to have a block this large, it will have too many GC blocks and scanning it will probably be slow. If I increase words per block, it will be wasteful for the smaller, 1MB SRAM block. The GC doesn't support specifying this per block or at runtime.

Superseded by https://github.com/openmv/openmv/pull/2719

PR #2720, opened 2025-06-21
drivers/sensors/lepton: Update lepton driver with asynchronous fb.

kwagyeman:
This PR updates the FLIR Lepton driver to use multi-frame buffer architecture. This allows VOSPI to run in the background now, accumulating frames without blocking the main thread.

Now with the new multi-CSI architecture, `sensor.get_frame_available()` can be used to poll if a frame is available after the first snapshot starts the driver off, along with `sensor.get_framebuffer()`  to grab the thermal image in a non-blocking way.

Improvements to baseline:
* Transpose is now handled in addition to hmirror and vflip allowing for full image rotation.
* Legacy nearest neighbor code has been removed and replaced with bilinear image scaling which can leverage the GPU.
* Now that vospi writes directly to the frame buffers the legacy single vospi buffer can be free'd.

TODO:
* Need to do thorough testing on all ports using the Lepton (including N6).
* Update for use with multi-sensor PR for the N6.

### DIFF
--- a/common/vospi.c
+++ b/common/vospi.c
@@ -60,7 +60,7 @@ typedef enum {
 typedef struct _vospi_state {
     int pid;
     int sid;
-    uint16_t *framebuffer;
+    framebuffer_t *fb;
     bool lepton_3;
     omv_spi_t spi_bus;
     volatile uint32_t flags;
@@ -143,33 +143,39 @@ void vospi_callback(omv_spi_t *spi, void *userdata, void *buf) {
         return;
     }
 
-    memcpy(vospi.framebuffer
-           + (vospi.pid * VOSPI_PID_SIZE_PIXELS)
-           + (vospi.sid * VOSPI_SID_SIZE_PIXELS),
-           base + VOSPI_HEADER_WORDS, VOSPI_PID_SIZE_PIXELS * sizeof(uint16_t));
+    vbuffer_t *buffer = framebuffer_get_tail(vospi.fb, FB_PEEK);
 
-    vospi.pid += 1;
-    if (vospi.pid == VOSPI_PIDS_PER_SID) {
-        vospi.pid = 0;
+    if (buffer) {
+        memcpy(((uint16_t *) buffer->data)
+               + (vospi.pid * VOSPI_PID_SIZE_PIXELS)
+               + (vospi.sid * VOSPI_SID_SIZE_PIXELS),
+               base + VOSPI_HEADER_WORDS, VOSPI_PID_SIZE_PIXELS * sizeof(uint16_t));
 
-        // For the FLIR Lepton 3 we have to receive all the pids in all the segments.
-        if (vospi.lepton_3) {
-            vospi.sid += 1;
-            if (vospi.sid == VOSPI_SIDS_PER_FRAME) {
-                vospi.sid = 0;
-                vospi.flags &= ~VOSPI_FLAGS_CAPTURE;
+        vospi.pid += 1;
+        if (vospi.pid == VOSPI_PIDS_PER_SID) {
+            vospi.pid = 0;
+
+            // For the FLIR Lepton 3 we have to receive all the pids in all the segments.
+            if (vospi.lepton_3) {
+                vospi.sid += 1;
+                if (vospi.sid == VOSPI_SIDS_PER_FRAME) {
+                    vospi.sid = 0;
+                    framebuffer_get_tail(vospi.fb, FB_NO_FLAGS);
+                }
+                // For the FLIR Lepton 1/2 we just have to receive all the pids.
+            } else {
+                framebuffer_get_tail(vospi.fb, FB_NO_FLAGS);
             }
-            // For the FLIR Lepton 1/2 we just have to receive all the pids.
-        } else {
-            vospi.flags &= ~VOSPI_FLAGS_CAPTURE;
         }
+    } else {
+        vospi.flags &= ~VOSPI_FLAGS_CAPTURE;
     }
 }
 
-int vospi_init(uint32_t n_packets, void *buffer) {
+int vospi_init(uint32_t n_packets, framebuffer_t *fb) {
     memset(&vospi, 0, sizeof(vospi_state_t));
     vospi.lepton_3 = n_packets > VOSPI_PIDS_PER_SID;
-    vospi.framebuffer = buffer;
+    vospi.fb = fb;
     // resync on first snapshot.
     vospi.flags = VOSPI_FLAGS_RESYNC;
 
@@ -189,9 +195,22 @@ int vospi_init(uint32_t n_packets, void *buffer) {
     return 0;
 }
 
+int vospi_deinit() {
+    omv_spi_transfer_abort(&vospi.spi_bus);
+    omv_spi_deinit(&vospi.spi_bus);
+    memset(&vospi, 0, sizeof(vospi_state_t));
+    return 0;
+}
+
 int vospi_snapshot(uint32_t timeout_ms) {
-    // Restart counters to capture a new frame.
-    vospi.flags |= VOSPI_FLAGS_CAPTURE;
+    framebuffer_free_current_buffer(vospi.fb);
+
+    if (!(vospi.flags & VOSPI_FLAGS_CAPTURE)) {
+        framebuffer_setup_buffers(vospi.fb);
+
+        // Restart counters to capture a new frame.
+        vospi.flags |= VOSPI_FLAGS_CAPTURE;
+    }
 
     // Snapshot start tick
     mp_uint_t tick_start = mp_hal_ticks_ms();
@@ -211,7 +230,7 @@ int vospi_snapshot(uint32_t timeout_ms) {
         }
 
         MICROPY_EVENT_POLL_HOOK
-    } while (vospi.flags & VOSPI_FLAGS_CAPTURE);
+    } while (!framebuffer_get_head(vospi.fb, FB_NO_FLAGS));
 
     return 0;
 }
--- a/drivers/sensors/lepton.c
+++ b/drivers/sensors/lepton.c
@@ -202,7 +202,7 @@ static int ioctl(omv_csi_t *csi, int request, va_list ap) {
         }
         case OMV_CSI_IOCTL_LEPTON_GET_RESOLUTION: {
             int *resolution = va_arg(ap, int *);
-            *resolution = 14;
+            *resolution =  lepton.radiometry ? 16 : 14;
             break;
         }
         case OMV_CSI_IOCTL_LEPTON_RUN_COMMAND: {
--- a/drivers/sensors/lepton.c
+++ b/drivers/sensors/lepton.c
@@ -69,7 +69,6 @@ typedef struct lepton_state {
     LEP_CAMERA_PORT_DESC_T port;
 } lepton_state_t;
 
-extern uint16_t _vospi_buf[];
 static lepton_state_t lepton;
 
 static int lepton_reset(omv_csi_t *csi, bool measurement_mode, bool high_temp_mode);
@@ -359,8 +358,6 @@ static int lepton_reset(omv_csi_t *csi, bool measurement_mode, bool high_temp_mo
 }
 
 static int reset(omv_csi_t *csi) {
-    static bool vospi_initialized = false;
-
     memset(&lepton, 0, sizeof(lepton_state_t));
     lepton.min_temp = LEPTON_MIN_TEMP_DEFAULT;
     lepton.max_temp = LEPTON_MAX_TEMP_DEFAULT;
@@ -369,11 +366,9 @@ static int reset(omv_csi_t *csi) {
         return OMV_CSI_ERROR_CTL_FAILED;
     }
 
-    if (vospi_initialized == false) {
-        if (vospi_init(lepton.v_res, _vospi_buf) != 0) {
-            return OMV_CSI_ERROR_CTL_FAILED;
-        }
-        vospi_initialized = true;
+    vospi_deinit();
+    if (vospi_init(lepton.v_res, csi->fb) != 0) {
+        return OMV_CSI_ERROR_CTL_FAILED;
     }
 
     return 0;
@@ -382,10 +377,6 @@ static int reset(omv_csi_t *csi) {
 static int snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
     framebuffer_t *fb = csi->fb;
     framebuffer_update_jpeg_buffer(fb);
-
-    if (fb->n_buffers != 1) {
-        framebuffer_set_buffers(fb, 1);
-    }
 
     if (csi->pixformat == PIXFORMAT_INVALID) {
         return OMV_CSI_ERROR_INVALID_PIXFORMAT;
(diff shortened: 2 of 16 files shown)


iabdalkader:
Should py_fir_lepton be refactored to use vospi now? I imagined there would be a lot of code in common.

kwagyeman:
I was just trying to do the minimum for now to get the PR done quickly. I think maybe that should wait until after your PR as we might want to re-use the frame buffer architecture for it as it has it's own custom triple buffer code in py_fir_lepton.

Alternatively, once the dual sensor driver feature is out there, maybe we just deprecate/delete the whole thing as it's redundant (not the other thermal sensors, but, just the lepton one since it can be made an auxiliary sensor now).

ports/stm32/omv_gpu.c
```diff
@@ -117,15 +117,23 @@ int omv_gpu_draw_image(image_t *src_img,
     nema_blit_subrect_fit(dst_rect->x, dst_rect->y, dst_rect->w, dst_rect->h,
                           src_rect->x, src_rect->y, src_rect->w, src_rect->h);
 
+    // Ensures any cached writes to dst are flushed.
+    SCB_CleanInvalidateDCache_by_Addr(dst_img->data, image_size(dst_img));
     // Flush source image
     SCB_CleanDCache_by_Addr(src_img->data, image_size(src_img));
+    // Ensure the GPU cache is clean before starting the GPU operation.
+    // Will start invalidating the GPU cache if not already invalidated.
+    HAL_ICACHE_Invalidate();
```
iabdalkader: Should this be `HAL_ICACHE_WaitForInvalidateComplete();` ?

kwagyeman: It calls that internally, but, also kicks off the invalidate first if not running.

iabdalkader: This would invalidate it twice after a draw. I think we should call `HAL_ICACHE_WaitForInvalidateComplete`

iabdalkader:
> Alternatively, once the dual sensor driver feature is out there, maybe we just deprecate/delete the whole thing as it's redundant (not the other thermal sensors, but, just the lepton one since it can be made an auxiliary sensor now).

Of course I prefer to remove that code if possible, but will it work? It uses a different bus.

ports/stm32/omv_gpu.c
kwagyeman: Works, updated

kwagyeman:
It's only really needed for the Pure Thermal, but, that's perfect for dual sensors anyway and is fixed. Attaching a FLIR Lepton via the pin headers can then be dropped as a feature. Not sure if anyone is using this anyway. The dual sensor module removes any real need to attach a FLIR Lepton the hard way via the pin headers anyway.

iabdalkader:
> It's only really needed for the Pure Thermal, but, that's perfect for dual sensors anyway and is fixed. Attaching a FLIR Lepton via the pin headers can then be dropped as a feature. Not sure if anyone is using this anyway. The dual sensor module removes any real need to attach a FLIR Lepton the hard way via the pin headers anyway.

Good point. Even if they do, on a custom board, they could just redefine the pins. Okay let's drop it. We could remove it now and then after dual_csi it will just work again.

ports/stm32/omv_spi.c
```diff
@@ -95,7 +95,7 @@ DEFINE_SPI_INSTANCE(6)
 
 static omv_spi_t *omv_spi_descr_all[6] = { NULL };
 #if defined(STM32N6)
-static DMA_NodeTypeDef OMV_ATTR_SECTION(OMV_ATTR_ALIGNED(omv_dma_nodes_all[6 * 2], 32), ".dma_buffer");
+static DMA_NodeTypeDef OMV_ATTR_SECTION(OMV_ATTR_ALIGNED(omv_dma_nodes_all[6 * 2], 32), ".d2_dma_buffer");
```
iabdalkader: Should this be `OMV_VOSPI_DMA_BUFFER` ?

kwagyeman:
Okay, will do tomorrow once back at work. New code seems to be quite stable with the FLIR Lepton now. I'll copy this PR and integrate with dual sensor support to make sure that all works too.

Note, I'm noticing the linkup speed with the new lepton adapter board to be pretty much instant after running the script. Linkup with the old original flir lepton board is less reliable. Sometimes it works on the first try, other times it needs 2-3 resets taking 3-6 seconds.

iabdalkader:
> https://github.com/openmv/openmv/pull/2720/commits/ac1598f03bb2b884392141555bb0d9c68fa8aa9f

Not a bad idea, but you could also just switch to HPDMA? It has a 64-bits port to AXI where the old DMA memory is. You will need to use  `DMA_HPDMA_LINEAR_NODE` and may need to swap PORT0/1. Although I'm not sure if it's better to do this with CSI also running at the same time.

EDIT: For node type, it should be detected somehow not hardcoded. The ports should also be assigned based on the direction.

ports/stm32/omv_spi.c
kwagyeman: It could be. It's shared though for all SPI DMA. Suggestions?

iabdalkader: What's shared, I don't follow. You defined `OMV_VOSPI_DMA_BUFFER` for N6 but then hardcoded it.

kwagyeman:
That's a good idea. Would mean faster access to the VOSPI buffer by the processor. I can explore tomorrow.

ports/stm32/omv_spi.c
kwagyeman: If HPDMA works this doesn't need to be changed.

ports/stm32/omv_spi.c
```diff
@@ -330,8 +330,8 @@ static int omv_spi_dma_init(omv_spi_t *spi, uint32_t direction, omv_spi_config_t
     dma_init->Direction = direction;
     dma_init->SrcInc = (direction == DMA_PERIPH_TO_MEMORY) ? DMA_SINC_FIXED : DMA_SINC_INCREMENTED;
     dma_init->DestInc = (direction == DMA_MEMORY_TO_PERIPH) ? DMA_DINC_FIXED : DMA_DINC_INCREMENTED;
-    dma_init->SrcDataWidth = (config->datasize == 8) ?  DMA_SRC_DATAWIDTH_BYTE : DMA_SRC_DATAWIDTH_WORD;
-    dma_init->DestDataWidth = (config->datasize == 8) ?  DMA_DEST_DATAWIDTH_BYTE : DMA_DEST_DATAWIDTH_WORD;
+    dma_init->SrcDataWidth = DMA_SRC_DATAWIDTH_WORD;
+    dma_init->DestDataWidth = DMA_DEST_DATAWIDTH_WORD;
```
iabdalkader, line 334: Should we do this? What if we're transferring an uneven amount of bytes or shorts?

kwagyeman: I've had it set for the H7 to just work on 4-byte chunks. It means the SPI driver in DMA mode needs to be a multiple of 4-bytes. Haven't run into any problems with this limitation. I'd prefer to minimize the transaction count as the buses are all AHB which get clogged pretty easily once you start to do multiple things at once.

iabdalkader: >  I'd prefer to minimize the transaction count as the buses are all AHB which get clogged pretty easily once you start to do multiple things at once.

You should increase the burst size, especially if using the AXI.

drivers/sensors/lepton.c
```diff
@@ -369,11 +367,8 @@ static int reset(omv_csi_t *csi) {
         return OMV_CSI_ERROR_CTL_FAILED;
     }
 
-    if (vospi_initialized == false) {
-        if (vospi_init(lepton.v_res, _vospi_buf) != 0) {
-            return OMV_CSI_ERROR_CTL_FAILED;
-        }
-        vospi_initialized = true;
```
iabdalkader, line 376: Why initialize vospi on every reset? Also, we need to check fb!=NULL because it's only allocated when `csi` object is created.

iabdalkader:
I don't see any performance gain with this, with the N6 I get 4FPS.. Anyway, this is needed to make it work with dual_csi. The delay in reset is needed because it gets called from omv_csi_reset, otherwise I2C gets stuck, and FB is only set after CSI object is created:

```diff
diff --git a/drivers/sensors/lepton.c b/drivers/sensors/lepton.c
index 7fae58b8a..1e9fef18c 100644
--- a/drivers/sensors/lepton.c
+++ b/drivers/sensors/lepton.c
@@ -366,6 +366,9 @@ static int match(omv_csi_t *csi, size_t id) {
 }
 
 static int reset(omv_csi_t *csi) {
+    // Extra delay after power-on
+    mp_hal_delay_ms(1000);
+
     vospi_deinit();
     memset(&lepton, 0, sizeof(lepton_state_t));
     lepton.min_temp = LEPTON_MIN_TEMP_DEFAULT;
@@ -375,10 +378,9 @@ static int reset(omv_csi_t *csi) {
         return OMV_CSI_ERROR_CTL_FAILED;
     }
 
-    if (vospi_init(lepton.v_res, csi->fb) != 0) {
+    if (csi->fb && vospi_init(lepton.v_res, csi->fb) != 0) {
         return OMV_CSI_ERROR_CTL_FAILED;
     }
-
     return 0;
 }
 
@@ -512,9 +514,6 @@ int lepton_init(omv_csi_t *csi) {
     csi->frame_sync = 0;
     csi->mono_bpp = 1;
 
-    // Extra delay after power-on
-    mp_hal_delay_ms(1000);
-
     if (reset(csi) != 0) {
         return -1;
     }
```

For HPDMA, it doesn't work without the following code. I'm not sure what it does exactly or how it works, please search the examples for `HAL_RIF_RISAF_ConfigBaseRegion`.

```C
    RISAF_BaseRegionConfig_t risaf_conf;
    __HAL_RCC_RISAF_CLK_ENABLE();

    /* set up base region configuration for AXISRAM1 and 2 */
    risaf_conf.StartAddress = 0;

    risaf_conf.Filtering = RISAF_FILTER_ENABLE;
    risaf_conf.PrivWhitelist = RIF_CID_MASK;
    risaf_conf.ReadWhitelist = RIF_CID_MASK;
    risaf_conf.WriteWhitelist = RIF_CID_MASK;

    /* AXISRAM1 is secure */
    risaf_conf.Secure = RIF_ATTRIBUTE_SEC;
    risaf_conf.EndAddress = 0x000FFFFF;
    HAL_RIF_RISAF_ConfigBaseRegion(RISAF2, RISAF_REGION_1, &risaf_conf);

    /* AXISRAM2 is non-secure */
    risaf_conf.Secure = RIF_ATTRIBUTE_SEC;
    risaf_conf.EndAddress = 0x000FFFFF;
    HAL_RIF_RISAF_ConfigBaseRegion(RISAF3, RISAF_REGION_1, &risaf_conf);
```

And: 

```C
    if (direction == DMA_MEMORY_TO_PERIPH) {
        dma_init->TransferAllocatedPort = DMA_SRC_ALLOCATED_PORT0 | DMA_DEST_ALLOCATED_PORT1;
    } else {
        dma_init->TransferAllocatedPort = DMA_SRC_ALLOCATED_PORT1 | DMA_DEST_ALLOCATED_PORT0;
    }

...
.NodeType = DMA_HPDMA_LINEAR_NODE,
```

I don't see any performance difference with HPDMA, so it must be something else.

drivers/sensors/lepton.c
kwagyeman: Done

ports/stm32/omv_spi.c
kwagyeman: Given HPDMA isn't really needed for this, I left it with GPDMA to avoid any more changes being required.

kwagyeman:
PR Updated. Since your testing with HPDMA shows it doesn't really matter I'm not sure it if makes sense to put the effort into switching over to it. Might be better to keep it free for use with the parallel interface for a second image sensor later.

As for deleting py_fir_lepton. I'd prefer to do that after dual csi is merged so that I can do it in one go and update the Pure Thermal examples at the same time. I want to avoid leaving the board in a broken state: https://github.com/openmv/openmv/issues/2724

Otherwise, let me know what other changes you want in this PR. Once I'm done with all the changes, I'll retest functionality on all ports to catch any last issues. It's working right now though with all boards.

ports/mimxrt/main.c
```diff
@@ -222,6 +223,9 @@ int main(void) {
 
 soft_reset_exit:
     mp_printf(MP_PYTHON_PRINTER, "MPY: soft reboot\n");
+    #if OMV_ENABLE_VOSPI || OMV_LEPTON_ENABLE
+    vospi_deinit();
+    #endif
```
iabdalkader, line 228: We could add a new function to be called on softreset, what do you think? Maybe future aux sensors will also need to deinit something. I've already added a finalizer to CSI-NG that resets the FB and calls abort.

EDIT: Actually, maybe we should just call `omv_csi_abort()`  and implement `lepton->abort()` which in turn calls `vospi_abort()` -> `omv_spi_transfer_abort()` ? Note that `omv_csi_init0` already aborts the transfer, but here (right before the softreset) is probably a better place to abort the transfer, right?

common/vospi.c
```diff
@@ -183,15 +189,24 @@ int vospi_init(uint32_t n_packets, void *buffer) {
...
-    if (omv_spi_init(&vospi.spi_bus, &spi_config) != 0) {
-        return -1;
-    }
-    return 0;
+    return omv_spi_init(&vospi.spi_bus, &spi_config);
+}
+
+int vospi_deinit() {
+    int ret = omv_spi_deinit(&vospi.spi_bus);
+    memset(&vospi, 0, sizeof(vospi_state_t));
```
iabdalkader:
```suggestion
```
The state is already reset in `vospi_init`. Any reason to reset it again?

iabdalkader:
> Since your testing with HPDMA shows it doesn't really matter I'm not sure it if makes sense to put the effort into switching over to it

It's hard to really see if it matters give how slow the FLIR is anyway, but if it saves bandwidth then we should probably look into it at some point in the future.  One more reason to look into HPDMA is that MP uses GPDMA for I2C/SPI right now, but with no IRQs. If it starts using IRQs too there will be conflicts. We could use HPDMA exclusively, and leave GPDMA for MP to avoid any conflicts.

ports/mimxrt/main.c
kwagyeman: Yes, omv_csi_abort() is the way to go.

common/vospi.c
kwagyeman: Not really, just wanted to leave the state zeroed can remove.

kwagyeman: Removed

kwagyeman:
@iabdalkader - Tested now. All good.

iabdalkader:
While we're at it, I've added flags/errors to support non-blocking snapshot, but I need `vospi_snapshot` to kick start the transfer and return if there's no ready frame and `timeout == 0` (or via any other means, like add flags if you like). The following seems to work, but you may want to implement it differently:
```diff
diff --git a/common/vospi.c b/common/vospi.c
index f0a8dc6ed..f59741588 100644
--- a/common/vospi.c
+++ b/common/vospi.c
@@ -196,6 +196,10 @@ int vospi_deinit() {
     return omv_spi_deinit(&vospi.spi_bus);
 }
 
+bool vospi_active(void) {
+    return (vospi.flags & VOSPI_FLAGS_CAPTURE);
+}
+
 int vospi_snapshot(uint32_t timeout_ms) {
     framebuffer_free_current_buffer(vospi.fb);
 
@@ -215,6 +219,10 @@ int vospi_snapshot(uint32_t timeout_ms) {
             vospi_resync();
         }
 
+        if (timeout_ms == 0) {
+            return -1;
+        }
+
         if ((mp_hal_ticks_ms() - tick_start) > timeout_ms) {
             omv_spi_transfer_abort(&vospi.spi_bus);
             vospi.pid = 0;
```

With that I tested nonblocking capture and it's working fine. It looks something like this:
```Python
img1 = None

while True:
    clock.tick()
    img0 = csi0.snapshot()
    fimg = csi1.snapshot(update=False, blocking=False)
    if fimg is not None:
        img1 = fimg
    if img1 is not None:
        img0.draw_image(img1, 0, 0, x_scale=0.5, y_scale=0.5)
```

PR #2735, opened 2025-06-26
misc: Refactor CSI initialization code.

iabdalkader:


iabdalkader:
Note that the last commit only renames files, but git couldn't tell the difference.

common/omv_csi.c
```diff
@@ -459,222 +437,58 @@ int omv_csi_probe(omv_i2c_t *i2c) {
...
-            #if (OMV_SOFTCSI_ENABLE == 1)
-            case SOFTCSI_ID:
-                init_ret = softcsi_init(csi);
-                break;
-            #endif // (OMV_SOFTCSI_ENABLE == 1)
+        if (init_fun ==  NULL) {
+            return OMV_CSI_ERROR_ISC_UNSUPPORTED;
+        } else if (init_fun(csi) != 0) {
+            return OMV_CSI_ERROR_ISC_INIT_FAILED;
+        }
```
kwagyeman, line 464: Move this after setting the correct clock frequency to match the original behavior.

drivers/sensors/sensor_config.h
```diff
@@ -0,0 +1,217 @@
...
+    #endif
+
+    #if OMV_LEPTON_ENABLE
+    { LEPTON_ID,     OMV_LEPTON_CLK_FREQ, lepton_init },
+    { LEPTON_1_5,    OMV_LEPTON_CLK_FREQ, lepton_init },
+    { LEPTON_1_6,    OMV_LEPTON_CLK_FREQ, lepton_init },
+    { LEPTON_2_0,    OMV_LEPTON_CLK_FREQ, lepton_init },
+    { LEPTON_2_5,    OMV_LEPTON_CLK_FREQ, lepton_init },
+    { LEPTON_3_0,    OMV_LEPTON_CLK_FREQ, lepton_init },
+    { LEPTON_3_5,    OMV_LEPTON_CLK_FREQ, lepton_init },
```
kwagyeman: Not strictly required as LEPTON_ID becomes LEPTON_1_5 and etc. during init.

drivers/sensors/sensor_config.h
```diff
@@ -0,0 +1,217 @@
...
+    #endif
+
+    #if OMV_MT9M114_ENABLE
+    { MT9M114_ID, OMV_MT9M114_CLK_FREQ, mt9m114_init },
+    #endif
+
+    #if OMV_BOSON_ENABLE
+    { BOSON_ID,      OMV_BOSON_CLK_FREQ, boson_init },
+    { BOSON_320_ID,  OMV_BOSON_CLK_FREQ, boson_init },
+    { BOSON_640_ID,  OMV_BOSON_CLK_FREQ, boson_init },
```
kwagyeman: Also not strictly needed.

common/omv_csi.c
```diff
@@ -459,222 +437,58 @@ int omv_csi_probe(omv_i2c_t *i2c) {
...
-            case OV9650_ID:
-                init_ret = ov9650_init(csi);
-                break;
-            #endif // (OMV_OV9650_ENABLE == 1)
-
-            #if (OMV_MT9V0XX_ENABLE == 1)
-            case MT9V0X2_ID_V_1:
-            case MT9V0X2_ID_V_2:
-                // Force old versions to the newest.
-                csi->chip_id = MT9V0X2_ID;
```
kwagyeman, line 531: This is another special case that's missing. It could be moved into mt9v0xx_init() as it's just a chip ID change.

drivers/sensors/sensor_config.h
```diff
@@ -0,0 +1,217 @@
...
+    { OV9650_ID, OMV_OV9650_CLK_FREQ, ov9650_init },
+    #endif
+
+    #if OMV_MT9V0XX_ENABLE
+    { MT9V0X2_ID_V_1, OMV_MT9V0XX_CLK_FREQ, mt9v0xx_init },
+    { MT9V0X2_ID_V_2, OMV_MT9V0XX_CLK_FREQ, mt9v0xx_init },
+    { MT9V0X2_ID,     OMV_MT9V0XX_CLK_FREQ, mt9v0xx_init },
+    { MT9V0X4_ID,     OMV_MT9V0XX_CLK_FREQ, mt9v0xx_init },
+    { MT9V0X4_C_ID,   OMV_MT9V0XX_CLK_FREQ, mt9v0xx_init },
+    { MT9V0X2_C_ID,   OMV_MT9V0XX_CLK_FREQ, mt9v0xx_init },
```
kwagyeman: These don't appear until after mt9v0xx_init() too.

drivers/sensors/sensor_config.h
iabdalkader: Yes it is, this is looked up before init, it has the init function right there.

kwagyeman: Yes, but init should be called after setting the correct clock frequency, since some drivers depend on the clock frequency being set before initialization.

common/omv_csi.c
iabdalkader: No, we need aux set in init.

drivers/sensors/sensor_config.h
iabdalkader: Oh you mean we override to set those?

kwagyeman: Current working behavior is we lookup the sensor id, set the right clock freq based on that, then call init, and then init may change the sensor id.

iabdalkader: Yes I know that, it needs to change. Unless you have a specific sensor failing with this behavior, it will be changed.

kwagyeman: Let me check...

common/omv_csi.c
kwagyeman: Okay, I checked everything and only lepton could have issues. Let me test that sensor with this PR. All other sensors should be okay with it.

common/omv_csi.c
kwagyeman: I'll update the IDE to handle this.

common/omv_csi.c
iabdalkader: `sensor_init` should do the bare minimum: set ops, flags etc.. You do a lot more in some init functions, I was thinking we could rename it to `init_ops` and have another init or post_init, if needed. Alternatively, I might be able to store a "default_aux" flag in the config table this way I'll know whether to set the clock or not, then the dynamic aux flag can be changed after.

iabdalkader: Note that if the sensor is detected, the clock should be good enough to read/write regs.

kwagyeman: For the FLIR Lepton by itself, this does break it. Its clock needs to be 24-25MHz. 12 MHz doesn't work. I get an init failure.

kwagyeman: For the dual sensor breakout this doesn't matter since it has it's own clock in that case.

kwagyeman: This commit I think solves the problem: https://github.com/openmv/openmv/pull/2737/commits/d5f1701cd025a25a6ddc472cc7a6853044f8a01a

Should add back the 1second delay to it though.

kwagyeman: Can confirm the commit fixes it with the 1 second delay kept.

iabdalkader: So it's unrelated to where the clock init is called?

kwagyeman: Yeah, it's all good otherwise for each sensor. Just need that commit.

drivers/sensors/sensor_config.h
iabdalkader: Why did you resolve this?

iabdalkader:
Okay updated the table.

kwagyeman:
Before merging this, I want to test with all sensors.

iabdalkader:
It's best to do that test after all these PRs are merged.

kwagyeman:
Okay, go ahead then.

PR #2736, opened 2025-06-26
common/csi: Decouple the clock from main CSI state.

iabdalkader:
This patch decouples the clock from the main CSI state, further separating the CSI instance from the port's CSI driver. Additionally, it allows CSIs to use different clocks, in theory, though they must first be detected somehow in order to switch clocks.

As a side effect of sharing the clock, set_frequency will be called more frequently (the default call plus once per CSI). However, the frequency is now checked, and the clock is only reconfigured if the new frequency exceeds a configurable tolerance.

Finally, get_clk_frequency now accepts a boolean to return either the exact clock frequency, for sensors that require it, or the nominal frequency.

Note that sensor drivers should be checked to see which ones require the exact clock and which require the nominal clock. I best guessed them for now.

Tested imxrt, alif, stm32.

### DIFF
--- a/common/omv_csi.c
+++ b/common/omv_csi.c
@@ -61,6 +61,10 @@
 #define OMV_CSI_I2C_REINIT  (1)
 #endif
 
+#ifndef OMV_CSI_CLK_TOLERANCE
+#define OMV_CSI_CLK_TOLERANCE   (500000)
+#endif
+
 #ifndef __weak
 #define __weak    __attribute__((weak))
 #endif
@@ -120,6 +124,7 @@ uint16_t resolution[][2] = {
 };
 
 static omv_i2c_t csi_i2c;
+static omv_clk_t csi_clk;
 omv_csi_t csi_all[OMV_CSI_MAX_DEVICES] = {0};
 
 __weak void omv_csi_init0() {
@@ -178,14 +183,15 @@ __weak int omv_csi_init() {
 
         memset(csi, 0, sizeof(omv_csi_t));
         csi->i2c = &csi_i2c;
+        csi->clk = &csi_clk;
         csi->fb = framebuffer_get(-1);
         csi->color_palette = rainbow_table;
         omv_csi_ops_init(csi);
-    }
 
-    // Configure the csi external clock (XCLK).
-    if (omv_csi_set_clk_frequency(OMV_CSI_CLK_FREQUENCY) != 0) {
-        return OMV_CSI_ERROR_TIM_INIT_FAILED;
+        // Configure the csi external clock (XCLK).
+        if (omv_csi_set_clk_frequency(csi, OMV_CSI_CLK_FREQUENCY) != 0) {
+            return OMV_CSI_ERROR_TIM_INIT_FAILED;
+        }
     }
 
     // Detect and initialize sensor(s).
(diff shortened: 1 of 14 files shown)


common/omv_csi.c
```diff
@@ -540,14 +547,16 @@ int omv_csi_probe(omv_i2c_t *i2c) {
...
-        // Allow reconfiguring (or disabling) the external clock
-        // if just one sensor is detected, or for main sensors.
-        if (dev_count == 1 || !csi->auxiliary) {
-            omv_csi_set_clk_frequency(csi->clk_hz);
+        // Sensors can request disabling the clock (clk_hz == 0). This
+        // is allowed only if the clock is not shared (dev_count == 1).
+        // If the frequency is valid, the clock can be reconfigured as
+        // all sensors should share a common frequency.
+        if (clk_hz != 0 || dev_count == 1) {
+            omv_csi_set_clk_frequency(csi, clk_hz);
```
kwagyeman: Reviewed the PR. Just the change with omv_csi_set_clk_frequency() here and the removal below stand out.

The FLIR Lepton and PAG7936 are both 24 MHz... and then OV5640 + Lepton are 24 MHz both or 24 and 25 MHz if rev y. 

So, the above kinda works, but, the aux sensor could override the main sensor clock in this case for like the rev y? Why not keep the previous behavior?

iabdalkader: The assumption is that if you use multiple sensors, they must all either share a common clock frequency, or use separate clock sources (internal clock, onboard OSC, or maybe a different timer). For an OV5640/Lepton, you could either use an onboard OSC or drop the REV-Y thing.

iabdalkader: I see what you mean, in case Lepton gets detected last it will revert the frequency to 24MHz... Let me think about it, though I'd prefer to drop the rev-y hack if it's not important.

kwagyeman: Yeah... I don't like the rev y thing too, but, there are lots of those chips out there that only run at 400 MHz versus 480 MHz. They can't do 24 MHz exactly, only 25 MHz.

iabdalkader: The N6 can't do an exact 24MHz either, doesn't seem to be a problem. Note please review the sensors drivers and see which need exact/nominal frequency.

kwagyeman: Mmm, for the 400 MHz H7 versions I guess you'd hit 23.529 MHz. Close enough to 24 MHz.

Okay, yeah, I think then it's fine to drop the rev y. I was probably trying to avoid a non-perfect freq before. I don't see any issues with the sensor drivers then.

There would only be an issue if the clock was faster than the requested speed, but, the set_clock_freq won't do that.

iabdalkader: Note to restore this:
```C
// If more than one aux sensor was detected, the clock
// hasn't been changed, so reconfigure using this freq.
if (dev_count > 1) {
    omv_csi_set_clk_frequency(csi->clk_hz);
}
```
I'd have to revert `csi->clk_hz` and having two clk_hz/freq just seemed like it would be very confusing. Alternatively, I could perform the lookup twice.  I'd like to remove rev-y special case regardless.

kwagyeman: Approved

kwagyeman (approved):

common/omv_csi.c
iabdalkader: I've thought about this some more.. And while the old logic handles the following case:
- An aux sensor that uses say 12MHz clock standalone and an independent clock in dual config. 

It fails to deal with this case:
- Multiple *main* sensors are detected (e.g. PAG + OV), with PAG being detected last. In this case, the old logic would set the clock twice, since they're both main, and PAG will revert OV's clock back to 24MHz.

I decided to revert the old logic, as it handles the first case which is more common. However, the following rule must be set to handle the second case: Main sensors on multi-sensors modules must share a common frequency. More generally: sensors must either share a common clock frequency, or use independent clocks + aux flag set. Therefore, there can't be any rev-y hack, as the OV5640 could be used with another main sensor and in this case it *must share the frequency*. So please remove the OV rev-y stuff in a follow-up PR.

kwagyeman: Will remove it. I think I have one of these rev-y boards around too. So, I can test.

PR #2739, opened 2025-06-27
common: Remove xalloc.

iabdalkader:
Originally meant to abstract gc_collect but we could just use m_alloc and friends. Also was meant to provide functions like alloc0, alloc_maybe etc.. which are all available in MP anyway.


kwagyeman:
Don't see any issues. Seems like 1:1 replacement.

PR #2742, opened 2025-06-29
drivers/sensors: Add lepton abort support.

kwagyeman:
Changing the resolution/pixformat between snapshots is now safe with the FLIR Lepton.

### DIFF
--- a/common/vospi.c
+++ b/common/vospi.c
@@ -196,6 +196,11 @@ int vospi_deinit() {
     return omv_spi_deinit(&vospi.spi_bus);
 }
 
+int vospi_abort(void) {
+    vospi.flags &= ~VOSPI_FLAGS_CAPTURE;
+    return 0;
+}
+
 bool vospi_active(void) {
     return vospi.flags & VOSPI_FLAGS_CAPTURE;
 }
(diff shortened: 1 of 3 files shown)


common/vospi.c
```diff
@@ -196,6 +196,11 @@ int vospi_deinit() {
     return omv_spi_deinit(&vospi.spi_bus);
 }
 
+int vospi_abort(void) {
+    vospi.flags &= ~VOSPI_FLAGS_CAPTURE;
```
iabdalkader, line 200: Shouldn't the SPI transfer be aborted here? This is not enough to stop SPI transfers.

kwagyeman: Yes, we could do that, but, then VOSPI has to reinit which can take a quite a while. Disabling the capture flag is enough to allow for frame buffer geometry changes and etc. Then vospi deinit() is still called on soft reboot. We do need to deinit though on CSI object destruction... whenever the framebuffer is deleted.

Will add.

iabdalkader: No, vospi would just need to start the transfer again, which it does already. As for soft-reboot, I'm changing that abort will be called on soft-reboot.

iabdalkader: Let me resend this along with other fixes.

PR #2743, opened 2025-06-29
common/csi: Add csi shutdown op.

iabdalkader:


common/vospi.c
```diff
@@ -196,6 +196,11 @@ int vospi_deinit() {
     return omv_spi_deinit(&vospi.spi_bus);
 }
 
+int vospi_abort(void) {
+    vospi.flags &= ~VOSPI_FLAGS_CAPTURE;
+    return omv_spi_transfer_abort(&vospi.spi_bus);
```
kwagyeman:
```suggestion
    vospi.flags &= ~VOSPI_FLAGS_CAPTURE;
    int ret = omv_spi_transfer_abort(&vospi.spi_bus);
    vospi.flags |= VOSPI_FLAGS_RESYNC;
    vospi.pid = 0;
    vospi.sid = 0;
    return ret;
```

ports/stm32/omv_csi.c
```diff
@@ -192,6 +192,13 @@ static int stm_csi_config(omv_csi_t *csi, omv_csi_config_t config) {
         DCMI->CR &= ~(DCMI_CR_JPEG_Msk << DCMI_CR_JPEG_Pos);
         DCMI->CR |= (csi->pixformat == PIXFORMAT_JPEG) ? DCMI_JPEG_ENABLE : DCMI_JPEG_DISABLE;
         #else
+
+        // Reset pipe state to allow reconfiguring them. Note that
+        // abort() doesn't reset the state unless the pipe is active.
+        for (size_t i=0; i<DCMIPP_NUM_OF_PIPES; i++) {
+            csi->dcmi.PipeState[i] = HAL_DCMIPP_PIPE_STATE_RESET;
+        }
```
kwagyeman: This change seems to make set_pixformat() work only once per boot of the camera. Like, if I run a script, stop it, and then run another script then set_pixformat fails with an error until the system is power cycled.

common/vospi.c
kwagyeman: pid/sid need to be zeroed and rsync set to startup the vospi interface again if you do something like set_pixformat() -> snapshot() ->set_pixformat() -> snapshot().

kwagyeman:
Can confirm that the dual sensor works smoothly and the reset issue appears to be resolved.

Tested:

N6:
OV7725 -> Works
OV2640 -> Works
OV5640 -> Works
MT9M114 -> Works
MT9V034 -> Works
Lepton -> Works.
Boson -> Works
GENX320 -> Works
PAG7920 -> Works
PAG6100 -> Works
PS5520 -> Works
PAG7936 -> Works
PAG7936 + Lepton -> Works

Changing the pixformat on the fly is not working though. But, shutdown/reset fixes are good.

common/vospi.c
iabdalkader: Okay, I refactor all the code that does similar abort.

ports/stm32/omv_csi.c
iabdalkader: Fixed.. Before these changes the DCMI was initialized twice resetting all state, so any problem like that is likely related to HAL states not getting reset anymore. The HAL, being brain-damaged as it is, doesn't handle reconfiguring things like that. Also some functions expect state ready, or state_init, or okay with either.

iabdalkader:
I tested every possible combination (in both measurement and regular modes) with the changes in this PR:

Lepton 2.5:
* FIR standalone (works).
* FIR + Soft-CSI (works).
* FIR + RGB + Soft-CSI (works).

Lepton 3.5:
* FIR standalone -> Measurement mode broken, all red using 5->30.
* FIR + Soft-CSI  (works).
* FIR + RGB + Soft-CSI (works).

Not sure why 3.5 measurement mode doesn't work very well when put in the single adapter module. Could it be power issue? I use the same range with the dual module and it works fine.

common/vospi.c
```diff
@@ -136,10 +136,7 @@ void vospi_callback(omv_spi_t *spi, void *userdata, void *buf) {
         || (!vospi_check_crc(base))
     #endif
         || (vospi.lepton_3 && (pid == VOSPI_SPECIAL_PACKET) && (sid != vospi.sid))) {
-        vospi.pid = 0;
-        vospi.sid = 0;
-        omv_spi_transfer_abort(&vospi.spi_bus);
-        vospi.flags |= VOSPI_FLAGS_RESYNC;
+        vospi_abort();
```
kwagyeman, line 139: Don't clear the VOSPI_FLAGS_CAPTURE flag here (abort clears it). This breaks the lepton (by itself), working on all my boards.

kwagyeman:
I tested on all these platforms:

M4:
OV7725 -> Works
OV2640 -> Works

M7:
OV7725 -> Works

H7:
OV7725 -> Works
OV2640 -> Works
OV5640 -> Works
MT9M114 -> Works
MT9V034 -> Works
Lepton -> Works
Boson -> Not enabled
GENX320 -> Not enabled
PAG7920 -> Works
PAG6100 -> Works

H7 Plus:
OV7725 -> Works
OV2640 -> Works
OV5640 -> Works
MT9M114 -> Works
MT9V034 -> Works
Lepton -> Works
Boson -> Works
GENX320 -> Works
PAG7920 -> Works
PAG6100 -> Works

RT1062:
OV7725 -> Works
OV2640 -> Works
OV5640 -> Works
MT9M114 -> Works
MT9V034 -> Works
Lepton -> Works
Boson -> Works
GENX320 -> Works
PAG7920 -> Works
PAG6100 -> Works

N6:
OV7725 -> Works
OV2640 -> Works
OV5640 -> Works
MT9M114 -> Works
MT9V034 -> Works
Lepton -> Works.
Boson -> Works
GENX320 -> Works
PAG7920 -> Works
PAG6100 -> Works
PS5520 -> Works
PAG7936 -> Works
PAG7936 + Lepton -> Works

AE3:
PAG7936 -> Works.

ARDUINO NICLA:
GC2145 -> Works

ARDUINO Portenta:
HMB0360 -> Works

Arduino Giga:
HMB0360 -> Works

Everything works except for an issue with the capture flag being cleared during automatic resync in the vospi driver that needs to be fixed. What that fix it's 100%.

common/vospi.c
iabdalkader: I can just set it back again after, because abort() needs to clear it I think.

kwagyeman: That's fine, or pass an arg to abort not to clear it.

PR #2746, opened 2025-07-02
ports/all: Refactor CSI port init.

iabdalkader:
* Removes duplicate CSI initialization code (almost identical in every port).
* Makes it easier to make changes in the future (e.g., decouple the clock).
* `omv_csi_init` is still completely replaceable.

Notes:
* One change from old code, made all control pins `PULL_NONE`, just doesn't make much sense to select one pull or the other.
* Some ports require the IP clock to generate the external clock, so those have to initialize CSI earlier.
* There was an extra `csi->detected=True` in alif port, probably due to duplicate code, copy&paste this is no longer needed, it's done in common code.

While at it, nrf and rp2 were missing `csi_init0` (still unknown though if they actually work). Tested the rest of the ports (imxrt, stm, alif).

### DIFF
--- a/ports/alif/omv_csi.c
+++ b/ports/alif/omv_csi.c
@@ -71,6 +71,66 @@
                          CAM_INTR_OUTFIFO_OVERRUN | \
                          CAM_INTR_BRESP_ERR)
 
+static uint32_t omv_csi_get_fb_offset(omv_csi_t *csi);
+
+void CAM_IRQHandler(void) {
+    uint32_t mask = 0;
+    omv_csi_t *csi = omv_csi_get(-1);
+    CPI_Type *cpi = csi->base;
+
+    uint32_t status = cpi_get_interrupt_status(cpi);
+
+    if (status & CAM_INTR_VSYNC) {
+        mask |= CAM_INTR_VSYNC;
+    }
+
+    if (status & CAM_INTR_HSYNC) {
+        mask |= CAM_INTR_HSYNC;
+    }
+
+    if (status & CAM_INTR_INFIFO_OVERRUN) {
+        mask |= CAM_INTR_INFIFO_OVERRUN;
+        omv_csi_abort(csi, true, true);
+    }
+
+    if (status & CAM_INTR_OUTFIFO_OVERRUN) {
+        mask |= CAM_INTR_OUTFIFO_OVERRUN;
+        omv_csi_abort(csi, true, true);
+    }
+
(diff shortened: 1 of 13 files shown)


ports/alif/omv_csi.c
```diff
@@ -235,13 +294,12 @@ int alif_csi_snapshot(omv_csi_t *csi, image_t *dst_image, uint32_t flags) {
         cpi_irq_handler_clear_intr_status(cpi, CPI_IRQ_FLAGS);
         cpi_enable_interrupt(cpi, CPI_IRQ_FLAGS);
         NVIC_ClearPendingIRQ(CAM_IRQ_IRQn);
+        NVIC_SetPriority(CAM_IRQ_IRQn, IRQ_PRI_CSI);
         NVIC_EnableIRQ(CAM_IRQ_IRQn);
 
         // Reset CSI and start the capture.
-        cpi->CAM_CTRL = 0;
-        cpi->CAM_CTRL |= CAM_CTRL_SW_RESET;
+        cpi->CAM_CTRL = CAM_CTRL_SW_RESET;
```
kwagyeman, line 303: Did you test the AE3 and is it still working with this change? I remember this register being buggy and there being a reason for writing to the reg this way.

kwagyeman: I guess, you said you tested it.

Wonder why we wrote it that way... don't quite remember anymore.

kwagyeman (approved):

ports/alif/omv_csi.c
iabdalkader: > Wonder why we wrote it that way... don't quite remember anymore.

Exactly... I don't think there was any reason, I just wanted to disable the capture and reset, this does both.

iabdalkader: Ah so in the Alif "drivers" soft-reset is like this:
```C
void cpi_software_reset(CPI_Type *cpi)
{
    cpi->CAM_CTRL = 0;
    cpi->CAM_CTRL |= CAM_CTRL_SW_RESET;
    cpi->CAM_CTRL = 0;
    cpi->CAM_CTRL = (CAM_CTRL_START | CAM_CTRL_FIFO_CLK_SEL);
}
```
Doesn't say much about how it works in the datasheet, like is it instantaneous or does it require a few cycles?
image

kwagyeman: Not sure... Might want to leave how it previously works given the driver is stable right now. Note you also changed it in the CSI IRQ callback.

PR #2747, opened 2025-07-03
Kwabena/genx320 event mode

kwagyeman:


kwagyeman:
@nlyubova-psee - Can you take a look at this PR? I have some more testing to do with it. But, it's pretty much done. You should be able to run it on the H7 Plus or RT1062.

Please see the examples I added for how to use event mode. Note that generally, histogram mode (the default) produces much nicer-looking images than event mode can. I guess this is because the logic onboard the sensor is able to handle a much faster event data rate internally than what is possible with the EVT2.0 format over 8-bit parallel. So... the image generated by looking at the event data stream is somewhat of a subsampling of the true event rate.

I have some examples of the data rate without generating frames. For the H7 Plus, it can do about 250 FPS at max of 2048 event buffers. So, 512,000 events per second. This is less FPS than histogram mode, but, the processor has to due more work post-processing things.

Anyway, do you have any ideas for any kinda of simple things we can demonstrate on the event stream using numpy like array processing? I didn't have much inspiration beyond selecting all ON/OFF events and drawing the histogram image from RAW events.

...

EDIT: I think I may update this to allow the event buffer to be larger than 2048 events. This should result in nicer-looking event images as DMA can handle larger data bursts without CPU involvement. The smaller 2048 buffer size appears to be problematic because after the CPU starts post-processing, it drops the next buffer coming in while it's busy working. You can see the issue with the long exposure examples... all events don't seem to be captured.

lib/imlib/draw.c
```diff
@@ -708,6 +708,39 @@ void imlib_draw_string(image_t *img,
     }
 }
 
+void imlib_draw_event_histogram(image_t *img, ec_event_t *ec_event, int num_events, int gain) {
```
nlyubova-psee, line 713: Why not drawing a simple image what we call a CDFrame (pixel become white for ON event, and black for OFF events, otherwise stays gray)?  Any color pallets can be used.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode.py
```diff
@@ -0,0 +1,59 @@
...
+    # Reads 2048 events from the camera.
+    # Returns the number of valid events (0-2048) or a negative error code.
+    valid_events = sensor.ioctl(sensor.IOCTL_GENX320_READ_EVENTS, events)
+
+    # Draws the events on the image. If clear=True then the image is cleared
+    # to the default value of "brightness". Event histogram bins have "contrast"
+    # added to them for PIX_ON_EVENT events and subtracted from them for
+    # PIX_OFF_EVENT events clampped between 0 and 255. Pass clear=False to keep
+    # accumulating events in the histogram image.
+    img.draw_event_histogram(events, clear=True, brightness=128, contrast=64)
```
nlyubova-psee: Sorry for maybe a stupid question, why do you use "events" here and not "valid_events" ?

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_filtering.py
```diff
@@ -0,0 +1,65 @@
+# This work is licensed under the MIT license.
+# Copyright (c) 2013-2025 OpenMV LLC. All rights reserved.
+# https://github.com/openmv/openmv/blob/master/LICENSE
+#
+# This example shows off using the genx320 event camera from Prophesee
+# using event streaming mode.
```
nlyubova-psee, line 6: can we extend the description with something like this : "and filtering events based on their polarity (ON/OFF)" ?  To make it more different from the previous example.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_filtering.py
```diff
@@ -0,0 +1,65 @@
...
+#
+# Note that in event mode the processor has to post-process raw events
+# to populate the ndarray. This takes some CPU time but is relatively
+# efficent. However, drawing the event histogram is VERY inefficient
+# on the CPU. FOR MAXIMUM PERFORMANCE YOU NEED TO PROCESS EVENTS USING
+# ULAB NDARRAY FUNCTIONS AND NOT DRAW THE EVENT HISTOGRAM.
```
nlyubova-psee: is it drawing that is inefficient or computation of the histogram ?

nlyubova-psee (commented):
Hi @kwagyeman , thank you for your work !
I can test it only on Monday (sorry for delay).

No worry about "good looking images" as they are useful only for people and not for algorithms. Good data are much more useful for algorithms rather than a beautiful image.

It is sad to hear that "it drops the next buffer coming in while it's busy working". Can we improve it somehow ? For example, if we tune biases to get less events, then processing should be theoretically faster. Will it still drop data in this case ?

We will think about a good example (maybe frequency measurement).

lib/imlib/draw.c
kwagyeman: If you pump the contrast up you basically get that. Changing the color can be done after drawing the frame using the .to_evt_dark(copy=True) on the image. And then that can be flushed to the IDE via .flush().

When I increase the event buffer size you can get images comparable to the regular histogram mode with this.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode.py
kwagyeman: valid_events is a count of events. The ndarray events already contains its size in it's object so you don't have to pass it separately.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_filtering.py
kwagyeman: Sure

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_filtering.py
kwagyeman: Yeah, so, it's a random access to SDRAM where the image is stored. Each read/write accesses a completely different location. So, 2048 pixels being draw means up to 2048 new SDRAM row changes. These are costly because the entire architecture for the CPU is designed to access things in linear order.

Modern architectures can handle going backwards and forwards fine... albeit, forward access is guaranteed to be optimized in every silicon path and not just the CPU.

....

Given I'm already seeing performance issues... I'm thinking it will be faster to draw the image image using fb_alloc() in SRAM and then memcpy it to the final image location in SDRAM than trying to draw it in SDRAM directly.

I can also accelerate drawing the image using ARM Helium. There's scatter/gather support which can be leveraged. Not going to do that though in this PR.

kwagyeman: I looked into this and it's actually quite efficient. Only 150us for ~1500 events. However, this is because the image ends up in SRAM as the onchip heap is big enough to store 320x320 grayscale. If it were to end up off chip then there would be a big performance issue.

I will remove the note.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode.py
kwagyeman: Renamed

kwagyeman:
I've resolved the issue with events not producing smooth images. It was because the camera frame output didn't match the frame buffer size. I also added support for multiple event buffer sizes. Small buffers for fast returns and large ones for capturing more events.

I'm getting some ghosting still where events appear to be stuck, so, I think there's still some bugs in event mode to fix, but, all the pieces are here now I think.

Something very cool, with a long exposure continuously writing events to the same image buffer you can get grayscalish looking images out of the long exposure example.

https://github.com/user-attachments/assets/dcac1741-f50c-43db-8f17-e255a8a38029

modules/py_image.c
```diff
@@ -1269,7 +1269,7 @@ static MP_DEFINE_CONST_FUN_OBJ_KW(py_image_save_obj, 2, py_image_save);
 
 static mp_obj_t py_image_flush(mp_obj_t img_obj) {
     framebuffer_t *fb = framebuffer_get(0);
-    framebuffer_update_jpeg_buffer(fb);
+    framebuffer_update_jpeg_buffer(fb, py_image_cobj(img_obj));
```
iabdalkader: It was meant to update the FB after drawing to it, since all images originally were FB pointers. I think the cleaner way to fix this would be to refactor `framebuffer_update_jpeg_buffer` to accept an image, since that's what it really expects. You could probably reuse the image pointer passed to every `snapshot` (careful though as this might be NULL).

Could probably get away with:
```C
image_t framebuffer_to_image(framebuffer_t *fb) {
    image_t image = {
        .w = fb->w,
        .h = fb->h,
        .size = fb->size,
        .pixfmt = fb->pixfmt,
        .pixels = framebuffer_get_buffer(fb, fb->head)->data,
    }
    return image;
}

if (flags & OMV_CSI_CAPTURE_FLAGS_UPDATE) {
    image_t img = framebuffer_to_image(fb);
    framebuffer_update_jpeg_buffer(img);
}
```
Just one additional line before `update_jpeg_buffer`.

iabdalkader:
This is shaping up to be a large PR, best start splitting things otherwise it may take too long to process. Also, I feel the allocations in the Python scripts are very inefficient, so much so it needs a collect in the loop. Is there any way around that? Could we do that processing in C instead? It would also hide some of the complexity.

kwagyeman:
Only the ndarray manipulation needs a collection. Otherwise, everything I've created for allocating and writing to the ndarray doesn't. Not much you can do about memory allocations when using ULAB. It's not really even needed that much though, I just added it to make sure the timing result was really consistent.

modules/py_image.c
kwagyeman: I was actually going to do that initially, but, it calls a bunch of functions internally that need the FB pointer. However, none of the functions it passes the FB pointer to actually use the pointer... so, I can drop the pointer from those functions and almost everything in framebuffer_update_jpeg_buffer. 

But, it does check `fb->streaming_enabled` internally. How should I handle this? This expands the scope of the refactor.

kwagyeman:
Okay, I've resolved the old events issue. Just needed to slice the output event array by the valid event count which is a very fast operation.

In event mode with 1024 event buffers the frame rate can hit 2000 FPS!!! 2048 buffers (default for the script) hit 1000 FPS! Drawing slows down 2048 event buffers to 500 FPS. Still crazy fast. Below video with the H7 Plus.

https://github.com/user-attachments/assets/b8ba5550-e05e-4c64-be3a-89463acf35a6

Wow! The N6 is fast. It hits >2500 FPS at 1024 events. 300us per frame when not drawing the image. When drawing it's still 800 FPS!!!

RT1062 is working perfectly too! Hits 2000 FPS when not drawing the events and about 900 FPS when drawing events.

@nlyubova-psee - Ready for your review. Will do changes as requested by Ibrahim in the mean-time.

modules/py_image.c
kwagyeman: I thought about how to refactor this to get rid of the fb_streaming_enabled issue. Will send this in another PR.

iabdalkader: Let's see it, but on second look it does expect a framebuffer, so maybe just copy the image to be flushed to fb if it's not already the fb ? And let's remove the `src_image` confusing stuff, it's not needed.

kwagyeman: > so maybe just copy the image to be flushed to fb if it's not already the FB

Not possible, the FB in event mode is like 1024x4 pixels. Have to decouple.

iabdalkader: Not following? When calling `img.flush()` if the image != FB you could just memcpy it to fb, before calling `update_jpeg_buffer`.

kwagyeman: In event mode the fb is like 1024x4 bytes pixels which represents 1024 events (4-bytes per event). The generated image is 320x320 pixels. So, I can't even put the image in the frame buffer. I need to directly flush the image into the jpeg buffer.

iabdalkader: I see the problem. Okay, we could refactor `framebuffer_update_jpeg_buffer` to call a new function that accepts `image_t`, which could also be called by flush. `framebuffer_update_jpeg_buffer` can still do all the checks, like `streaming_enabled` etc.. before calling the new function. That mode that prints the image to serial, how useful is it? We could also remove it.

kwagyeman: > That mode that prints the image to serial, how useful is it? We could also remove it.

I'm thinking dropping it and just printing a message that the frame buffer can't be displayed because the image is too large. It only works on cameras that don't drop text like the STM32 series because they aren't using TinyUSB. But, we also don't need to support this backup anymore as you can just get a higher end OpenMV Cam now.

As for the refactor, I was actually going to completely remove the `streaming_enabled` and the logic in the `omv` module that controls it as you can just pass `update=False` to sensor now. So, it's not needed at all anymore.

iabdalkader: > I'm thinking dropping it and just printing a message that the frame buffer can't be displayed because the image is too large

Yes we should drop it. As for the message, and text dropping, I don't like printing a lot of messages. Can we draw the error on the frame buffer instead?

> As for the refactor, I was actually going to completely remove the streaming_enabled and the logic in the omv module that controls it as you can just pass update=False to sensor now. So, it's not needed at all anymore.

Okay, I never really liked it, seems confusing to have two flags that control streaming.

kwagyeman: > Yes we should drop it. As for the message, and text dropping, I don't like printing a lot of messages. Can we draw the error on the frame buffer instead?

Good idea. I can send an escape sequence and have OpenMV IDE decode it.

kwagyeman: https://github.com/openmv/openmv/pull/2755 - Will rebase after PR is merged.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_filtering.py
nlyubova-psee: In CDFrame, we can skip drawing all the events that happen for the same pixel over the accumulation period (you can simply set the pixel to 1 or 0 once). 
In case of large accumulation periods (that is the case for 250fps), usually more than 1 event happens per pixel and we do not need to draw all of them.

If you have already considered it, then please ignore my comment.

kwagyeman: From a performance standpoint, there's no difference in incrementing a counter per pixel position or just setting it once. The lookup dominates the operation time. So, might as well go with a counter versus a flag as this gives you 8-bit values versus just 0/255. Storing the histogram as a binary value (1-bit) would be slower due to the shifting and masking per pixel.

nlyubova-psee:
Thank you very much @kwagyeman for your great work! and very sorry for our late response.

We have tested your PR on GenX320 camera.
From a quick test, the streaming performance looks good.

Here are few comments:
* streaming with genx320_event_mode.py looks great
* streaming with genx320_event_mode_deep_buffer.py and genx320_event_mode_deep_buffer_filtering.py lags (some delays are visible) and its purpose is not clear to us. Could you please update its description to make it more clear ? (for the moment, it is the same # This example shows off using the genx320 event camera from Prophesee using event streaming mode.)
* streaming with genx320_event_mode_filtering.py looks ok
* streaming with genx320_event_mode_high_speed.py looks interesting, we will check the integrity of the output data  
* streaming with genx320_event_mode_long_exposure.py looks interesting, especially if we generate a timesurface (pixel value is set to the last timestamp instead of counting events per pixel) instead of a histogram

Maybe it make sense to rename the old examples by adding "histogram" into the script name ?  (to make clear which script uses which data format)

Few more generic comments:
* the sensor biases used by default (that are low sensitive) can be more sensitive to see more details (not urgent, we will maybe do a PR one day).
* I have an impression that ioctl for deactivating AFK filter does not work anymore as it always filter light flickering (we will re-test it to confirm later)

We consider that some more tests are required to prove the integrity of streamed data. In case of missing/lost data, it will be good to mention the limitation somewhere as a notice.

nlyubova-psee:
> Okay, I've resolved the old events issue. Just needed to slice the output event array by the valid event count which is a very fast operation.
> 
> In event mode with 1024 event buffers the frame rate can hit 2000 FPS!!! 2048 buffers (default for the script) hit 1000 FPS! Drawing slows down 2048 event buffers to 500 FPS. Still crazy fast. Below video with the H7 Plus.
> Recording.2025-07-04.155404.mp4
> 
> Wow! The N6 is fast. It hits >2500 FPS at 1024 events. 300us per frame when not drawing the image. When drawing it's still 800 FPS!!!
> 
> RT1062 is working perfectly too! Hits 2000 FPS when not drawing the events and about 900 FPS when drawing events.
> 
> @nlyubova-psee - Ready for your review. Will do changes as requested by Ibrahim in the mean-time.

Very interesting !
Visually it looks like you use a frame-based camera with a pulsed illumination.
Basically you see the trajectory of object motion (even if you have some gaps between positions), so it could be useful for fast object tracking.

However, I think a timesurface is more appropriate visualization here rather than a histogram. Actually, you have information which object position was the first one, the next ones and the last one but you show them equally in a frame (so we kind of loosing this temporal information in the frame)

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_deep_buffer.py
```diff
@@ -0,0 +1,55 @@
+# This work is licensed under the MIT license.
+# Copyright (c) 2013-2025 OpenMV LLC. All rights reserved.
+# https://github.com/openmv/openmv/blob/master/LICENSE
+#
+# This example shows off using the genx320 event camera from Prophesee
+# using event streaming mode.
```
nlyubova-psee: Could you please update the description to make the sample more clear (for example say that the buffer size is very large)?

nlyubova-psee (approved):
Even if we plan to test the data integrity, the PR is general is ok for me.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_deep_buffer.py
kwagyeman: @nlyubova-psee - Sure, will try to process updates to the PR tonight so you can re-review tomorrow morning.

Regarding this:

> streaming with genx320_event_mode_long_exposure.py looks interesting, especially if we generate a timesurface (pixel value is set to the last timestamp instead of counting events per pixel) instead of a histogram

If you just set the contrast to like 256 then this will force the pixel value to the latest event value in the histogram.

kwagyeman:
@nlyubova-psee - I've updated the examples with your requested changes. I've also gone ahead and removed support for this new feature in the old sensor API which is now deprecated in favor of the new CSI module and updated all the examples to match this. I've retested them all and they work.

Please let me know what you mean by "timesurface", like, if you just max out the contrast at 256 then you'll get an image where events are only the latest value as it will swing the histogram between 0 and 255 (adding +256 or -256). However, this doesn't seem like it needs a separate example.

Also, should we disable the AFK by default? It makes the image blocky often.

Anyway, please let me know any other thoughts. I think the PR is just about ready.

...

Regarding frequency measurement. ULAB has support for fft's this could somehow be used with the event ndarray to measure frequency. Not sure how exactly though.

nlyubova-psee:
> Also, should we disable the AFK by default? It makes the image blocky often.

Agree I prefer to deactivate it. 

> Please let me know what you mean by "timesurface", like, if you just max out the contrast at 256 then you'll get an image where events are only the latest value as it will swing the histogram between 0 and 255 (adding +256 or -256). However, this doesn't seem like it needs a separate example.

Timasurface is a 2D image where each pixel has the timestamp of the last event happened for this pixel. Thus, it integrates some history of motion over the frame. Visually it maybe does not help :) but algorithms can profit of this time information. Anyway, it is not important to do it. Just an idea.

nlyubova-psee:
> Regarding frequency measurement. ULAB has support for fft's this could somehow be used with the event ndarray to measure frequency. Not sure how exactly though.

I will try to check how to do frequency measurement example next week.

kwagyeman:
Okay, other than disabling AFK by default. Please test the examples and let me know if the PR is ready for Ibrahim to finish his code review.

kwagyeman:
@nlyubova-psee - Can you finish your review of this PR this week? Customers are asking... Let me know about the AFK.

iabdalkader:

https://forums.openmv.io/t/how-to-achieve-high-speed-led-frequency-detection-10s-of-khz-with-genx320-on-openmv-h7-plus/11207/11

drivers/sensors/genx320.c
```diff
@@ -581,6 +589,106 @@ static int ioctl(omv_csi_t *csi, int request, va_list ap) {
...
             break;
         }
+        case OMV_CSI_IOCTL_GENX320_READ_EVENTS: {
+            // Set event mode if not already set.
+            ret = set_event_mode(csi, true);
+            if (ret < 0) {
+                break;
+            }
+
+            if (is_valid_framesize(csi->framesize, true) < 0) {
```
iabdalkader: If I'm following this correctly, you should probably check here that the frame size is >= the ndarray size.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # Reads up to 2048 events from the camera.
+    # Returns the number of valid events (0-2048) or a negative error code.
+    # Note that old events in the buffer are not cleared to save CPU time.
+    event_count = csi0.ioctl(csi.IOCTL_GENX320_READ_EVENTS, events)
+
+    # Draws the events on the image. If clear=True then the image is cleared
+    # to the default value of "brightness". Event histogram bins have "contrast"
+    # added to them for PIX_ON_EVENT events and subtracted from them for
+    # PIX_OFF_EVENT events clampped between 0 and 255. Pass clear=False to keep
```
iabdalkader:
```suggestion
    # PIX_OFF_EVENT events clamped between 0 and 255. Pass clear=False to keep
```

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode.py
```diff
@@ -0,0 +1,56 @@
...
+# 0: event type
+# 1: seconds timestamp
+# 2: milliseconds timestamp
+# 3: microseconds timestamp
+# 4: x coordinate (0-319 for the genx320)
+# 5: y coordinate (0-319 for the genx320)
+events = np.zeros((2048, 6), dtype=np.uint16)
+
+# Initialize the sensor.
+csi0 = csi.CSI(cid=csi.GENX320)
```
iabdalkader, line 30:
```suggestion
csi0 = csi.CSI()
```
You only have to specify the ID if it's used with another sensor, otherwise you get the default and only sensor.

kwagyeman: Yes, my thought here is that by specifying the cid, the script will work for the upcoming dual module too. Is that sensible?

iabdalkader: Yes, good point. Though it will probably also require its own example(s).

kwagyeman: Yeah, of course. But, future proofing things here for these examples.

drivers/sensors/genx320.c
iabdalkader: Note that the ioctl could just set the frame size internally, based on the ndarray size, which would make it impossible for the user to set one res and pass a smaller ndarray. Though I'm torn between that and staying consistent with all the other scripts.

kwagyeman: The py_csi_ng layer checks that the array matches the ndarray size. So, you get a nice error message about the array size being wrong.

> Note that the ioctl could just set the frame size internally, based on the ndarray size, which would make it impossible for the user to set one res and pass a smaller ndarray. Though I'm torn between that and staying consistent with all the other scripts.

Yeah, it's possible, but, would need to make sure the ndarray passed a list of qualifications in the py_csi_ng layer. Like, I have to keep the dma sizes, line buffers, and etc. in mind for the frame sizes. Not impossible to handle though, but, it kinda means you aren't passing a frame size and pixformat anymore which makes things inconsistent with other scripts as you noticed.

Making it work either way is fine with me.

iabdalkader: Ah I see it's validate in csi.

> makes things inconsistent with other scripts as you noticed.

Yeah, let's keep it this way.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode.py
kwagyeman: Fixed

kwagyeman:
Prophesee informed me they will have completed a final review of things by mid-next week.

lib/imlib/draw.c
```diff
@@ -708,6 +708,39 @@ void imlib_draw_string(image_t *img,
...
+                        if ((x < img->w) && (y < img->h)) {
+                            int index = (y * img->w) + x;
+                            img->data[index] = __USAT(((int32_t) img->data[index]) + gain, UINT8_T_BITS);
+                        }
+                        break;
+                    }
+                    default: {
+                        break;
+                    }
+                }
```
iabdalkader: Are there other types that could be drawn differently?
```suggestion
                size_t index = (y * img->w) + x;
                int32_t delta = (ec_event[i].type == EC_PIX_OFF_EVENT) ? -gain : gain;
                img->data[index] = __USAT(((int32_t) img->data[index]) + delta, UINT8_T_BITS);
```

drivers/sensors/genx320.c
```diff
@@ -82,50 +76,69 @@
...
 #define EHC_DIFF3D_N_BITS_SIZE          (7) // signed 8-bit value
 
-#if (OMV_GENX320_CAL_ENABLE == 1)
-static bool hot_pixels_disabled = false;
-#endif // (OMV_GENX320_CAL_ENABLE == 1)
+static bool reset_called = false;
+static bool event_mode = false;
 static int32_t contrast = CONTRAST_DEFAULT;
 static int32_t brightness = BRIGHTNESS_DEFAULT;
+static uint64_t event_time = 0;
```
iabdalkader: You should group these variables in a state struct `genx_state_t` (like vospi). You could also add the snapshot pointer to that struct, and set a pointer to it in `csi->priv`, or just access it from global variable.

Also, I think if we just add an `issd` pointer to this struct, it will simplify a lot of code.

```C
typedef enum _genx_mode {
    GENX_MODE_HISTO,
    GENX_MODE_EVENT,
} genx_mode_t;

typedef struct _genx_state {
    int32_t contrast;
    int32_t brightness;
    uint64_t event_time_ms;
    const struct issd *issd;
    genx_mode_t mode;
    omv_csi_snapshot_t snapshot;
} genx_state_t;

static genx_state_t genx;
```

drivers/sensors/genx320.c
```diff
@@ -82,50 +76,69 @@
...
+
+static int set_event_mode(omv_csi_t *csi, bool enable) {
+    if (reset_called) {
+        if (event_mode == enable) {
+            return 0;
+        }
+
+        psee_sensor_stop(csi, event_mode ? &dcmi_evt : &dcmi_histo);
+        psee_sensor_destroy(csi);
+    }
```
iabdalkader:
```suggestion
static int genx_set_mode(omv_csi_t *csi, genx_mode_t mode) {
    genx_state_t *genx = csi->pirv; // Or global doesn't matter
      
    if (genx->mode == mode) {
        return 0;
    }

    if (genx->issd) {
        psee_sensor_stop(csi, gnex->issd);
        psee_sensor_destroy(csi);
    }
    
    genx->issd = (mode == GENX_MODE_EVENT) ? &dcmi_evt : &dcmi_histo;
```

drivers/sensors/genx320.c
```diff
@@ -82,50 +76,69 @@
...
 
     // Configure Packet and Frame sizes
-    psee_sensor_write(csi, CPI_PACKET_SIZE_CONTROL, ACTIVE_SENSOR_WIDTH);
-    psee_sensor_write(csi, CPI_PACKET_TIME_CONTROL, ACTIVE_SENSOR_WIDTH << CPI_PACKET_TIME_CONTROL_PERIOD_Pos |
-                      HSYNC_CLOCK_CYCLES << CPI_PACKET_TIME_CONTROL_BLANKING_Pos);
-    psee_sensor_write(csi, CPI_FRAME_SIZE_CONTROL, ACTIVE_SENSOR_HEIGHT);
+    int packet_width = enable ? resolution[csi->framesize][0] : ACTIVE_SENSOR_WIDTH;
+    int packet_height = enable ? resolution[csi->framesize][1] : ACTIVE_SENSOR_HEIGHT;
+    int packet_hsync = enable ? EVENT_HSYNC_CLOCK_CYCLES : HISTO_HSYNC_CLOCK_CYCLES;
```
iabdalkader:
```suggestion
    uint32_t packet_width = enable ? resolution[csi->framesize][0] : ACTIVE_SENSOR_WIDTH;
    uint32_t packet_height = enable ? resolution[csi->framesize][1] : ACTIVE_SENSOR_HEIGHT;
    uint32_t packet_hsync = enable ? EVENT_HSYNC_CLOCK_CYCLES : HISTO_HSYNC_CLOCK_CYCLES;
```

drivers/sensors/genx320.c
```diff
@@ -159,40 +173,57 @@ static int reset(omv_csi_t *csi) {
...
+    reset_called = false;
+    event_mode = false;
+    contrast = CONTRAST_DEFAULT;
+    brightness = BRIGHTNESS_DEFAULT;
+    event_time = 0;
+
+    // Set histogram mode by default.
+    return set_event_mode(csi, false);
+}
+
```
iabdalkader:
```suggestion
static int reset(omv_csi_t *csi) {
    genx_state_t *genx = csi->priv;
    
    csi->color_palette = NULL;
    genx->contrast = CONTRAST_DEFAULT;
    genx->brightness = BRIGHTNESS_DEFAULT;
    genx->event_time = 0;

    // Set histogram mode by default.
    if (genx_set_mode(csi, GENX_MODE_HISTO)) {
        return OMV_CSI_ERROR_CSI_INIT_FAILED;
    }

    return 0;
}
```

drivers/sensors/genx320.c
```diff
@@ -159,40 +173,57 @@ static int reset(omv_csi_t *csi) {
...
+                              INTEGRATION_DEF_PREIOD, EHC_WITHOUT_PADDING) != EHC_OK) {
+            return -1;
+        }
     }
-    #else
-    psee_sensor_start(csi, &dcmi_evt);
-    #endif //  (OMV_GENX320_EHC_ENABLE == 1)
+
+    reset_called = true;
+    event_mode = enable;
```
iabdalkader:
```suggestion
```

drivers/sensors/genx320.c
```diff
@@ -220,11 +251,14 @@ static int set_pixformat(omv_csi_t *csi, pixformat_t pixformat) {
...
 static int set_framesize(omv_csi_t *csi, omv_csi_framesize_t framesize) {
-    return (framesize == OMV_CSI_FRAMESIZE_320X320) ? 0 : -1;
+    return is_valid_framesize(framesize, false);
 }
 
 static int set_framerate(omv_csi_t *csi, int framerate) {
-    #if (OMV_GENX320_EHC_ENABLE == 1)
+    if (event_mode) {
+        return -1;
+    }
```
iabdalkader:
```suggestion
    if (genx->mode == GENX_MODE_EVENT) {
        return -1;
    }
```

drivers/sensors/genx320.c
```diff
@@ -590,10 +698,17 @@ static int ioctl(omv_csi_t *csi, int request, va_list ap) {
...
+    id &= 0x7FFFFFFF;
+    return (id == (GENX320_ID_ES & 0x7FFFFFFF)) ||
+           (id == (GENX320_ID_MP & 0x7FFFFFFF));
+}
+
 int genx320_init(omv_csi_t *csi) {
     // Initialize csi structure
     csi->reset = reset;
     csi->sleep = sleep;
+    csi->match = match;
```
iabdalkader, line 726:
```suggestion
    csi->match = match;
    
    memset(&genx, 0, sizeof(genx_state_t));
    genx.snapshot = csi->snapshot;
    csi->priv = &genx; // Or could keep it global.
    
```

iabdalkader (changes requested):

lib/imlib/draw.c
kwagyeman: There are other event types in the buffer that have to be ignored. The trigger ones. So, I still need to check for OFF/ON per message. I can update to get rid of the inner switch though. Updated.

drivers/sensors/genx320.c
kwagyeman: Done

drivers/sensors/genx320.c
kwagyeman: Done

drivers/sensors/genx320.c
kwagyeman: Done

drivers/sensors/genx320.c
kwagyeman: Done

drivers/sensors/genx320.c
kwagyeman: Done

drivers/sensors/genx320.c
kwagyeman: Done

drivers/sensors/genx320.c
kwagyeman: Done, the mode check has to be gated though by the the issd being valid otherwise you'd do nothing for the first call.

kwagyeman:
Applied all changes, but event mode is not working anymore... after doing so and rebasing. Debugging things right now... histogram mode still works. Not sure if it's the changes or other PRs to master after the rebase.

Will push after I fix the issue with this.

drivers/sensors/genx320.c
iabdalkader: You could just add this:
```C
typedef enum _genx_mode {
    GENX_MODE_INVALID = 0,
    GENX_MODE_HISTO,
    GENX_MODE_EVENT,
} genx_mode_t;
```

Note I don't see any changes to this file.

kwagyeman: Now updated.

kwagyeman:
Okay, I fixed the bugs with the PR and applied all suggestions. It's now working on the H7 Plus, RT1062, and N6 correctly. Event mode stopped working because with the new snapshot code, omv_csi_snapshot() has to be called versus the snapshot platform implementation; otherwise, the buffer management doesn't happen.

I made this work by adding a new flag, `OMV_CSI_CAPTURE_FLAGS_BYPASS`, which the calls inside the `ioctl` pass to `omv_csi_snapshot()`, which then gets passed to the driver `snapshot()` wrapper to just call the platform snapshot function and return versus doing any post-processing on the received data to generate the histogram image. Only the genx320 driver needs this, all other drivers overriding snapshot are currently fine. I tried other approaches to fix this but, this seems like the cleanest. Let me know if you have a better idea for this.

The PR is now rebased on master.

drivers/sensors/genx320.c
```diff
@@ -160,39 +189,61 @@ static int reset(omv_csi_t *csi) {
...
+
+static int reset(omv_csi_t *csi) {
+    genx_state_t *genx = csi->priv;
+
+    csi->color_palette = NULL;
+    genx->contrast = CONTRAST_DEFAULT;
+    genx->brightness = BRIGHTNESS_DEFAULT;
+    genx->event_time_us = 0;
+    genx->issd = NULL;
+    genx->mode = GENX_MODE_INVALID;
```
iabdalkader: If you reset those variables, calling reset a second time will Not cause `psee_sensor_stop` and `psee_sensor_destroy` to be called from set_mode. If you remove them, it will work correctly: on reset the state is in BSS so it's zeroed so no stop&destroy. On following reset calls, they'll be set and it will always stop&destroy.

iabdalkader:
> Let me know if you have a better idea for this.

I don't follow. That flag is only set in event mode, so check if it's in event mode?

I see now what's going on. Yes the common code caused an unforeseeable issue, but only because you want to alternate between events and regular snapshots automatically, right? Is that even a use case? I don't think it should be, you've explicitly set the frame size to `EVT_32768` (so clearly don't want an image later), and even if you do that's why we have the draw_histo in case you also need to draw that image. So I think you should add an ioctl to set the mode explicitly and once set it just works in that mode. This also removes the need to set pix/size in snapshot like that, and perhaps also the need to call set_pixformat/set_framesize form the script in event mode?

To recap: There should be two mutually exclusive modes:
* Histo/Snapshot mode.
* Event mode.

There should be no need to alternate between the two, because in event mode you can also visualize the events read via draw_histogram.

drivers/sensors/genx320.c
kwagyeman: > If you reset those variables, calling reset a second time will Not cause psee_sensor_stop and psee_sensor_destroy to be called from set_mode.

Yes, but, reset() does a hardware reset of the camera. I tried removing them being set to the default and the camera stops working. psee_sensor_stop/psee_sensor_destroy don't appear to be safe to call after a hardware reset.

Hmmm, but, for dual camera mode this would be an issue as the camera won't necessarily be hardware reset. I'll debug why calling these locks up the camera after a hardware reset.

kwagyeman:
Yeah, I think preventing alternation fixes the issue. Setting the frame size to one of the events modes is enough for me to know if we are in event or histogram mode.

iabdalkader:
> Setting the frame size to one of the events modes is enough for me to know if we are in event or histogram mode.

Do we need to set pixformat/framesize at all for event mode if snapshot will not be called? Can't you infer the size of the buffer from read_events? BTW I think this works for us because if not set (i.e., in event mode) snapshot() should fail.

kwagyeman:
It's more than the whole camera driver system needs the frame size set. Lots of code configures the DMA system based on the frame size. So, if that's not set then nothing will work right. Then, not all frame sizes are safe to use and pass DMA size rules and etc. so, handling any buffer size won't work.

iabdalkader:
> It's more than the whole camera driver system needs the frame size set. Lots of code configures the DMA system based on the frame size

Set it from the driver? Just like what you have been doing for snapshot? I feel an ioctl to set mode + read events should would make a very clear and concise API:

```Python
csi0.reset()
csi0.ioctl(SET_MODE, EVENT)? 

while True:
   csi0.ioctl(READ_EVENTS, ...)
```

Remember after the mode changes, you won't be calling snapshot at all. It will be called from the ioctl, which could then set the framesize/format before calling it. Would this work?

kwagyeman:
Yeah, I can do that.

kwagyeman:
Still working on this, will have it done tomorrow.

drivers/sensors/genx320.c
kwagyeman: Okay, resolved this, with set_mode, everything became simpler, and I was able to remove resetting these variables.

kwagyeman:
Set mode is implemented, code is simpler too. API is as suggested now.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_calibration.py
```diff
@@ -0,0 +1,68 @@
...
+# 3: microseconds timestamp
+# 4: x coordinate (0-319 for the genx320)
+# 5: y coordinate (0-319 for the genx320)
+events = np.zeros((2048, 6), dtype=np.uint16)
+
+# Initialize the sensor.
+csi0 = csi.CSI(cid=csi.GENX320)
+csi0.reset()
+csi0.ioctl(csi.IOCTL_GENX320_SET_MODE, csi.GENX320_MODE_EVENT,
+           csi.EVT_2048)  # Must be EVT_1024/2048/.../65536
```
iabdalkader:
```suggestion
# Event size must be a power of two between 1024 and 65536.
csi0.ioctl(csi.IOCTL_GENX320_SET_MODE, csi.GENX320_MODE_EVENT, csi.EVT_2048)
```
The linter allows 120 chars. Please check the other scripts too.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_calibration.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # Reads 2048 events from the camera.
+    # Returns the number of valid events (0-2048) or a negative error code.
+    event_count = csi0.ioctl(csi.IOCTL_GENX320_READ_EVENTS, events)
+
+    # Draws the events on the image. If clear=True then the image is cleared
+    # to the default value of "brightness". Event histogram bins have "contrast"
+    # added to them for PIX_ON_EVENT events and subtracted from them for
+    # PIX_OFF_EVENT events clamped between 0 and 255. Pass clear=False to keep
+    # accumulating events in the histogram image.
```
iabdalkader:
```suggestion
    # Render events into a histogram image.
    # If clear=True, the image is reset to "brightness" before drawing.
    # For each PIX_ON_EVENT, add "contrast" to the bin value;
    # for each PIX_OFF_EVENT, subtract it and clamp to [0, 255].
    # If clear=False, histogram accumulates over multiple calls.
```
Better? Please use AI to review/refine these comments.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_calibration.py
```diff
@@ -0,0 +1,68 @@
...
+# ndarray to hold events from the camera
+# must be EVT_res events by 6 values
+#
+# 0: event type
+# 1: seconds timestamp
+# 2: milliseconds timestamp
+# 3: microseconds timestamp
+# 4: x coordinate (0-319 for the genx320)
+# 5: y coordinate (0-319 for the genx320)
+events = np.zeros((2048, 6), dtype=np.uint16)
```
iabdalkader:
```suggestion
# Stores camera events
# Shape: (EVT_res, 6) where EVT_res is the event resolution
# Columns:
#   [0]  Event type
#   [1]  Seconds timestamp
#   [2]  Milliseconds timestamp
#   [3]  Microseconds timestamp
#   [4]  X coordinate 0 to 319 for GENX320
#   [5]  Y coordinate 0 to 319 for GENX320
events = np.zeros((2048, 6), dtype=np.uint16)
```

drivers/sensors/genx320.c
```diff
@@ -563,24 +580,152 @@ static int ioctl(omv_csi_t *csi, int request, va_list ap) {
...
+                    break;
+                }
+
+                ret = omv_csi_set_framesize(csi, framesize);
+                if (ret < 0) {
+                    break;
+                }
+            } else {
+                ret = -1;
+            }
```
iabdalkader: `set_framesize` already validates if framesize is compatible with the mode.

I'd put a pin in this until you check the custom frame size thing.

```suggestion
            int framesize = va_arg(ap, int);

            if (mode != OMV_CSI_GENX320_MODE_HISTO || mode != OMV_CSI_GENX320_MODE_EVENT) {
                ret = -1;
                break;
            }

            if ((ret = genx_set_mode(csi, mode, OMV_CSI_FRAMESIZE_320X320)) {
                break;
            }

            if ((ret = omv_csi_set_framesize(csi, framesize))) {
                break;
            }

            if ((ret = omv_csi_set_pixformat(csi, PIXFORMAT_GRAYSCALE))) {
                break;
            }
```

common/omv_csi.h
```diff
@@ -188,6 +188,16 @@ typedef enum {
     OMV_CSI_FRAMESIZE_QXGA,     // 2048x1536
     OMV_CSI_FRAMESIZE_WQXGA,    // 2560x1600
     OMV_CSI_FRAMESIZE_WQXGA2,   // 2592x1944
+    #if (OMV_GENX320_ENABLE == 1)
+    // Event Resolutions
```
iabdalkader: I have an idea that might help us get rid of these constants. It's so simple: We just add a dummy custom resolution that gets set by drivers. The res table is mutable, so you can update it. Nothing else has to change:
```C
uint16_t resolution[][2] = {
    {0,    0   },
    {0,    0   },    /* Custom set by drivers */
```

The `ioctl` can use it like:
```C
if (mode == OMV_CSI_GENX320_MODE_EVENT) {
    size_t ndarray_size = va_arg(ap, int);
   // Validate & set the frame size
    if (ndarray_size < 1024 || ndarray_size > 65536 || !is_power_of_two(ndarray_size)) {
        return -1;
    }

    // After this point, any code that indexes the resolution table
    // using csi->framesize or OMV_CSI_FRAMESIZE_CUSTOM will see the
    // custom width/height set here.
    resolution[OMV_CSI_FRAMESIZE_CUSTOM][0] = 1024;
    resolution[OMV_CSI_FRAMESIZE_CUSTOM][1] = ndarray_size >> 8;

    if ((ret = genx_set_mode(csi, OMV_CSI_GENX320_MODE_EVENT))) {
        return -1;
    }

    if ((omv_csi_set_pixformat(csi, PIXFORMAT_GRAYSCALE))) {
        return -1; 
    }
    
    if ((omv_csi_set_framesize(csi, OMV_CSI_FRAMESIZE_CUSTOM))) {
        return -1;
    }
}
```

Finally the ioctl can just pass ndarray size:
```Python
# or pass events.shape[0] or whatever
csi0.ioctl(csi.IOCTL_GENX320_SET_MODE, csi.GENX320_MODE_EVENT, 2048)
```

Note that more constants take more RAM, as the res table is in data now, and require more docs. Also, every time we need some arbitrary resolution we'll have to add a new constant for it.

We could use this custom size internally for now (not export to Python). If we want to export it at some point, we could make set_framesize accept tuple, and drivers could validate the custom size and accept/reject it.

drivers/sensors/genx320.c
```diff
@@ -82,50 +76,73 @@
...
+        }
+    }
 
-static AFK_HandleTypeDef psee_afk;
+    return -1;
+}
 
-static int reset(omv_csi_t *csi) {
-    csi->color_palette = NULL;
+static int genx_set_mode(omv_csi_t *csi, genx_mode_t mode, int framesize) {
```
iabdalkader: Just for consistency, since you're not using the prefix for any other functions.

```suggestion
static int set_active_mode(omv_csi_t *csi, genx_mode_t mode, int framesize) {
```

drivers/sensors/genx320.c
```diff
@@ -82,50 +76,73 @@
...
-    psee_sensor_write(csi, CPI_PACKET_SIZE_CONTROL, ACTIVE_SENSOR_WIDTH);
-    psee_sensor_write(csi, CPI_PACKET_TIME_CONTROL, ACTIVE_SENSOR_WIDTH << CPI_PACKET_TIME_CONTROL_PERIOD_Pos |
-                      HSYNC_CLOCK_CYCLES << CPI_PACKET_TIME_CONTROL_BLANKING_Pos);
-    psee_sensor_write(csi, CPI_FRAME_SIZE_CONTROL, ACTIVE_SENSOR_HEIGHT);
+    uint32_t packet_width = resolution[framesize][0];
+    uint32_t packet_height = resolution[framesize][1];
+    uint32_t packet_hsync = (mode == OMV_CSI_GENX320_MODE_EVENT) ? EVENT_HSYNC_CLOCK_CYCLES : HISTO_HSYNC_CLOCK_CYCLES;
+    psee_sensor_write(csi, CPI_PACKET_SIZE_CONTROL, packet_width);
+    psee_sensor_write(csi, CPI_PACKET_TIME_CONTROL, packet_width << CPI_PACKET_TIME_CONTROL_PERIOD_Pos |
+                      packet_hsync << CPI_PACKET_TIME_CONTROL_BLANKING_Pos);
```
iabdalkader: Just some splits for readability.
```suggestion
    uint32_t packet_hsync = (mode == OMV_CSI_GENX320_MODE_EVENT) ?
                            EVENT_HSYNC_CLOCK_CYCLES : HISTO_HSYNC_CLOCK_CYCLES;
    psee_sensor_write(csi, CPI_PACKET_SIZE_CONTROL, packet_width);
    psee_sensor_write(csi, CPI_PACKET_TIME_CONTROL,
                      packet_width << CPI_PACKET_TIME_CONTROL_PERIOD_Pos |
                      packet_hsync << CPI_PACKET_TIME_CONTROL_BLANKING_Pos);
```

drivers/sensors/genx320.c
```diff
@@ -563,24 +580,152 @@ static int ioctl(omv_csi_t *csi, int request, va_list ap) {
...
+            image_t image;
+            ret = omv_csi_snapshot(csi, &image, 0);
+            if (ret < 0) {
+                break;
+            }
+
+            // Invalidate frame.
+            csi->fb->pixfmt = PIXFORMAT_INVALID;
+
+            ret = event_post_process(genx, (ec_event_t *) va_arg(ap, ec_event_t *), &image);
```
iabdalkader: Could we move this to snapshot? Also do we need to invalidate on every call or just once when setting the mode?
```suggestion
            // Invalidate frame.
            csi->fb->pixfmt = PIXFORMAT_INVALID;

            ret = event_post_process(genx, (ec_event_t *) va_arg(ap, ec_event_t *), &image);
```

drivers/sensors/genx320.c
```diff
@@ -341,127 +394,91 @@ static void disable_hot_pixels(omv_csi_t *csi, uint8_t *histogram) {
...
-        fb_free();
-        hot_pixels_disabled = true;
     }
-    #endif // (OMV_GENX320_CAL_ENABLE == 1)
 
-    int ret = ((omv_csi_snapshot_t) csi->priv)(csi, image, flags);
+    return valid_count;
+}
+
+static int snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
```
iabdalkader: Just for consistency, `snapshot` is typically the last function before init (i.e., move after ioctl).

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_calibration.py
```diff
@@ -0,0 +1,68 @@
...
+    # Reads 2048 events from the camera.
+    # Returns the number of valid events (0-2048) or a negative error code.
+    event_count = csi0.ioctl(csi.IOCTL_GENX320_READ_EVENTS, events)
+
+    # Draws the events on the image. If clear=True then the image is cleared
+    # to the default value of "brightness". Event histogram bins have "contrast"
+    # added to them for PIX_ON_EVENT events and subtracted from them for
+    # PIX_OFF_EVENT events clamped between 0 and 255. Pass clear=False to keep
+    # accumulating events in the histogram image.
+    img.draw_event_histogram(events[:event_count], clear=True, brightness=128, contrast=64)
```
iabdalkader, line 54: Should we call this just `draw_events`? And have it flush the buffer?

drivers/sensors/genx320.c
kwagyeman: Okay

common/omv_csi.h
kwagyeman: Yeah, I like this, also, if the ndarray is a power of two it will generally pass DMA size rules. Of course, it can't be too small, but, what you have basically handles the issue.

drivers/sensors/genx320.c
kwagyeman: Snapshot doesn't have an arg to write to the ndarray, so, it can't be in that. Also, I have to disable the post-processing when doing calibration as I need to raw data so it would make that more complex.

The invalidate I think is obsolete, it was needed when you were able to switch things back and forth between event and histogram mode. I can probably remove, but, it will need to be called when changing modes to prevent a stale image from sticking around and producing a glitchy output trying to visualize an event data stream as a histogram.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_calibration.py
kwagyeman: I didn't want to call it draw_events() as I think that would be like drawing crosses and etc. The 2D image is an event histogram, it's better to have the long name not to confuse people on this.

As for flushing the buffer, it's not really meant to be like a snapshot method. I prefer the independent flush call.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_calibration.py
kwagyeman: Updating my tool settings...

drivers/sensors/genx320.c
iabdalkader: I see, it's also called from `OMV_CSI_IOCTL_GENX320_CALIBRATE`. It's still possible though, if you like, you could just store the ndarray pointer in `priv`. If set do post_process. And calibrate should set it to NULL.

The invalidate seems like it should be in set_mode.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_calibration.py
iabdalkader: > I didn't want to call it draw_events() as I think that would be like drawing crosses and etc.

The events are not a generic histogram right? If they were we would have called it `draw_histogram`. The function draws something very specific with its own format. How about `draw_ec_events` ? Or `draw_genx_events`?

Okay about `flush`, just wanted to save a step, but it is better to be explicit.

kwagyeman: Maybe `draw_ec_event_histogram`? I made the ec event stuff to abstract away from the genx320 so everything outside of the camera driver has a more generic name to allow us to support future higher resolution event cameras.

Fine with `draw_ec_events` too, drawing events in a histogram is the only way to really visualize them, the histogram part is redundant.

iabdalkader: Yeah `draw_ec_events` is fine. What does EC stand for?

kwagyeman: Event Camera :)

iabdalkader: Actually, I think it should be `draw_histogram` you don't pass the whole events array right? Just a slice which is probably the histogram? It would be great if it's generic enough to visualize any histogram.

kwagyeman: Uh, it only applies for the ec event data structure though. Regarding the slice, that's just me selecting the valid events from the event buffer. Probably `draw_ec_histogram` would capture the name well.

iabdalkader: Let's just leave it as is, resolving this.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_calibration.py
kwagyeman: Done

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_calibration.py
kwagyeman: Done

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_calibration.py
kwagyeman: Done

common/omv_csi.h

drivers/sensors/genx320.c
kwagyeman: Done

drivers/sensors/genx320.c
kwagyeman: Done

drivers/sensors/genx320.c
kwagyeman: Done

drivers/sensors/genx320.c
kwagyeman: Moved invalidate to `set_active_mode`. Would like to leave the rest as is.

kwagyeman:
Okay, all changes done. Custom resolution mode is working great.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_deep_buffer_filtering.py
```diff
@@ -0,0 +1,63 @@
...
+# ndarray to hold events from the camera
+# must be EVT_res events by 6 values
+#
+# 0: event type
+# 1: seconds timestamp
+# 2: milliseconds timestamp
+# 3: microseconds timestamp
+# 4: x coordinate (0-319 for the genx320)
+# 5: y coordinate (0-319 for the genx320)
+events = np.zeros((32768, 6), dtype=np.uint16)  # Must be 1024/2048/.../65536
```
iabdalkader: You missed updating the comments in some of the examples to match the others. This is one of them. Also, while we're at it, I'd remove the trailing comment and add it to the array comments more clearly, for example:
```Python
# Stores camera events
# Shape: (EVT_res, 6) where EVT_res is the event resolution
# EVT_res: must be a power of two between 1024 and 65536.
# Columns:
#   [0]  Event type
#   [1]  Seconds timestamp
#   [2]  Milliseconds timestamp
#   [3]  Microseconds timestamp
#   [4]  X coordinate 0 to 319 for GENX320
#   [5]  Y coordinate 0 to 319 for GENX320
```

drivers/sensors/genx320.c
```diff
@@ -590,10 +704,32 @@ static int ioctl(omv_csi_t *csi, int request, va_list ap) {
     return ret;
 }
 
+static int snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
```
iabdalkader: I've added a post-process op here: https://github.com/openmv/openmv/pull/2803. I see now how it can be useful for this driver and others. You can remove this `snapshot` if you want. If you do please move the post processing functions here instead (call them `post_process_histo`, `post_process_event`). Note you will need to store the events buffer in the state, as well as the current the post-processing function (which should be set in `set_active_mode`, depending on the mode). After that, it should just work.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_histogram_mode_grayscale_led_tracking.py
```diff
@@ -9,24 +9,25 @@
...
-sensor.reset()
-sensor.set_pixformat(sensor.GRAYSCALE)  # Must always be grayscale.
-sensor.set_framesize(sensor.B320X320)  # Must always be 320x320.
-sensor.set_brightness(128)  # Leave at 128 generally (this is the default).
-sensor.set_contrast(16)  # Increase to make the image pop.
-sensor.set_framerate(200)
+csi0 = csi.CSI(cid=csi.GENX320)
+csi0.reset()
+csi0.pixformat(csi.GRAYSCALE)  # Must always be grayscale.
+csi0.framesize(csi.B320X320)  # Must always be 320x320.
```
iabdalkader: I removed all these custom resolution constants in https://github.com/openmv/openmv/pull/2804 let's merge that first and then just use the w/h directly.

```suggestion
csi0.framesize((320, 320))
```

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_deep_buffer_filtering.py
kwagyeman: Done

drivers/sensors/genx320.c
kwagyeman: Done

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_histogram_mode_grayscale_led_tracking.py
kwagyeman: Done, testing everything now.

kwagyeman:
PR Updated. Everything is tested.

iabdalkader (approved):

iabdalkader:
Merged!

PR #2749, opened 2025-07-04
 ports/stm32: Refactor DMA code.

iabdalkader:
I still need to refactor `py_audio.c` but that will require retesting every board with a MIC. Will update this later.

### DIFF
--- a/ports/stm32/omv_csi.c
+++ b/ports/stm32/omv_csi.c
@@ -62,6 +62,10 @@
 #define DCMIPP_PIPE         (DCMIPP_PIPE1)
 #endif
 
+#ifndef OMV_CSI_DMA_XFER_PORTS
+#define OMV_CSI_DMA_XFER_PORTS (0)
+#endif
+
 #define LINE_WIDTH_ALIGNMENT (16)
 
 extern uint8_t _line_buf;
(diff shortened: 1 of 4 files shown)


ports/stm32/omv_csi.c
```diff
@@ -102,37 +106,19 @@ void omv_csi_mdma_irq_handler(void) {
...
-        csi->dma.Init.FIFOThreshold = DMA_FIFO_THRESHOLD_FULL;
-        csi->dma.Init.MemBurst = DMA_MBURST_INC4;
-        csi->dma.Init.PeriphBurst = DMA_PBURST_SINGLE;
-    
-        // Initialize the DMA stream
-        HAL_DMA_DeInit(&csi->dma);
-        if (HAL_DMA_Init(&csi->dma) != HAL_OK) {
+        // Configure and initialize DMA.
+        if (stm_dma_init(&csi->dma, OMV_CSI_DMA_CHANNEL, OMV_CSI_DMA_REQUEST,
+                    DMA_PERIPH_TO_MEMORY, true, OMV_CSI_DMA_XFER_PORTS, &stm_dma_csi_init)) {
```
kwagyeman: It looks like circular mode is getting set to true for the F4/F7/H7 versus normal mode.

Mode will end up being DMA_CIRCULAR versus DMA_NORMAL for the older boards.

ports/stm32/stm_dma.c
```diff
@@ -38,6 +40,89 @@ static DMA_HandleTypeDef *dma_handle[32];
...
+    #if defined(STM32H7)
+    .Request = 0,
+    #else
+    .Channel = 0,
+    #endif
+    .Direction = 0,
+    .PeriphInc = DMA_PINC_DISABLE,
+    .MemInc = DMA_MINC_ENABLE,
+    // Note for F4/F7, used byte or half word.
+    .PeriphDataAlignment = DMA_PDATAALIGN_WORD,
```
kwagyeman: dma_init->PeriphDataAlignment = (config->datasize == 8) ? DMA_PDATAALIGN_BYTE : DMA_PDATAALIGN_HALFWORD;

Seems to be missing for the F4/F7. Without this SPI DMA will be broken.

ports/stm32/stm_dma.c
```diff
@@ -38,6 +40,89 @@ static DMA_HandleTypeDef *dma_handle[32];
...
+    #elif defined(STM32F4) || defined(STM32F7) || defined(STM32H7)
+    #if defined(STM32H7)
+    .Request = 0,
+    #else
+    .Channel = 0,
+    #endif
+    .Direction = 0,
+    .PeriphInc = DMA_PINC_DISABLE,
+    .MemInc = DMA_MINC_ENABLE,
+    // Note for F4/F7, used byte or half word.
```
kwagyeman: This note is applicable for SPI DMA, not the DCMI DMA.

ports/stm32/stm_dma.c
iabdalkader: Would half word work? Can you test it? I don't have F4/F7 board.

kwagyeman: No, you have to match the datawidth exactly because there's no fifo on the SPI hardware. So, it's not going to unpack values into the datawidth. They only added the FIFO on the H7 and N6.

iabdalkader: The F7 has a FIFO, also the DMA is what does packing/unpacking. Please test it.

iabdalkader: Note the FIFO is not even used for SPI, see my comment below.

kwagyeman: Tested LED Shield Triple Buffer output. It's broken (no changes to the PR). Master works.

> The F7 has a FIFO, also the DMA is what does packing/unpacking. 

Yes, but, it's not going to do that unless you tell it the destination is a word or byte. For the H7/N6, the SPI Peripheral itself has a FIFO too, so you can write 32 bits into it and it will unpack to the data size.

kwagyeman: Testing DMA_PDATAALIGN_HALFWORD now.

iabdalkader: > Tested LED Shield Triple Buffer output. It's broken. Master works.

Is that on the F4 or F7?

> For the H7/N6, the SPI Peripheral itself has a FIFO too, so you can write 32 bits into it and it will unpack to the data size.

The F7's SPI also has a FIFO:

image

kwagyeman: No changes to the PR for the F7 doesn't work. If I change to DMA_PDATAALIGN_HALFWORD that works for the F7.

iabdalkader: >  No changes to the PR for the F7 doesn't work. 

What does that mean?

> If I change to DMA_PDATAALIGN_HALFWORD that works for the F7.

Does byte work for F4 then? I'll just hard code half-word for F7, byte for F4.

ports/stm32/omv_csi.c
iabdalkader: I thought we were using circular, I guess it gets set to circular later (or not).. Will use `DMA_NORMAL`.

iabdalkader: Actually, with `circular=False`, the DMA will be initialized immediately for the N6 it needs further initialization for circular mode. I tested `mode = DMA_CIRCULAR` and it seems to work fine for H7, or am I missing something?

kwagyeman: I think the circular arg should be true for the n6 and false for older boards. However, if circular works fine with the H7 plus I can test on the F7 and F4 and see if it's okay. Was just pointing out the mode change from previous code.

iabdalkader: It does work, I think maybe it gets changed later in the driver. The problem that `circular` arg is how I can tell if N6 DMA is done or will need further initialization, without adding yet another arg that gets a value based on ifdef's.

kwagyeman: Yeah, tested F4 and F7 and they work fine. So, I guess it's fine to leave as is. Doesn't seem to hurt anything.

ports/stm32/stm_dma.c
iabdalkader: Will add `psize` and `msize` args here too to avoid any issues with older MCUs.

iabdalkader: Fixed. Source and destination sizes are now passed in bytes, and then mapped to the right constant. This saves a lot of ifdef's outside of `stm_dma.c`

iabdalkader: Please test SPI/DCMI again on old/new boards.

iabdalkader:
Refactored `py_audio.c` too. Tested MDF (N6), DFSDM (Nicla) and SAI (Portenta) all works.

kwagyeman:
Reviewing now.

kwagyeman:
The code inspection review looks good. Testing everything now...

kwagyeman:
M4, M7, H7, H7P are all good. Nothing is enabled for the N6 and a lot of the defines need to be added. Will have to enable and verify later.

iabdalkader:
> Nothing is enabled for the N6 and a lot of the defines need to be added

What do you mean? It already uses that code for FLIR/SPI and MDF, and I tested both.

kwagyeman:
Was talking about the SPI LCD and TV Shield for the N6. Will send a PR to enable those later.

PR #2751, opened 2025-07-05
misc: Update GCC toolchain.

iabdalkader:


iabdalkader:
FYI @kwagyeman. It seems to have added a lot of text size for some reason 🤷🏼‍♂️

PR #2752, opened 2025-07-05
drivers/sensors: Remove OV5640 REV_Y special case.

kwagyeman:
Tested on a Rev-Y board 400 MHz H7. It's not needed anymore. System works fine without it using OV5640.

I think this was added initially because the set_clk_freq code used to pick a higher freq than requested if it couldn't meet the target. However, after that bug was discovered and fixed this check is not necessary anymore.

### DIFF
--- a/boards/ARDUINO_PORTENTA_H7/omv_boardconfig.h
+++ b/boards/ARDUINO_PORTENTA_H7/omv_boardconfig.h
@@ -48,10 +48,6 @@
 #define OMV_OV5640_CLK_FREQ                 (12500000)
 #define OMV_OV5640_PLL_CTRL2                (0x7E)
 #define OMV_OV5640_PLL_CTRL3                (0x13)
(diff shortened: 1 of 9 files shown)


iabdalkader:
> I think this was added initially because the set_clk_freq code used to pick a higher freq than requested if it couldn't meet the target

I think it was added to get the highest FPS, but I could be wrong.

kwagyeman:
Yeah, maybe, there were a lot of bugs were we also dealing with at the time with everything and the DMA offload.

iabdalkader:
BTW, if this is ever needed, there's is a more generic way to achieve the same result: just call `get_clk(nominal=false)` somewhere from the OV5640 driver, and if it's not an exact 24MHz, set to 25MHz. This will increase the clock on every rev and on N6 too.  We'll probably need a post_init op for this.

iabdalkader (approved):

PR #2755, opened 2025-07-06
lib/imlib: Fix update_jpeg_buffer to update from the passed image.

kwagyeman:
* Removes the disable_fb call from the API. This is necessary to get rid of this global frame buffer disable call.
* Updates snapshot in the old csi module to support the update flag and switches scripts from disable_fb to that.
* Updates framebuffer_update_jpeg_buffer to accept an image and updates all the places where it's called to either use the frame buffer or the image object you actually wanted sent to the jpeg buffer.
* Finally, removes the old fallback image transfer code to the IDE which was very slow and doesn't really work anymore.

Here's what happens if you try to capture HD JPEG images on the H7:

https://github.com/user-attachments/assets/ccee19cd-9218-4d03-b138-b9f0ecd22afd


lib/imlib/framebuffer.c
```diff
@@ -95,34 +95,27 @@ static void jpegbuffer_init_from_image(image_t *img) {
...
 
                 mutex_unlock(&jpegbuffer->lock, MUTEX_TID_OMV);
             }
-
-            if (does_not_fit) {
-                printf("Warning: JPEG/PNG too big! Trying framebuffer transfer using fallback method!\n");
-                int new_size = framebuffer_encoded_size(src);
-                fb_alloc_mark();
-                uint8_t *temp = fb_alloc(new_size, FB_ALLOC_NO_HINT);
-                framebuffer_encode(temp, src);
```
iabdalkader: But why leave these functions and this functionality at all?

> encode_for_ide if True the image is encoded in a way that the IDE can display it if printed by doing print(image). This is useful for debugging purposes over UARTs via Open Terminal in the IDE.

I'd just drop the whole thing.

modules/py_csi.c
```diff
@@ -142,9 +142,29 @@ static mp_obj_t py_omv_csi_flush() {
 }
 static MP_DEFINE_CONST_FUN_OBJ_0(py_omv_csi_flush_obj, py_omv_csi_flush);
 
-static mp_obj_t py_omv_csi_snapshot(size_t n_args, const mp_obj_t *args, mp_map_t *kw_args) {
+static mp_obj_t py_omv_csi_snapshot(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
```
iabdalkader: I don't think we should update/backport fixes to the obsolete module, it should be frozen to make it obsolete in time. Maybe those examples that need `update=False` should use the new API instead? Eventually we'll have to move to the new API and remove the old one.

lib/imlib/framebuffer.c
```diff
@@ -95,34 +95,27 @@ static void jpegbuffer_init_from_image(image_t *img) {
 }
 
 void framebuffer_update_jpeg_buffer(image_t *src) {
-    static int overflow_count = 0;
+    static int overflow_count = 0, copy_overflow_count = 0;
```
iabdalkader: The IDE can count that.

modules/py_csi.c
kwagyeman: Done

lib/imlib/framebuffer.c
kwagyeman: Removed. Now that img.flush() actually flushes the image you want to and that you can stop snapshot() from updating the main FB it can be removed.

lib/imlib/framebuffer.c
kwagyeman: Changed it to a simpler system on the camera.

kwagyeman:
https://github.com/user-attachments/assets/b8d0a658-7aed-4230-93f8-7f15df1d1521

Okay, the error message is now internally displayed from the IDE. This means that it is part of IDE translation. Also, the blinking is done from a timer so it's flashing is consistent no matter the message rate. Finally, because only an escape code is sent it will never leak as text into the terminal on script stop.

With the current IDE nothing will be displayed. But, on the next IDE release you'll see the message.

iabdalkader:
The error message handling is much better. This is good to go, just one comment please use `csi0` instead of `sensor`.

kwagyeman:
PR Updated with changes requested.

PR #2758, opened 2025-07-07
modules/py_spi_display: Add vflip/hmirror support.

kwagyeman:
Ads support for a new 320x240 LCD Shield. This new LCD shield has a multi-touch touch controller onboard. Will send a PR to enable it via a python module later. It also has the tear pin hooked up to the CPU so it can be perfectly synced. Will add support for that later too.

Just sending this PR now to get it posted. Also, cleaned up the LCD shield example. The notes about the PWM/DAC controller don't apply to the RT1062 and N6, but, are also rarely used.

https://github.com/user-attachments/assets/bb983188-702f-46be-ab9b-d92e279a769b


iabdalkader:

https://forums.openmv.io/t/openmv-h7-plus-with-new-fw-is-now-seen-as-h7/11199/13

scripts/examples/50-OpenMV-Boards/60-Shields/60-LCD-Shield/lcd_shield.py
```diff
@@ -0,0 +1,26 @@
...
+import sensor
+import time
+import display
+import image
+
+sensor.reset()
+sensor.set_pixformat(sensor.RGB565)
+sensor.set_framesize(sensor.LCD)
+
+lcd = display.SPIDisplay(vflip=True, hmirror=True)
```
iabdalkader, line 29: Please revert the notes about the backlight controller.  Also, would `portrait` arg work here?

iabdalkader (changes requested):

scripts/examples/50-OpenMV-Boards/60-Shields/60-LCD-Shield/lcd_shield.py
kwagyeman: Reverted, regarding portrait, no, that's a rotate 90, so it would be like transpose + hmirror at the same time. vflip+hmirror is a rotation of 180.

iabdalkader (approved):

PR #2766, opened 2025-07-13
imlib: Rework frame buffer management.

iabdalkader:
The main motivation for this work is to fix the FB double buffer issue, and possible race condition in triple buffer mode. However, as an added bonus, this also offers some optimization, code safety, new features and major cleanups.

---

* The frame buffer now uses two queues to manage the buffers: a free queue and a read queue. It juggles buffers between queues based on the operation and flags. The frame buffer code has been simplified a lot, since much of the complexity has been pushed out to the new queue component. It is also much more flexible now. For example, we could add a third queue if needed, or check the queue sizes and so on.

* The new SPSC lock-free queue component uses standard atomics and acquire and release semantics to ensure there are no issues with reordering by either the CPU or the compiler on current or future platforms. It is also generic enough to be used by other code and flexible enough to be allocated either dynamically using malloc or from a preallocated memory chunk.

* The frame buffer and JPEG buffer headers have been decoupled from their raw buffer memory and placed in SRAM. Additionally, when using three buffers or fewer, the new frame buffer queues are also placed in SRAM. Before this change, accessing fb->head, fb->tail, fb->buf_size, fb->w, etc. would all trigger DRAM accesses unless cached, but even then they could be evicted and accessed from DRAM again. Hopefully, this improves DRAM bandwidth usage, even on older cameras. I don’t know why I didn’t fix this earlier, I think it started with legacy cameras that lacked DRAM and just stayed that way. 

* 1, 2, 3 and N buffers all work, see tests below. Also, the FPS of the 2 buffers is now the same as the 3 buffers, just as expected. Furthermore, buffer expansion is supported on static and dynamic frame buffers. It's also possible to expand a single buffer or multiple buffers to use all the remaining FB memory (minus any fb_alloc'd memory, in the case of static frame buffers).

---

### Tests:

* Tested single, dual, triple CSIs with FLIR/VOSPI.
* Tested buffer expansion on static/dynamic FB with/without fb_alloc.
* Tested copy_to_fb by loading images.
* Tested on N6, 4P, RT1060.

### Frame buffer usage tests:

The following tests cover all cases. There's a 500ms delay after snapshot to see if the fb/queues behave as expected. For single and FIFO the capture stops. For 2 buffers, the producer holds on to the second buffer until the first one is released. Finally, for triple buffer mode, the producer alternates between the two free buffers.

```
// One buffer, stops.
snapshot buffer 0x91900020
next buffer 0x0

// 2 buffers:
snapshot buffer 0x91925840
next buffer 0x91900000

// 3 buffers:
snapshot buffer 0x9194b060
next buffer 0x91900000
next buffer 0x91925820
next buffer 0x91900000
next buffer 0x91925820

// > 3 buffers:
snapshot buffer 0x91900060
next buffer 0x91925860
next buffer 0x9194b080
next buffer 0x919708a0
next buffer 0x0
```

### DIFF
--- a/common/omv_csi.c
+++ b/common/omv_csi.c
@@ -267,6 +267,13 @@ __weak int omv_csi_abort(omv_csi_t *csi, bool fifo_flush, bool in_irq) {
             framebuffer_flush_buffers(csi->fb, false);
         }
     }
+
+    #if defined(OMV_CSI_FSYNC_PIN)
+    if (csi->frame_sync) {
+        omv_gpio_write(OMV_CSI_FSYNC_PIN, 0);
+    }
+    #endif
+
     return 0;
 }
 
@@ -1526,11 +1533,61 @@ __weak int omv_csi_copy_line(omv_csi_t *csi, void *dma, uint8_t *src, uint8_t *d
 }
 
 __weak int omv_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
-    int ret = OMV_CSI_ERROR_CTL_UNSUPPORTED;
+    vbuffer_t *buffer = NULL;
+
+    if (!csi->snapshot) {
+        return OMV_CSI_ERROR_CTL_UNSUPPORTED;
+    }
+
+    if (csi->pixformat == PIXFORMAT_INVALID) {
+        return OMV_CSI_ERROR_INVALID_PIXFORMAT;
+    }
+
+    if (csi->framesize == OMV_CSI_FRAMESIZE_INVALID) {
+        return OMV_CSI_ERROR_INVALID_FRAMESIZE;
+    }
+
+    if (omv_csi_check_framebuffer_size(csi) == -1) {
+        return OMV_CSI_ERROR_FRAMEBUFFER_OVERFLOW;
+    }
+
+    buffer = framebuffer_acquire(csi->fb, FB_FLAG_USED | FB_FLAG_PEEK);
+
+    // Compress the previous framebuffer for the IDE preview and release it.
+    // Note: We must check if the buffer has been used before releasing it,
+    // as it might have been captured in non-blocking mode but not used yet.
+    if (buffer && (buffer->flags & VB_FLAG_USED)) {
+        if (flags & OMV_CSI_CAPTURE_FLAGS_UPDATE) {
+            image_t tmp;
+            framebuffer_init_image(csi->fb, &tmp);
+            framebuffer_update_jpeg_buffer(&tmp);
+        }
+
+        // Release the previous buffer from used queue -> free queue.
+        framebuffer_release(csi->fb, FB_FLAG_USED | FB_FLAG_INVALIDATE);
+    }
+
+    // Toggle FSYNC.
+    #if defined(OMV_CSI_FSYNC_PIN)
+    if (csi->frame_sync) {
+        omv_gpio_write(OMV_CSI_FSYNC_PIN, 1);
+    }
+    #endif
 
     // Call the sensor specific function.
-    if (csi->snapshot) {
-        ret = csi->snapshot(csi, image, flags);
+    int ret = csi->snapshot(csi, image, flags);
+
+    // Toggle FSYNC.
+    #if defined(OMV_CSI_FSYNC_PIN)
+    if (csi->frame_sync) {
+        omv_gpio_write(OMV_CSI_FSYNC_PIN, 0);
+    }
+    #endif
+
+    if (ret == 0) {
+        // Mark this buffer to be released on the next call.
+        buffer = framebuffer_acquire(csi->fb, FB_FLAG_USED | FB_FLAG_PEEK);
+        buffer->flags |= VB_FLAG_USED;
     }
 
     return ret;
--- a/common/omv_csi.c
+++ b/common/omv_csi.c
@@ -260,12 +260,8 @@ __weak int omv_csi_abort(omv_csi_t *csi, bool fifo_flush, bool in_irq) {
     csi->last_frame_ms = 0;
     csi->last_frame_ms_valid = false;
 
-    if (csi->fb) {
-        if (fifo_flush) {
-            framebuffer_flush_buffers(csi->fb, true);
-        } else if (!csi->disable_full_flush) {
-            framebuffer_flush_buffers(csi->fb, false);
-        }
+    if (csi->fb && fifo_flush && !csi->disable_full_flush) {
+        framebuffer_flush(csi->fb);
     }
 
     #if defined(OMV_CSI_FSYNC_PIN)
@@ -353,7 +349,7 @@ __weak int omv_csi_reset(omv_csi_t *csi, bool hard) {
     }
 
     // Reset framebuffers
-    framebuffer_flush_buffers(csi->fb, true);
+    framebuffer_flush(csi->fb);
     return 0;
 }
 
@@ -774,7 +770,7 @@ __weak int omv_csi_set_pixformat(omv_csi_t *csi, pixformat_t pixformat) {
     csi->fb->pixfmt = PIXFORMAT_INVALID;
 
     // Auto-adjust the number of frame buffers.
-    omv_csi_set_framebuffers(csi, -1);
+    omv_csi_set_framebuffers(csi, -1, false);
 
     // Reconfigure the hardware if needed.
     return omv_csi_config(csi, OMV_CSI_CONFIG_PIXFORMAT);
@@ -818,7 +814,7 @@ __weak int omv_csi_set_framesize(omv_csi_t *csi, omv_csi_framesize_t framesize) 
     csi->fb->pixfmt = PIXFORMAT_INVALID;
 
     // Auto-adjust the number of frame buffers.
-    omv_csi_set_framebuffers(csi, -1);
+    omv_csi_set_framebuffers(csi, -1, false);
 
     // Reconfigure the hardware if needed.
     return omv_csi_config(csi, OMV_CSI_CONFIG_FRAMESIZE);
@@ -932,7 +928,7 @@ __weak int omv_csi_set_windowing(omv_csi_t *csi, int x, int y, int w, int h) {
     csi->fb->pixfmt = PIXFORMAT_INVALID;
 
     // Auto-adjust the number of frame buffers.
-    omv_csi_set_framebuffers(csi, -1);
+    omv_csi_set_framebuffers(csi, -1, false);
 
     // Reconfigure the hardware if needed.
     return omv_csi_config(csi, OMV_CSI_CONFIG_WINDOWING);
@@ -1256,7 +1252,7 @@ __weak bool omv_csi_get_auto_rotation(omv_csi_t *csi) {
     return csi->auto_rotation;
 }
 
-__weak int omv_csi_set_framebuffers(omv_csi_t *csi, int count) {
+__weak int omv_csi_set_framebuffers(omv_csi_t *csi, size_t count, bool expand) {
     // Disable any ongoing frame capture.
     omv_csi_abort(csi, true, false);
 
@@ -1275,7 +1271,17 @@ __weak int omv_csi_set_framebuffers(omv_csi_t *csi, int count) {
     // Otherwise, use the real frame size.
     csi->fb->frame_size = resolution[csi->framesize][0] * resolution[csi->framesize][1] * 2;
     #endif
-    return framebuffer_set_buffers(csi->fb, count);
+
+    if (count == -1) {
+        for (size_t i=3; i>0; i--) {
+            if (!framebuffer_resize(csi->fb, i, expand)) {
+                return 0;
+            }
+        }
+        return -1;
+    }
+
+    return framebuffer_resize(csi->fb, count, expand);
 }
 
 __weak int omv_csi_set_special_effect(omv_csi_t *csi, omv_csi_sde_t sde) {
@@ -1443,7 +1449,7 @@ __weak int omv_csi_auto_crop_framebuffer(omv_csi_t *csi) {
     }
 
     // Auto-adjust the number of frame buffers.
-    omv_csi_set_framebuffers(csi, -1);
+    omv_csi_set_framebuffers(csi, -1, false);
     return 0;
 }
 
--- a/common/queue.c
+++ b/common/queue.c
@@ -1,0 +1,190 @@
+/*
+ * SPDX-License-Identifier: MIT
+ *
+ * Copyright (C) 2013-2024 OpenMV, LLC.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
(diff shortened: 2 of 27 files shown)


iabdalkader:
While testing this I also found a subtle race condition in the current framebuffer code. It can easily be reproduced under the following conditions:

* N6, PAG7936, 240FPS, 3 buffers.
* A strategically placed delay in the consumer's path which allows the IRQ to run one or more times.

With that, the DMA writes to the buffer currently held by snapshot.

```diff
diff --git a/lib/imlib/framebuffer.c b/lib/imlib/framebuffer.c
index 5f932a59b..932b75f9e 100644
--- a/lib/imlib/framebuffer.c
+++ b/lib/imlib/framebuffer.c
@@ -24,6 +24,7 @@
  * Framebuffer functions.
  */
 #include <stdio.h>
+#include "py/mphal.h"
 #include "mpprint.h"
 #include "fmath.h"
 #include "framebuffer.h"
@@ -334,6 +335,7 @@ void framebuffer_setup_buffers(framebuffer_t *fb) {
 }
 
 vbuffer_t *framebuffer_get_head(framebuffer_t *fb, framebuffer_flags_t flags) {
+    static uint32_t time_ms = 0;
     int32_t new_head = (fb->head + 1) % fb->n_buffers;
 
     // Single Buffer Mode.
@@ -354,6 +356,7 @@ vbuffer_t *framebuffer_get_head(framebuffer_t *fb, framebuffer_flags_t flags) {
         } else {
             new_head = sampled_tail;
         }
+        mp_hal_delay_ms(time_ms++ % 10);
         // Video FIFO Mode.
     } else {
         if (fb->head == fb->tail) {

diff --git a/ports/stm32/omv_csi.c b/ports/stm32/omv_csi.c
index 2ea7aca96..907cf7184 100644
--- a/ports/stm32/omv_csi.c
+++ b/ports/stm32/omv_csi.c
@@ -368,6 +368,8 @@ static uint32_t get_dcmi_hw_crop(omv_csi_t *csi, uint32_t bytes_per_pixel) {
     return 0;
 }
 
+static volatile vbuffer_t *dma_buffer=NULL;
+
 static void stm_csi_frame_event(omv_csi_t *csi, uint32_t pipe) {
     framebuffer_t *fb = csi->fb;
 
@@ -390,6 +392,7 @@ static void stm_csi_frame_event(omv_csi_t *csi, uint32_t pipe) {
         return;
     }
 
+    dma_buffer=0;
     framebuffer_get_tail(fb, FB_NO_FLAGS);
 
     if (csi->frame_cb.fun) {
@@ -399,6 +402,8 @@ static void stm_csi_frame_event(omv_csi_t *csi, uint32_t pipe) {
     #if defined(STM32N6)
     // Get the destination buffer address.
     vbuffer_t *buffer = framebuffer_get_tail(fb, FB_PEEK);
+    dma_buffer = buffer;
+
     if (buffer == NULL) {   // FIFO is FULL
         omv_csi_abort(csi, false, false);
     } else if (csi->mipi_if) {
@@ -693,6 +698,10 @@ static int stm_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
         }
     }
 
+    if (buffer == dma_buffer) {
+        printf("snapshot buffer: 0x%p dma buffer: 0x%p\n", buffer, dma_buffer);
+    }
+
     // In JPEG 3 mode, the transfer must be aborted as it waits for data indefinitely.
     if (!csi->mipi_if && (csi->pixformat == PIXFORMAT_JPEG) && (csi->jpg_format == 3)) {
         omv_csi_abort(csi, true, false);
```

Test script: 

```Python
import time
import csi

print(["0x%x"%i for i in csi.devices()])

csi0 = csi.CSI()
csi0.reset(hard=True)
csi0.pixformat(csi.RGB565)
csi0.framesize(csi.VGA)
csi0.framerate(240)
csi0.framebuffers(3)

clock = time.clock()
while True:
    clock.tick()
    img0 = csi0.snapshot()
    print(clock.fps())
```


Output:
```
158.8785
snapshot buffer: 0x919fa0a0 dma buffer: 0x919fa0a0
158.9327
158.8032
snapshot buffer: 0x9197d080 dma buffer: 0x9197d080
157.7753
snapshot buffer: 0x919fa0a0 dma buffer: 0x919fa0a0
156.5996
156.6667
157.2536
157.3157
157.5492
157.7802
158.0087
snapshot buffer: 0x91900060 dma buffer: 0x91900060
```

lib/imlib/framebuffer.c
```diff
@@ -84,364 +83,276 @@ void framebuffer_init_from_image(framebuffer_t *fb, image_t *img) {
...
+            // if the queue size is 1.
+            size_t old_tail = queue_prev_tail(fb->used_queue);
+            vbuffer_t *old_buffer = fb->used_queue->items[old_tail];
+
+            size_t new_head = fb->free_queue->head;
+            vbuffer_t *new_buffer = fb->free_queue->items[new_head];
+
+            memset(old_buffer, 0, sizeof(vbuffer_t));
+            fb->used_queue->items[old_tail] = new_buffer;
+            fb->free_queue->items[new_head] = old_buffer;
```
kwagyeman: I stared at this for a few hours, thought it was bugged, then realized my head/tail terminology was swapped. After that, it makes 100% complete sense. I thought about poping from the used queue and putting it back on the free queue, but, that's not thread safe since you are modifying the consumer vars. You'd have to add locks to make it work. Swapping is the only way to do this cleanly with the used/free queue structure. The above is the way to solve this.

common/queue.c
```diff
@@ -0,0 +1,208 @@
...
+    void *item = q->items[new_tail];
+
+    if (!peek) {
+        // Release ensures the read completes before moving tail
+        atomic_store_explicit(&q->tail, new_tail, memory_order_release);
+    }
+    #endif
+    
+    return item;
+}
```
kwagyeman: This method isn't thread safe. I would suggest removing it. SPSC fifos can't really index backwards. Especially given the extra space in the FIFO to tell between empty and full.

common/vospi.c
```diff
@@ -137,37 +123,37 @@ void vospi_callback(omv_spi_t *spi, void *userdata, void *buf) {
...
-                if (vospi.sid == VOSPI_SIDS_PER_FRAME) {
-                    vospi.sid = 0;
-                    framebuffer_get_tail(vospi.fb, FB_NO_FLAGS);
-                }
-                // For the FLIR Lepton 1/2 we just have to receive all the pids.
-            } else {
-                framebuffer_get_tail(vospi.fb, FB_NO_FLAGS);
+    vbuffer_t *buffer = framebuffer_acquire(vospi.fb, FB_FLAG_FREE | FB_FLAG_PEEK, 0);
+
+    if (!buffer) {
```
kwagyeman, line 134: Suggest leaving vospi running versus stopping it on the frame drop. There's a large sync cost to bringing the bus up again. Just turning the capture flag off and then turning it on again is cheaper.

drivers/sensors/lepton.c
```diff
@@ -397,66 +397,75 @@ static int config(omv_csi_t *csi, omv_csi_config_t config) {
...
+        framebuffer_update_jpeg_buffer(&tmp);
+    }
+
+    size_t reset_retry = 0;
+    vbuffer_t *buffer = framebuffer_acquire(csi->fb, FB_FLAG_USED | FB_FLAG_PEEK, 0);
+
+    // Because this snapshot supports non-blocking mode, the buffer can't
+    // be released unconditionally, as it might have been captured in the
+    // background but not used yet..
+    if (buffer && (buffer->flags & VB_FLAG_USED)) {
```
kwagyeman: Can you make this standard for all drivers? Like in common? This way all drivers support nonblocking snapshot.

Also, this was tripping me up before! Great find!!! I was so confused why a blocking call was required first.

drivers/sensors/lepton.c
```diff
@@ -397,66 +397,75 @@ static int config(omv_csi_t *csi, omv_csi_config_t config) {
...
         return OMV_CSI_ERROR_INVALID_FRAMESIZE;
     }
 
     if (omv_csi_check_framebuffer_size(csi) == -1) {
         return OMV_CSI_ERROR_FRAMEBUFFER_OVERFLOW;
     }
 
-    for (int i = 0; i < LEPTON_SNAPSHOT_RETRY; i++) {
-        if (vospi_snapshot(LEPTON_SNAPSHOT_TIMEOUT) == 0) {
+    if (flags & OMV_CSI_CAPTURE_FLAGS_UPDATE) {
```
kwagyeman: Some of the if statements and the update flag here could be moved to common to become standard for all drivers.

ports/stm32/omv_csi.c
```diff
@@ -378,28 +378,27 @@ static void stm_csi_frame_event(omv_csi_t *csi, uint32_t pipe) {
     HAL_NVIC_EnableIRQ(csi->dma_irqn);
     #endif
 
-    // Reset DCMI_DMAConvCpltUser frame drop state.
     csi->first_line = false;
     if (csi->drop_frame) {
-        csi->drop_frame = false;
```
kwagyeman, line 384: I think you need to clear this flag or the frame drops won't stop.

common/omv_csi.c
```diff
@@ -260,12 +260,8 @@ __weak int omv_csi_abort(omv_csi_t *csi, bool fifo_flush, bool in_irq) {
     csi->last_frame_ms = 0;
     csi->last_frame_ms_valid = false;
 
-    if (csi->fb) {
-        if (fifo_flush) {
-            framebuffer_flush_buffers(csi->fb, true);
-        } else if (!csi->disable_full_flush) {
-            framebuffer_flush_buffers(csi->fb, false);
-        }
```
kwagyeman, line 268: What was going on here previously is:

fifo_flush==True

Flush the frame buffer and don't leave any valid frames to get via the get_fb() call.

fifo_flush==False && !csi->disable_full_flush

Flush the frame buffer, but, leave the current frame valid to get via the get_fb() call.

...

Given a way to grab the current frame was removed from the csi API, is the intention here that any get_fb() like functionality is to be removed?

Even so, csi->disable_full_flush is still valuable to prevent the frame buffer fifo from being emptied on abort. Suggest:

```
if (csi->fb && fifo_flush && (!csi->disable_full_flush)) {
    framebuffer_flush(csi->fb);
}
```

common/queue.c
iabdalkader: Yes I'll remove this, but it does handle the empty slot: If tail is at the empty slot, `tail-1` is the last valid slot, and if tail==0, use capacity (which is valid).

common/omv_csi.c
iabdalkader: > Flush the frame buffer, but, leave the current frame valid to get via the get_fb() call.

That's what's confusing about this, I don't know how to implement that, given the current implementation. You could still grab the current frame with `acquire(HEAD|PEEK)` but there has to be a valid frame first to grab. I can certainly add the flag to the check as you suggested. If there's something else that could be done, let me know.

drivers/sensors/lepton.c
iabdalkader: I could refactor it in another PR. I also realized that jpeg_update needs to be inside that if statement, otherwise it will update the frame before drawing on it. While at it, I'd also like to drop `omv_csi_auto_crop_framebuffer`, you could simply just use a smaller res, set a window, change format explicitly.

common/vospi.c
iabdalkader: Won't it go out of sync if you turn of the capture flag? It will stop counting pid's/sid's.

common/queue.c
kwagyeman: Yes, queue_is_empty() prevents you from looking at the empty space. :)

common/vospi.c
kwagyeman: No, because it just discards packets until it receives the first in a stream: https://github.com/openmv/openmv/blob/master/common/vospi.c#L123

It only aborts if it's synced to a stream and then the packet IDs don't match the sequence it's been receiving. But, after receiving a frame the pids will have been reset back to 0.

common/omv_csi.c
kwagyeman: I think `acquire(HEAD|PEEK)` is fine to grab the frame, if this returns NULL then you should return None for the FB image object right? I think that makes sense.

drivers/sensors/lepton.c
iabdalkader: I made it so, they will all support non-blocking snapshots now, even though I don't see any benefit from this for standard sensors, but it removes a lot of duplicate code. However, now I'm not sure if FSYNC pin makes any sense, especially in non-blocking mode.

kwagyeman: I'd return an error if `csi->frame_sync` and the non-blocking flag is set. It still makes sense in blocking mode as frame generation is synced to snapshot.

iabdalkader: Seems the only sensor that sets that is `pag7920`. Should we even support that?

iabdalkader: Note that even in blocking mode, the FSYNC pin toggles at the wrong time. Typically, the DMA starts first, then the pin toggles, often mid-frame, and after that, there's no correlation between FSYNC and when the frame is captured.

lib/imlib/framebuffer.c
iabdalkader: > thought about poping from the used queue and putting it back on the free queue, but, that's not thread safe since you are modifying the consumer vars

If you actually try this and run it long enough, the queues eventually fail and DMA overwrites the consumer buffer. I know doing so breaks SPSC semantics, however, I couldn't trace the steps that lead into that state.

```C
// Triple buffer: Swap the old buffer with the latest.
if ((buffer = queue_pop_back(fb->used_queue, false))) {
    memset(buffer, 0, sizeof(vbuffer_t));
    queue_push(fb->free_queue, buffer);
}
```

common/omv_csi.c
iabdalkader: > I think acquire(HEAD|PEEK) is fine to grab the frame, if this returns NULL then you should return None for the FB image object right? I think that makes sense.

Calling `acquire(FREE|PEEK)` from non-snapshot code, should never fail especially if it follows a `resize(1)`. I don't know/remember every use case, you should test these corner cases and let me know if something needs to be fixed.

drivers/sensors/lepton.c
kwagyeman: The camera doesn't output data until FSYNC happens, so, it's not really mid frame as no frame occurs until the toggle.

fsync is how we trigger the global shutter sensor in triggered mode: https://github.com/openmv/openmv/blob/master/drivers/sensors/mt9v0xx.c#L535
https://github.com/openmv/openmv/blob/master/drivers/sensors/mt9v0xx.c#L487

I've never used it for the pag7920. That would be something from pixart.

common/omv_csi.c
kwagyeman: Okay, will do.

lib/imlib/framebuffer.c
kwagyeman: Yeah, definately unsafe. Anyway, great solution!!!

drivers/sensors/lepton.c
iabdalkader: If you use it with non-blocking mode, it doesn't make any sense. Anyway, I moved FSYNC control to common code.

common/vospi.c
iabdalkader: > There's a large sync cost to bringing the bus up again.

Okay I've made it not abort just clear the capture flag, on no buffers, but there wasn't any huge sync delay, it's just 200ms, and it's probably not needed.

ports/stm32/omv_csi.c
iabdalkader: Reverted.

iabdalkader:
All ports supported.

kwagyeman:
Okay, will have the PR tested by the EoD.

kwagyeman:
Wasn't able to get testing done today. Will get it done tomorrow.

lib/imlib/framebuffer.c
iabdalkader: Okay keeping this for now, but I moved it to a function in queue.c.

kwagyeman:
Testing results:

OpenMV2:
* CSI 1, 2, 3 buffers all good.

OpenMV3:
* CSI 1, 2, 3 buffers all good.

OpenMV4:
* CSI 1, 2, 3 buffers all good.
* FLIR 1, 2, 3 buffers all good.

OpenMV4P:
* CSI 1, 2, 3 buffers all good.
* FLIR 1, 2, 3 buffers all good.

OpenMV RT1062:
* CSI 1, 2, 3 buffers all good.
* FLIR 1, 2, 3 buffers all good.

OpenMV N6:
* CSI 1, 2, 3 buffers all good.
* FLIR 1, 2, 3 buffers all good.
* Dual CSI + Lepton all good.

OpenMV AE3:
* CSI 1, 2, 3 buffers all good.

Everything appears to be working. Let me know if there are any specific things to check.

common/omv_csi.c
kwagyeman: Not seeing any issues. Things are working as expected.

iabdalkader:
> Everything appears to be working. Let me know if there are any specific things to check.

I'd test code that uses fb_alloc, and code that loads images to FB (copy_to_fb, py_helper stuff), otherwise it's pretty much done.

kwagyeman:
Okay, I'll do some more testing later today.

kwagyeman:
Copy_to_fb, heavy fb_alloc() usage in apriltags, qrcodes, and barcodes, and etc. is all working. ImageReading playback and etc. is all operational too. LGTM!

Issue #2724, opened 2025-06-23
Refactor py_fir_lepton to dual csi system


kwagyeman:
Once dual csi is merged remove the older module and update all examples.

PR #2771, opened 2025-07-21
modules/py_fir_lepton: Remove lepton control through fir module.

kwagyeman:
Removes the old FIR module and enables the Pure Thermal to use the new csi interface with the Lepton as a secondary sensor.

Fixes: https://github.com/openmv/openmv/issues/2724


ports/stm32/stm_pwm.c
```diff
@@ -82,7 +82,7 @@ int stm_pwm_start(TIM_HandleTypeDef *tim, TIM_TypeDef *inst, uint32_t channel, u
 
         if (HAL_TIM_PWM_Init(tim) != HAL_OK ||
             HAL_TIM_PWM_ConfigChannel(tim, &TIMOCHandle, channel) != HAL_OK ||
-            HAL_TIM_PWM_Start(tim, channel != HAL_OK)) {
+            HAL_TIM_PWM_Start(tim, channel) != HAL_OK) {
```
iabdalkader: Whoa! Good find!

iabdalkader:
This is all good, except for [ports/stm32: Add support for aux pwdn/rst/clk pins.](https://github.com/openmv/openmv/pull/2771/commits/2208f8c61120d6556c834a233bbe3348e249ce93). You're going to have to think of clean way to do that, otherwise we're better off keeping `py_fir_lepton`.  Here's my suggestion:

- Configure those aux pins in PT's `board_init.c` in MicroPython enough to make the module detectable in probe/scan on boot.

If you never need to change the pins again, we're done, if you do:
- Add power and reset `omv_gpio_t` pointers to `omv_csi_t`.
- In `omv_csi_init`, set the board defined pins (default) for all CSIs initially, then in `lepton_init` override them.

Note I cherry-picked the PWM fix and merged it.

kwagyeman:
I tried setting the pins statically already, the issue is a bit more complex than that. So, if I set the default pin states for the lepton such that omv_csi_probe() finds the LEPTON on the first pass of polarity_configs then you get get a single FLIR Lepton detected as the attached sensor. The OV5640+LEPTON is found on the second pass through polarity_configs `{ OMV_CSI_ACTIVE_HIGH, OMV_CSI_ACTIVE_LOW },`. So, to make it work I need to mirror the rst/pwdn pin states to the aux pins.

Adding a mirror mode functionality to the omv_gpio pins would be one solution, but, maybe it would be easier just to modify the polarity_configs with an ifdef to have `{ OMV_CSI_ACTIVE_HIGH, OMV_CSI_ACTIVE_LOW },` first for just for the Pure Thermal? Then I can statically init the lepton pins and the changes are minimal. Thoughts?

iabdalkader:
> maybe it would be easier just to modify the polarity_configs with an ifdef to have `{ OMV_CSI_ACTIVE_HIGH, OMV_CSI_ACTIVE_LOW },` first for just for the Pure Thermal?

Yes, that's much better, other cams can use this as well (maybe AE3). Why can't this be the default as I had it before?

kwagyeman:
Random cameras stop working with the order you had previously. I'll do an audit of various orders to determine which work. Having a controllable order for specific boards sounds good otherwise.

kwagyeman:
Updated, now depends on https://github.com/openmv/micropython/pull/135. Will do the polarity audit tomorrow.

scripts/examples/50-OpenMV-Boards/51-Pure-Thermal/00-HelloWorld/pure_thermal.py
```diff
@@ -4,36 +4,65 @@
...
+    % (
+        csi1.ioctl(csi.IOCTL_LEPTON_GET_WIDTH),
+        csi1.ioctl(csi.IOCTL_LEPTON_GET_HEIGHT),
+    )
+)
+
+print(
+    "Radiometry Available: "
+    + ("Yes" if csi1.ioctl(csi.IOCTL_LEPTON_GET_RADIOMETRY) else "No")
+)
```
iabdalkader, line 55: About these broken lines, the linter allows for a max line of 120 so you don't need to break those like that I think. Also you could use shorter variables, like `min_temp_c`

iabdalkader:
This has been merged, but note that unlike the dual sensor, this will not reset/power down lepton with the main sensor. To get similar behavior, lepton has to override the pins in `lepton_init`.

scripts/examples/50-OpenMV-Boards/51-Pure-Thermal/00-HelloWorld/pure_thermal.py
kwagyeman: Oh, cool. Let me change my pep8 settings to match this.

kwagyeman:
Yep, for now though, the OEM reset issued by the lepton driver seems to be enough. Will do that audit on polarity states and then test 2766 thoughly on all cameras.

PR #2773, opened 2025-07-21
tflm: Remove out of date tf documentation.

kwagyeman:
Current ML module documentation covers tensorflow support now. And with ROMFS building models into the firmware is not required anymore and can be done via the IDE.

Fixes: https://github.com/openmv/openmv/issues/2763


iabdalkader:
Should we add something on romfs, how to configure it via json file and its options?

kwagyeman:
Sure

PR #2787, opened 2025-07-27
misc: Optimize Lepton delays.

iabdalkader:
* Fix infinite loops in the SDK (just in the command we use).
* Reduce delay to 950ms. This works for 2.5, unsure about Lepton 3, but could make it configurable if it doesn't work.
* Optimize hard-reset/power-on delay. `reset()` delay is calculated based on last hard-reset/power-on timestamp:
```
diff_ms = (ticks_ms() - last_hard_reset_ms)
if diff_ms < MIN_BOOT_DELAY
   delay_ms(MIN_BOOT_DELAY - diff_ms)
```


kwagyeman:
Seems good. Works as expected on all boards.

iabdalkader:
Moving sync issues fix to new PR.

PR #2788, opened 2025-07-28
ports: Add GPU matrix support.

kwagyeman:
This PR adds initial matrix transformation support to draw image which can be used anywhere in the code.

Right now it just supports 2x3 affline matrices and 3x3 perspective transform matrices. However, I plan to expand support in the future for general purpose image remap transformation matrices which can be used for lens correction in the future (the CPU would handle this).

The code is written so that if the GPU can handle the operation it will. Otherwise, things fallback to the CPU based renderer. Note that I did not implement any of the CPU fallback path, so, draw_image will just ignore the `mat` argument if the GPU doesn't support it (and only the N6 has a GPU that can do this right now).

https://github.com/user-attachments/assets/b3e9a9f7-0fee-4401-a9f8-3ac22a19ef98

```
import sensor
import time
import image
from ulab import numpy as np
import math

sensor.reset()
clock = time.clock()

angle_deg = 0  # Start angle

while True:
    clock.tick()

    img = sensor.snapshot()
    small_img = img.scale(x_scale=0.75, y_scale=0.75, hint=image.BILINEAR, copy=True)

    # Rotation center (in dst image space)
    cx = img.width() / 2
    cy = img.height() / 2

    # Update angle
    angle_rad = math.radians(angle_deg)
    angle_deg = (angle_deg + 1) % 360  # increment rotation

    cos_t = math.cos(angle_rad)
    sin_t = math.sin(angle_rad)

    # 2x3 rotation matrix about center of large_img
    tx = cx - cos_t * cx + sin_t * cy
    ty = cy - sin_t * cx - cos_t * cy

    m = np.array([[cos_t, -sin_t, tx],
                  [sin_t,  cos_t, ty],
                  [0,  0, 1]])

    # Center position in destination (main image)
    x = (img.width() // 2) - (small_img.width() // 2)
    y = (img.height() // 2) - (small_img.height() // 2)

    # Draw rotated small_img in-place
    img.draw_image(small_img, x, y, transform=m)

    print(clock.fps())
```

https://github.com/user-attachments/assets/b948fb60-9a71-4678-91b7-dd5ca77a50a1

```
import time
import csi
import image
from ulab import numpy as np
import math

m = np.array([[ 0.8776188251549479, 0.02006060498589142, 38.14425632554175],
              [0.03851867279446007, 0.7883679987364723, 15.810020073591545],
              [0.0004941236255517049, -3.450005798236746e-05, 1.0]])

print(["0x%x"%i for i in csi.devices()])

alpha_pal = image.Image(256, 1, image.GRAYSCALE)
for i in range(256):
    alpha_pal[i] = int(math.pow((i / 255), 2) * 255)

csi0 = csi.CSI()
csi0.reset(hard=True)
csi0.pixformat(csi.RGB565)
csi0.framesize(csi.QVGA)

csi2 = csi.CSI(cid=csi.LEPTON)
csi2.reset(hard=False)
csi2.pixformat(csi.GRAYSCALE)
csi2.framesize(csi.QVGA)
csi2.ioctl(csi.IOCTL_LEPTON_SET_MODE, True, False)
csi2.ioctl(csi.IOCTL_LEPTON_SET_RANGE, 20.0, 40.0)

print(csi0, csi2)
clock = time.clock()

tmp = image.Image(csi0.width(), csi0.height(), csi0.pixformat())
img2 = image.Image(csi2.width(), csi2.height(), csi2.pixformat())

while True:
    clock.tick()
    img0 = csi0.snapshot()
    csi2.snapshot(update=False, blocking=False, image=img2)
    img0.draw_image(img2, 0, 0, transform=m, color_palette=image.PALETTE_IRONBOW,
                                       alpha_palette=alpha_pal,
                                       hint=image.BILINEAR)
    print(clock.fps())
```

### DIFF
--- a/modules/py_helper.c
+++ b/modules/py_helper.c
@@ -28,6 +28,9 @@
 #include "framebuffer.h"
 #include "py_helper.h"
 #include "py_assert.h"
+#if defined(MODULE_ULAB_ENABLED)
+#include "ndarray.h"
+#endif // MODULE_ULAB_ENABLED
 #if MICROPY_PY_CSI
 #include "omv_csi.h"
 #endif
@@ -102,6 +105,40 @@ const void *py_helper_arg_to_palette(const mp_obj_t arg, uint32_t pixfmt) {
         palette = img->data;
     }
     return palette;
+}
+
+void *py_helper_arg_to_transform(const mp_obj_t arg) {
+    if (arg == mp_const_none) {
+        return NULL;
+    }
+
+    #if defined(MODULE_ULAB_ENABLED) && defined(OPENMV_N6)
+    if (!MP_OBJ_IS_TYPE(arg, &ulab_ndarray_type)) {
+        mp_raise_msg(&mp_type_TypeError, MP_ERROR_TEXT("Expected a ndarray"));
+    }
+
+    ndarray_obj_t *ndarray = MP_OBJ_TO_PTR(arg);
+
+    if (ndarray->dtype != NDARRAY_FLOAT) {
+        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected a ndarray with dtype float"));
+    }
+
+    if (!ndarray_is_dense(ndarray)) {
+        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected a dense ndarray"));
+    }
+
+    if (ndarray->ndim != 2) {
+        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected a 2D ndarray"));
+    }
+
+    if ((ndarray->shape[ULAB_MAX_DIMS-2] != 3) || (ndarray->shape[ULAB_MAX_DIMS-1] != 3)) {
+        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Invalid matrix shape!"));
+    }
+
+    return ndarray->array;
+    #else
+    mp_raise_msg(&mp_type_TypeError, MP_ERROR_TEXT("Transform operations are not supported!"));
+    #endif // MODULE_ULAB_ENABLED
 }
 
 rectangle_t py_helper_arg_to_roi(const mp_obj_t arg, const image_t *img) {
--- a/ports/stm32/omv_gpu.c
+++ b/ports/stm32/omv_gpu.c
@@ -120,6 +120,22 @@ int omv_gpu_draw_image(image_t *src_img,
         return -1;
     }
 
+    float dx0 = dst_rect->x;
+    float dy0 = dst_rect->y;
+    float dx1 = dst_rect->x + dst_rect->w;
+    float dy1 = dst_rect->y;
+    float dx2 = dst_rect->x + dst_rect->w;
+    float dy2 = dst_rect->y + dst_rect->h;
+    float dx3 = dst_rect->x;
+    float dy3 = dst_rect->y + dst_rect->h;
+
+    if (transform) {
+        nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) transform), &dx0, &dy0);
+        nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) transform), &dx1, &dy1);
+        nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) transform), &dx2, &dy2);
+        nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) transform), &dx3, &dy3);
+    }
+
     // Create command list.
     #if OMV_GPU_NEMA_MM_STATIC
     nema_buffer_t bo = {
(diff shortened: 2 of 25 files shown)


iabdalkader:
That's exactly what I needed, thanks!

* The parameter should be called transform.
* I don't think we will ever need 4x4, right? So can we just pass a 3x3 with identity row and detect affine or not in the code? Alternatively, we could pass a 3x3 matrix + `affine=True/False` but it's more args.

The homography matrix seems to be working very well, even though it was just a quick test of the script.  I'll try to generate a better one and add the script.

kwagyeman:
> The parameter should be called transform.

Sure, but, makes the args somewhat wordy...

> I don't think we will ever need 4x4, right? So can we just pass a 3x3 with identity row and detect affine or not in the code?

The N6 GPU always does things in 3x3, so, it doesn't matter, the affline part is just to handle if the user passes a 2x3 matrix for convenience.

iabdalkader:
> Sure, but, makes the args somewhat wordy...

No problem, transform just makes it more clear what it does than `mat`.

> The N6 GPU always does things in 3x3, so, it doesn't matter, the affline part is just to handle if the user passes a 2x3 matrix for convenience.

We could detect that it's an affine transform by just checking the last 3 values, and that's only needed for the software fallback (which we may never actually add), otherwise everything else will be cleaner: For example, you can just cast it to [3][3] and pass directly to Nema and save a memcpy.

kwagyeman:
Okay:
* Renamed all the args to transform.
* Code only works on N6, other boards will get an error on using the argument.
* Directly pass the ndarray to the gpu now to avoid memcopies.

ports/stm32/omv_gpu.c
```diff
@@ -119,6 +121,33 @@ int omv_gpu_draw_image(image_t *src_img,
...
+            nema_mat3x3_mul_vec_affine(*((nema_matrix3x3_t *) ndarray->array), &dx0, &dy0);
+            nema_mat3x3_mul_vec_affine(*((nema_matrix3x3_t *) ndarray->array), &dx1, &dy1);
+            nema_mat3x3_mul_vec_affine(*((nema_matrix3x3_t *) ndarray->array), &dx2, &dy2);
+            nema_mat3x3_mul_vec_affine(*((nema_matrix3x3_t *) ndarray->array), &dx3, &dy3);
+        } else if (ndarray->shape[ULAB_MAX_DIMS-2] == 3) {
+            nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) ndarray->array), &dx0, &dy0);
+            nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) ndarray->array), &dx1, &dy1);
+            nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) ndarray->array), &dx2, &dy2);
+            nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) ndarray->array), &dx3, &dy3);
+        } else {
```
iabdalkader: > Directly pass the ndarray to the gpu now to avoid memcopies.

The code will only accept a 3x3 matrix, so it can pass around a float pointer. If you need to detect that it's an affine transform in the software fallback, if that even gets implement at all, do so by checking the last row :

> We could detect that it's an affine transform by just checking the last 3 values, and that's only needed for the software fallback (which we may never actually add), otherwise everything else will be cleaner: For example, you can just cast it to [3][3] and pass directly to Nema and save a memcpy.

```suggestion
        nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) transform), &dx0, &dy0);
        nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) transform), &dx1, &dy1);
        nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) transform), &dx2, &dy2);
        nema_mat3x3_mul_vec(*((nema_matrix3x3_t *) transform), &dx3, &dy3);
```

modules/py_helper.c
```diff
@@ -104,6 +107,41 @@ const void *py_helper_arg_to_palette(const mp_obj_t arg, uint32_t pixfmt) {
...
+    }
+
+    if (ndarray->ndim != 2) {
+        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Expected a 2D ndarray"));
+    }
+
+    if (((ndarray->shape[ULAB_MAX_DIMS-2] != 2) && (ndarray->shape[ULAB_MAX_DIMS-2] != 3)) ||
+        (ndarray->shape[ULAB_MAX_DIMS-1] != 3)) {
+        mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Invalid matrix shape!"));
+    }
```
iabdalkader: The only allowed arg here will be a 3x3 matrix.

iabdalkader (changes requested):

ports/stm32/omv_gpu.c
kwagyeman: Dropped 2x3 support.

modules/py_helper.c
kwagyeman: Though it was cool to have 2x3 support too, but, it doesn't matter. Just means you have to pass an extra row in the in the 3x3 matrix.

Also, regarding 4x4 support, this will never be needed. All the necessary linear alegrba to create a 3x3 matrix from a 4x4 one can be done in ULAB. A camera matrix can project the 4x4 3D space into the 3x3 2D space.

PR #2791, opened 2025-07-31
boards/OPENMV_N6: Enable triple buffering for the N6 by default.

kwagyeman:
Needs https://github.com/openmv/micropython/pull/136.

DMA SPI transmit output works now on the N6 for the LCD and TV shield.

Regarding the max transfer size, the DMA controllers for the N6 always work in bytes when passing the length and have a limit of 64K for a block. So, the 16-bit size had to be halved.

Security config settings were also missing for DMA non-circular mode.

### DIFF
--- a/ports/stm32/stm_dma.c
+++ b/ports/stm32/stm_dma.c
@@ -352,6 +352,12 @@ int stm_dma_init(DMA_HandleTypeDef *dma_descr, void *dma_channel, uint32_t reque
     dma_init_done = !circular;
     #endif
 
+    #if defined(STM32N6)
+    if (stm_dma_sec_config(dma_descr) != HAL_OK) {
+        return -1;
+    }
+    #endif
+
     if (dma_init_done) {
         HAL_DMA_DeInit(dma_descr);
         if (HAL_DMA_Init(dma_descr) != HAL_OK) {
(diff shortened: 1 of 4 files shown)


ports/stm32/stm_dma.c
```diff
@@ -357,6 +357,24 @@ int stm_dma_init(DMA_HandleTypeDef *dma_descr, void *dma_channel, uint32_t reque
         if (HAL_DMA_Init(dma_descr) != HAL_OK) {
             return -1;
         }
+
+        #if defined(STM32N6)
```
iabdalkader: This should be done in `stm_dma_init` for regular and circular modes, if it's okay to configure the security stuff before the rest of the code in `stm_dma_ll_init`, otherwise it should be a new function: `stm_dma_sec_config`

iabdalkader (changes requested):

ports/stm32/stm_dma.c
kwagyeman: Added a new function `stm_dma_sec_config`.

iabdalkader: Can we configure security before `stm_dma_ll_init` gets called? If so we can just configure it for N6 in this function unconditionally, before `HAL_DMA_Init`.

kwagyeman: Updated with the requested changes. However, uh... it's kinda dependent on the HAL not changing on us. The functions can be called right now before init... but, that might not be the case in the future. The HAL doesn't check the ready state flag for setting the security config. They could always add that, though.

iabdalkader: In the examples it's done after the initialization. However, the only restriction on updating these registers is that the channel must be disabled, so it doesn't matter where it's called as long as it's before a transfer is started. I'll fix this in a separate PR.

PR #2793, opened 2025-08-02
common/vospi: Fix synchronization issues.

iabdalkader:
- Abort if synchronization takes too long.
- Reinit SPI on transfer restart. Fixes and issue on N6.
- Clean up vospi code.


kwagyeman:
Works great! Tested all configurations of FLIR Leptons on all cameras without issues. In fact, issues with the FLIR Lepton 3.5 syncing on the N6 were resolved too. That's now fast and reliable.

kwagyeman (approved):

PR #2722, opened 2025-06-23
Kwabena/speedup 2

kwagyeman:
WIP

Still have one more idea to try out for speed before finishing this.

PR #2798, opened 2025-08-05
lib/imlib: New SIMD ops for ML acceleration

kwagyeman:
SIMD ops broken out of PR: https://github.com/openmv/openmv/pull/2722

### DIFF
--- a/lib/imlib/simd.h
+++ b/lib/imlib/simd.h
@@ -1371,6 +1371,30 @@ static inline v128_t vldr_u8_pred(const uint8_t *p, v128_predicate_t pred) {
     #endif
 }
 
+static inline v128_t vldr_u8_widen_u32_gather_pred(const uint8_t *p,
+                                                   v128_t offsets,
+                                                   v128_predicate_t pred) {
+    #if (__ARM_ARCH >= 8)
+    return (v128_t) vldrbq_gather_offset_z_u32(p, offsets.u32, pred);
+    #else
+    return (v128_t) {
+        .u32 = { *(p + offsets.u32[0]) }
+    };
+    #endif
+}
+
+static inline v128_t vldr_s8_widen_s32_gather_pred(const int8_t *p,
+                                                   v128_t offsets,
+                                                   v128_predicate_t pred) {
+    #if (__ARM_ARCH >= 8)
+    return (v128_t) vldrbq_gather_offset_z_s32(p, offsets.u32, pred);
+    #else
+    return (v128_t) {
+        .s32 = { *(p + offsets.u32[0]) }
+    };
+    #endif
+}
+
 static inline void vstr_u8(uint8_t *p, v128_t v0) {
     #if (__ARM_ARCH >= 8)
     vstrbq(p, v0.u8);
@@ -1422,6 +1446,30 @@ static inline v128_t vldr_u16_pred(const uint16_t *p, v128_predicate_t pred) {
     #endif
 }
 
+static inline v128_t vldr_u16_widen_u32_gather_pred(const uint16_t *p,
+                                                    v128_t offsets,
+                                                    v128_predicate_t pred) {
+    #if (__ARM_ARCH >= 8)
+    return (v128_t) vldrhq_gather_shifted_offset_z_u32(p, offsets.u32, pred);
+    #else
+    return (v128_t) {
+        .u32 = { *(p + offsets.u32[0]) }
+    };
+    #endif
+}
+
+static inline v128_t vldr_s16_widen_s32_gather_pred(const int16_t *p,
+                                                    v128_t offsets,
+                                                    v128_predicate_t pred) {
+    #if (__ARM_ARCH >= 8)
+    return (v128_t) vldrhq_gather_shifted_offset_z_s32(p, offsets.u32, pred);
+    #else
+    return (v128_t) {
+        .s32 = { *(p + offsets.u32[0]) }
+    };
+    #endif
+}
+
 static inline void vstr_u16(uint16_t *p, v128_t v0) {
     #if (__ARM_ARCH >= 8)
     vstrhq(p, v0.u16);
@@ -1468,6 +1516,15 @@ static inline void vstr_u16_narrow_u8_pred(uint8_t *p, v128_t v0, v128_predicate
     if (pred > 1) {
         *((uint8_t *) (p + 1)) = v0.u8[2];
     }
+    #endif
+}
+
+static inline void vstr_u16_narrow_u8_scatter(uint8_t *p, v128_t offsets, v128_t v0) {
+    #if (__ARM_ARCH >= 8)
+    vstrbq_scatter_offset(p, offsets.u16, v0.u16);
+    #else
+    *(p + offsets.u16[0]) = v0.u8[0];
+    *(p + offsets.u16[1]) = v0.u8[2];
     #endif
 }
 
--- a/lib/imlib/simd.h
+++ b/lib/imlib/simd.h
@@ -844,7 +844,7 @@ static inline v128_t vshlc(v128_t v0, uint32_t *reg, uint32_t n) {
 
 static inline v128_t vadd_u32(v128_t v0, v128_t v1) {
     #if (__ARM_ARCH >= 8)
-    return (v128_t) vaddq(v0.u32, v1.u32);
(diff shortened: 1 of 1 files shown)


iabdalkader:
Why was this closed?

kwagyeman:
Oh, I misunderstood your message in the chat. Was going to break out this with commit: https://github.com/openmv/openmv/pull/2722/commits/9958d8ee6239d7569e6cf6287d0a1dd491cf7db7

I'll reopen, it's ready to go otherwise.

iabdalkader:
> Was going to break out this with commit

Send all your `simd.h` changes here.

kwagyeman:
This is everything.

lib/imlib/simd.h
```diff
@@ -1330,6 +1345,74 @@ static inline uint32_t vmladava_u16(v128_t v0, v128_t v1, uint32_t acc) {
...
 
+static inline v4x_rows_t vcvt_f32_u8(v128_t v0) {
+    #if (__ARM_ARCH >= 8)
+    v128_t b = (v128_t) vmovlbq(v0.u8);
+    v128_t t = (v128_t) vmovltq(v0.u8);
+    return (v4x_rows_t) {
+        .r0 = (v128_t) vcvtq(vmovlbq(b.u16)),
+        .r1 = (v128_t) vcvtq(vmovlbq(t.u16)),
+        .r2 = (v128_t) vcvtq(vmovltq(b.u16)),
+        .r3 = (v128_t) vcvtq(vmovltq(t.u16)),
```
iabdalkader, line 1356: Why this particular layout? Looks like this function isn't generic and for something specific. If so, shouldn't the name reflect that?

kwagyeman: It's the u8 to f32 conversion method. For u32 -> f32, this can be done in one instruction, 4 ints in yield 4 floats out. This maps exactly to one instruction as there's no data width expansion.

For u8, I have to unpack things using the mov top and move bottom instructions to unpack u8 values to u16, and then u16 to u32, and then do the conversion. Because of the data width expansion, there's never going to be an exact instruction match for this type of operation, but, it's still an operation in the family of converting an integer to a float32 so I named it similarly.

iabdalkader: I understand what it does, I'm asking why that particular interleaved order? Seems to end up with 0–3, 8–11, 4–7, 12–15.

> It's the u8 to f32 conversion method

So shouldn't be `vcvt_u8_f32` ?

iabdalkader: Let me rephrase:
```C
    v128_t b = (v128_t) vmovlbq(v0.u8);  // [0...7]
    v128_t t = (v128_t) vmovltq(v0.u8);   // [8...15]

    return (v4x_rows_t) {
        .r0 = (v128_t) vcvtq(vmovlbq(b.u16)),  // [0..3]

        .r1 = (v128_t) vcvtq(vmovlbq(t.u16)),   // [8..11]

        .r2 = (v128_t) vcvtq(vmovltq(b.u16)),   // [4...7]

        .r3 = (v128_t) vcvtq(vmovltq(t.u16)),    // [12...15]
```
Should it be 0..3, 4..7 etc..?

kwagyeman: > Seems to end up with 0–3, 8–11, 4–7, 12–15.

Not quite :)

`v128_t b = (v128_t) vmovlbq(v0.u8);`

Creates:

u8[0]-> u16[0]
u8[2]-> u16[1]
u8[4]-> u16[2]
u8[6]-> u16[3]
... etc.

And `v128_t t = (v128_t) vmovltq(v0.u8);`

u8[1]-> u16[0]
u8[3]-> u16[1]
u8[5]-> u16[2]
u8[7]-> u16[3]
... etc.

Then: `.r0 = (v128_t) vmovlbq(b.u16)`

u16[0] -> u32[0]
u16[2] -> u32[1]
u16[4] -> u32[2]
u16[6] -> u32[3]

So, r0 will contain:

u8[0]-> u32[0]
u8[4]-> u32[1]
u8[8]-> u32[2]
u8[12]-> u32[3]

r1 will have:

u8[1]-> u32[0]
u8[5]-> u32[1]
u8[9]-> u32[2]
u8[13]-> u32[3]

r2 will have:

u8[2]-> u32[0]
u8[6]-> u32[1]
u8[10]-> u32[2]
u8[14]-> u32[3]

r3 will have:

u8[3]-> u32[0]
u8[7]-> u32[1]
u8[11]-> u32[2]
u8[15]-> u32[3]

What's going on is that things are interleaved. When you see any of the v2x_rows_t or v4x_rows_t rows stuff values are interleaved across them. There's no good way to do something like 0–3, 4-7, 8-11, and 12-15 easily with Helium as the move instructions always skip a value when widening. Same for narrowing. This follows with how they implement loads and stores too like vld4: https://developer.arm.com/architectures/instruction-sets/intrinsics/%5B__arm_%5Dvld4q%5B_s32%5D. Values are always interleaved across rows.

> So shouldn't be vcvt_u8_f32 ?

Yes... That's probably better. Will change.

iabdalkader: > Not quite :)
> 
> v128_t b = (v128_t) vmovlbq(v0.u8);
> 
> Creates:
> 
> u8[0]-> u16[0]
> u8[2]-> u16[1]
> u8[4]-> u16[2]
> u8[6]-> u16[3]
> 

If that's true, then wouldn't `vmovlbq(v0.u16);` in `vcvt_f32_u16` do the same:
```
u16[0] -> u32[0]
u16[2] -> u32[1]
u16[4] -> u32[2]
u16[6] -> u32[3]
```

kwagyeman: Yeah, that's what it does. And then values are interleaved across two rows with `v2x_rows_t`.

```
u16[0] -> r0.u32[0]
u16[2] -> r0.u32[1]
u16[4] -> r0.u32[2]
u16[6] -> r0.u32[3]

u16[1] -> r1.u32[0]
u16[3] -> r1.u32[1]
u16[5] -> r1.u32[2]
u16[7] -> r1.u32[3]
```

iabdalkader: Okay, now why do we want them interleaved like that and not sequential?

iabdalkader: Seems the end result would be
```
r0: 0 4 8  12
r1: 1 5 9  13
r2: 2 6 10 14
r3: 3 7 11 15
```

kwagyeman: > Seems the end result would be

Correct!

There's no way to do sequential easily when widening values, same for narrowing. They designed it such that you have to do things in this interleaved way. Otherwise, you have to shift and mask values together one at a time.

For Helium, think about things in the concept of "lanes", generally, any operation can't leave it's lane. So, like, an 8-bit value will only ever be allowed to operate in the 8-bit lane it's in or be widened to the 16-bit lane it's inside of. For narrowing, a 16-bit value can only be narrowed into the top of bottom 8-bits of it's lane.

Interleaving across registers then flows out of this restriction as a way to deal with the data width expansion.

iabdalkader: I guess all other instructions expect this order so it's probably fine.

lib/imlib/simd.h
```diff
@@ -1290,6 +1345,74 @@ static inline uint32_t vmladava_u16(v128_t v0, v128_t v1, uint32_t acc) {
     #endif
 }
 
+static inline v4x_rows_t vcvt_f32_u8(v128_t v0) {
```
iabdalkader:
```suggestion
static inline v4x_rows_t vcvt_u8_f32(v128_t v0) {
```
Same for the others.

iabdalkader (changes requested):

kwagyeman:
Yup, will do, just finishing up the genx320 pr right now.

lib/imlib/simd.h
kwagyeman: Updated

iabdalkader (approved):

lib/imlib/simd.h
iabdalkader: @kwagyeman I missed this. Just re-request preview in the future.

kwagyeman:
Thanks, will get the SIMD stuff broken out of the ML PR today and benchmarked.

PR #2799, opened 2025-08-06
lib/stm32: Disable small size single buffer DMA mode.

kwagyeman:
The N6 crashes if you try to capture an image that's less than 64KB in size using any parallel bus sensor using `HAL_DCMI_Start_DMA`. The reason is that the HAL switches modes from the circular buffer linked list DMA to the non-circular buffer linked list DMA. Oddly, if you call with a size larger than 64KB first, the other path is taken in the HAL... which, after being executed once... prevents the smaller-sized path from crashing. So, more issues with linked list initialization in the HAL.

This commit just disables the single buffer path entirely as it's not needed and the crash no-longer happens.

Just capture an image at QQVGA to see the issue with any parallel sensor from cold boot.

### DIFF
--- a/lib/stm32/n6/src/stm32n6xx_hal_dcmi.c
+++ b/lib/stm32/n6/src/stm32n6xx_hal_dcmi.c
@@ -418,7 +418,7 @@ HAL_StatusTypeDef HAL_DCMI_Start_DMA(DCMI_HandleTypeDef *hdcmi, uint32_t DCMI_Mo
   /* Length should be converted to number of bytes */
   tmp_length = tmp_length * 4U;
 
-  if (tmp_length <= 0xFFFFU)
+  if (tmp_length <= 0)
   {
     /* Continuoues Mode */
     /* Enable the DMA Stream */


iabdalkader:
Would reverting back to init/deinit on config calls fix this or is it unrelated? I'm close to giving up on "init once and reuse".

kwagyeman:
No, it doesn't work on the first call ever. It's probably fixable in our drivers by changing how the list is setup for <= 64KB transfers. But, this is just weird. It seems like a HAL bug that the behavior of the driver would change based on the transfer size. Case in point, the code they run changes based on the transfer size... for no real reason too. Disabling it seems like the right approach.

iabdalkader:
> It seems like a HAL bug that the behavior of the driver would change based on the transfer size

All of the HAL DCMI driver historically did the same more or less.

> probably fixable in our drivers by changing how the list is setup for <= 64KB transfers

If it is, it's best to workaround it in our drivers, maintaining HAL changes is difficult.

kwagyeman:
Will figureout a better fix today.

kwagyeman:
Fixed by https://github.com/openmv/openmv/pull/2814.

PR #2801, opened 2025-08-08
scripts/libraries: Post-processing speedup.

kwagyeman:
From PR: https://github.com/openmv/openmv/pull/2722

Takes commits:
* [modules/py_ml: Return tensor references for post-processors.](https://github.com/openmv/openmv/pull/2722/commits/57c7fc537406eb754bffe5f723a21cb1ecb72f14)
* [scripts/libraries: Quantize threshold instead of dequantizing scores.](https://github.com/openmv/openmv/pull/2722/commits/a40976e1b3af68c2bacc2f3530e91a5137c6ab74)

This PR significantly speeds up post-processing of models by doing the following:
1. It avoids deep copying the tensor arena output. This also has the benefit of reducing heap pressure for larger models.
2. It thresholds the quantized int8 output tensor to find valid scores without dequantizing, this saves a huge amount of time.
3. Dequantization is only done on valid objects.

N6 Benchmarks:

| N6 Before | N6 After | Speedup |
|------------|-----------|---------|
| fomo-96x96-1-class-pp = 40 us | fomo-96x96-1-class-pp = 19 us | **2.11×** |
| fomo-96x96-1-class-pp_callback = 1380 us | fomo-96x96-1-class-pp_callback = 950 us | **1.45×** |
| yolov2-224x224-1-class-pp = 53 us | yolov2-224x224-1-class-pp = 24 us | **2.21×** |
| yolov2-224x224-1-class-pp_callback = 1851 us | yolov2-224x224-1-class-pp_callback = 1387 us | **1.33×** |
| yolov5-224x224-1-class-pp = 769 us | yolov5-224x224-1-class-pp = 26 us | **29.58×** |
| yolov5-224x224-1-class-pp_callback = 1580 us | yolov5-224x224-1-class-pp_callback = 1219 us | **1.30×** |
| yolov8-192x192-1-class-pp = 192 us | yolov8-192x192-1-class-pp = 28 us | **6.86×** |
| yolov8-192x192-1-class-pp_callback = 2551 us | yolov8-192x192-1-class-pp_callback = 1263 us | **2.02×** |
| yolov8-512x512-80-class-pp = 30367 us | yolov8-512x512-80-class-pp = 177 us | **171.57×** |
| yolov8-512x512-80-class-pp_callback = 168398 us | yolov8-512x512-80-class-pp_callback = 39397 us | **4.28×** |

The PR shows a speedup for the demo networks with one class output and a massive speedup a high resolution YOLOV8 network with tons of class outputs. Similar results for larger user networks are expected with fomo, yolov2, and yolov5.

### DIFF
--- a/scripts/libraries/ml/ml/postprocessing.py
+++ b/scripts/libraries/ml/ml/postprocessing.py
@@ -38,6 +38,23 @@ def mod(a, b):
     return a - (b * (a // b))
 
 
+def threshold(scores, threshold, scale, find_max=False, find_max_axis=1):
+    if scale > 0:
+        if find_max:
+            scores = np.max(scores, axis=find_max_axis)
+        return np.nonzero(scores > threshold)[0]
+    else:
+        if find_max:
+            scores = np.min(scores, axis=find_max_axis)
+        return np.nonzero(scores < threshold)[0]
+
+
+def quantize(model, value):
+    if model.output_dtype[0] == 'f':
+        return value
+    return (value / model.output_scale[0]) + model.output_zero_point[0]
+
+
 def dequantize(model, value):
     if model.output_dtype[0] == 'f':
         return value
@@ -57,13 +74,15 @@ class fomo_postprocess:
 
     def __call__(self, model, inputs, outputs):
         ob, oh, ow, oc = model.output_shape[0]
+        scale = model.output_scale[0]
+        t = quantize(model, self.threshold)
 
         # Reshape the output to a 2D array
         row_outputs = outputs[0].reshape((oh * ow, oc))
 
         # Threshold all the scores
-        score_indices = np.max(dequantize(model, row_outputs[:, _FOMO_CLASSES:]), axis=1)
-        score_indices = np.nonzero(score_indices > self.threshold)[0]
+        score_indices = row_outputs[:, _FOMO_CLASSES:]
+        score_indices = threshold(score_indices, t, scale, find_max=True, find_max_axis=1)
         if not len(score_indices):
             return _NO_DETECTION
 
@@ -121,8 +140,9 @@ class yolo_v2_postprocess:
         self.nms_sigma = nms_sigma
 
     def __call__(self, model, inputs, outputs):
-        ob, oh, ow, oc = model.output_shape[0]
-        class_count = (oc // self.anchors_len) - _YOLO_V2_CLASSES
+
+        def logit(x):
+            return np.log(x / (1.0 - x))
 
         def sigmoid(x):
             return 1.0 / (1.0 + np.exp(-x))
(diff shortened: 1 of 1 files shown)


iabdalkader:
We should merge #2816 first then rebase this, it's a good opportunity to test the profiler.

kwagyeman:
Okay, was expecting that.

scripts/examples/03-Machine-Learning/00-TensorFlow/tf_object_detection.py
```diff
@@ -40,39 +36,15 @@
...
+fomo = fomo_postprocess(threshold=0.4)
 
 clock = time.clock()
 while True:
     clock.tick()
 
     img = sensor.snapshot()
 
-    for i, detection_list in enumerate(model.predict([img], callback=fomo_post_process)):
+    for i, detection_list in enumerate(model.predict([img], callback=fomo)):
```
iabdalkader:
```suggestion
    for i, detection_list in enumerate(model.predict([img], callback=fomo)):
```
Please revert this. Also, keep changes to libraries and changes to examples in separate commits.

scripts/libraries/ml/ml/postprocessing.py
```diff
@@ -26,55 +26,126 @@
...
+_YOLO_V5_CY = const(1)
+_YOLO_V5_CW = const(2)
+_YOLO_V5_CH = const(3)
+_YOLO_V5_SCORE = const(4)
+_YOLO_V5_CLASSES = const(5)
+_YOLO_V8_CX = const(0)
+_YOLO_V8_CY = const(1)
+_YOLO_V8_CW = const(2)
+_YOLO_V8_CH = const(3)
+_YOLO_V8_CLASSES = const(4)
```
iabdalkader: I don't like this, please revert. In fact please drop [scripts/libraries: Cleanup post-processing.](https://github.com/openmv/openmv/pull/2801/commits/b9b738127ce2b409b7f0b1a22960071d13cbfe6e). Only move the functions/constants to golbal scope if they are shared, and they're not.

iabdalkader (changes requested):
The commits here make multiple changes to the same lines, each reverting the previous one. Why not just make the changes you want to make so I can read this? Like I understand commit `n` then commit `n+1` removes that, and does something else. Very valuable time is wasted on this.

Also, please move YoloV8 transpose to a new PR, and move NMS to new PR. Keep the absolute bare minimum in this PR.

scripts/libraries/ml/ml/postprocessing.py
kwagyeman: Sure, will have to move mod() to global scope though, but, can do that in one of the later commits.

scripts/examples/03-Machine-Learning/00-TensorFlow/tf_object_detection.py
kwagyeman: The example won't continue to work with the updated post-processing. Do you want me to update it to match the new fomo_post_process call in the library code? Like, your intention here is that the example shows off how to write a post-processor, correct?

kwagyeman:
Will drop https://github.com/openmv/openmv/pull/2801/commits/2be44b502759342dfa77d49e6d2690c4ee36a1c3 given the new performance profiler.

scripts/examples/03-Machine-Learning/00-TensorFlow/tf_object_detection.py
iabdalkader: I mean pass `fomo_postprocess(threshold=0.4)` directly, since 0.4 is the default you probably also don't need that.

scripts/libraries/ml/ml/postprocessing.py
kwagyeman: Done

scripts/examples/03-Machine-Learning/00-TensorFlow/tf_object_detection.py
kwagyeman: Done.

kwagyeman:
Restesting changes...

kwagyeman:
Okay, tested and updated.

iabdalkader:
I tested fomo, yolov2, yolov5 etc.. none of them show any significant improvement, maybe this is expected because they never had many outputs to begin with? Anyway, I tested YoLoV8n and it does show great improvement. Regardless of performance numbers, what's done here seems logical. Looking at the profiler, it seems the post processing basically spends all of its time `numerical_argmin_argmax_ndarray`, this would make a good target to optimize in ulab.

kwagyeman:
Yeah, it only really helps the large networks. Note, with this PR merged, a few networks that didn't fit on the AE3 will now fit because memory is used more efficiently.

The next SIMD PR addresses `numerical_argmin_argmax_ndarray.`. Going to rework what I sent previously though which was only a 4x speedup. I can actually hit 16x. There will be a noticeable improvement to all nets.

PR #2802, opened 2025-08-08
common/csi: Add support for custom resolutions.

iabdalkader:
See https://github.com/openmv/openmv/pull/2747#discussion_r2262310045


kwagyeman:
Works for me. Merge it!

iabdalkader:
Note: should we ever have two sensors that both need custom resolutions at the same time, I have a fix for that, but it's very unlikely.

PR #2804, opened 2025-08-09
modules/py_csi_ng: Support custom frame size.

iabdalkader:
When the `sensor` module is removed eventually, all of the `Bx` resolutions will be removed as well.

### DIFF
--- a/common/omv_csi.c
+++ b/common/omv_csi.c
@@ -77,51 +77,50 @@ typedef struct _i2c_dev {
 
 // Sensor frame size/resolution table.
 uint16_t resolution[][2] = {
-    {0,    0   },
-    {0,    0   },    /* Custom set by drivers */
+    [OMV_CSI_FRAMESIZE_INVALID]     = {0,    0},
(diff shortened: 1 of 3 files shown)


kwagyeman (approved):
Looks good to me!

iabdalkader:
> Looks good to me!

Are there other resolutions that we should remove form the new API? For example:
```C
    { MP_ROM_QSTR(MP_QSTR_LCD),             MP_ROM_INT(OMV_CSI_FRAMESIZE_LCD) },      /* 128x160   */
    { MP_ROM_QSTR(MP_QSTR_QQVGA2),          MP_ROM_INT(OMV_CSI_FRAMESIZE_QQVGA2) },   /* 128x160   */
```
Then we can replace those with their values in the scripts that use them, when updating to `csi` API.

Also, the super small reses like `OMV_CSI_FRAMESIZE_HQQQQVGA` and the like.

kwagyeman (approved):

iabdalkader:
Forgot to mention, I tested himax with `B320X320` and `framesize((320, 320))`.

PR #2814, opened 2025-08-15
ports/stm32: Fix STM32 DCMI Driver

kwagyeman:
Tested:

M4 - OV2640: All modes work.
M4 - OV7725: All modes work.
M7 - OV7725: All modes work.
H7P - All modes work across all supported camera modules.
N6 - All supported modes work across all supported camera modules.

Note on N6 Performance: For resolutions above VGA generally, you'll get frame timeout errors for the older parallel cameras. The issue is that the HPDMA controller doesn't quickly transfer data to SDRAM fast enough. I debugged the issue back to the DCMI fifo overflowing. It appears to suffer from the same issues we saw on the H7 Plus where you got frame timeouts if you targeted DMA2 to write to SDRAM directly. I tried adjusting the destination burst sizes, switching to 64-bit destination mode, and etc. nothing works. The controller just can't handle the long and sustained bursts from cameras as the line width get longer.

The solution to this issue is probably the same, to have a GPDMA controller reading the parallel bus, writing to two line buffers, and then HPDMA reading the line buffers to write to SDRAM. Probably can more or less copy how the MDMA implementation on the H7/H7 Plus work. However, this work is better left for another PR and to be done in the future when there's a need to increase parallel camera performance.

Note that the GENX320 sensor works without issues and can hit 2000 FPS in event mode with the current code and 380 fps in histogram mode on the N6. So, there's no pressing need for this work at this time. The BOSON and the global shutter camera also work to up to their max res. Only the OV5640 will experience a resolution limit, but, this is not important.

### DIFF
--- a/ports/stm32/omv_csi.c
+++ b/ports/stm32/omv_csi.c
@@ -66,7 +66,7 @@
 #endif
 
 #ifndef OMV_CSI_DMA_MAX_SIZE
-#define OMV_CSI_DMA_MAX_SIZE    (0xFFFFU * 4U)
+#define OMV_CSI_DMA_MAX_SIZE    (0xFFFFU)
 #endif
 
 #ifndef OMV_CSI_LINE_ALIGNMENT
@@ -625,8 +625,8 @@ static int stm_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
             // differing only in size, with an interrupt after every half of the transfer.
             if ((csi->pixformat == PIXFORMAT_JPEG) && (csi->jpg_format == 3)) {
                 // Start a one-shot transfer to the framebuffer, used only for JPEG mode 3.
-                uint32_t size = framebuffer_get_buffer_size(fb);
-                csi->dma_size = IM_MIN(size, (OMV_CSI_DMA_MAX_SIZE * 2U)) / 4;
+                uint32_t size = framebuffer_get_buffer_size(fb) / 4;
+                csi->dma_size = IM_MIN(size, OMV_CSI_DMA_MAX_SIZE * 2U);
                 csi->one_shot = true;
                 HAL_DCMI_Start_DMA(&csi->dcmi, DCMI_MODE_SNAPSHOT, (uint32_t) buffer->data, csi->dma_size);
             #if USE_MDMA
@@ -648,7 +648,7 @@ static int stm_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
                 }
 
                 // Disable circular mode for transfer sizes less than 64KB.
-                if (csi->dma_size * 4 <= OMV_CSI_DMA_MAX_SIZE / 4) {
+                if (csi->dma_size * 4 <= OMV_CSI_DMA_MAX_SIZE) {
                     HAL_DMAEx_List_ClearCircularMode(&csi->dma_queue);
                 } else {
                     HAL_DMAEx_List_SetCircularMode(&csi->dma_queue);
@@ -726,7 +726,7 @@ static int stm_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
                 size = buffer->offset;
             } else {
                 // HAL_DCMI_Start_DMA splits bigger transfers.
-                if (csi->dma_size > (OMV_CSI_DMA_MAX_SIZE / 4)) {
+                if (csi->dma_size > OMV_CSI_DMA_MAX_SIZE) {
                     csi->dma_size /= 2;
                 }
 


ports/stm32/omv_csi.c
```diff
@@ -617,6 +617,14 @@ static int stm_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
...
                 HAL_DCMI_Start_DMA_MB(&csi->dcmi, DCMI_MODE_CONTINUOUS, (uint32_t) &_line_buf, csi->dma_size, fb->v);
                 #else
+                // Handle YUV422 Source -> Y Destination using DCMI byte drop.
+                if ((csi->pixformat == PIXFORMAT_GRAYSCALE) && (csi->mono_bpp == 2)) {
+                    csi->dma_size /= 2;
+                    csi->dcmi.Instance->CR |= DCMI_CR_BSM_0;
+                } else {
+                    csi->dcmi.Instance->CR &= ~DCMI_CR_BSM_0;
+                }
+
```
iabdalkader: This should be in `config == OMV_CSI_CONFIG_PIXFORMAT`

ports/stm32/omv_csi.c
```diff
@@ -625,6 +625,16 @@ static int stm_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
...
+                // Turn on/off byte swapping for RGB/YUV formats.
+                for (size_t i = 0; i < OMV_ARRAY_SIZE(dma_nodes); i++) {
+                    if ((csi->pixformat == PIXFORMAT_RGB565 && csi->rgb_swap) ||
+                        (csi->pixformat == PIXFORMAT_YUV422 && csi->yuv_swap)) {
+                        dma_nodes[i].LinkRegisters[NODE_CTR1_DEFAULT_OFFSET] |= DMA_CTR1_DBX;
+                    } else {
+                        dma_nodes[i].LinkRegisters[NODE_CTR1_DEFAULT_OFFSET] &= ~DMA_CTR1_DBX;
+                    }
+                }
+
```
iabdalkader: This should also be in `config == OMV_CSI_CONFIG_PIXFORMAT`, and squashed with previous commit.

ports/stm32/omv_csi.c
```diff
@@ -611,13 +617,43 @@ static int stm_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
...
+                // HAL_DCMI_Start_DMA() for transfer sizes less than 64KB expects a non-circular linked list.
+                if ((csi->dma_size * 4) <= (OMV_CSI_DMA_MAX_SIZE / 4)) {
+                    // Modify the link list head to point to nothing.
+                    dma_nodes[0].LinkRegisters[NODE_CLLR_LINEAR_DEFAULT_OFFSET] = 0;
+                } else {
+                    // Modify the link list head to point to the tail node (default).
+                    dma_nodes[0].LinkRegisters[NODE_CLLR_LINEAR_DEFAULT_OFFSET] =
+                        DMA_CLLR_UT1 | DMA_CLLR_UT2 | DMA_CLLR_UB1 | DMA_CLLR_USA | DMA_CLLR_UDA | DMA_CLLR_ULL |
+                        (((uint32_t) &dma_nodes[1]) & DMA_CLLR_LA);
+                }
```
iabdalkader:
```suggestion
                // Disable circular mode for transfer sizes less than 64KB.
                if (csi->dma_size <= OMV_CSI_DMA_MAX_SIZE / 4) {
                    HAL_DMAEx_List_ClearCircularMode(&dma_queue);
                } else {
                    HAL_DMAEx_List_SetCircularMode(&dma_queue);
                }
```
The test seems to off to me. `dma_size` is number of elements, `OMV_CSI_DMA_MAX_SIZE` is bytes so either multiply `dma_size` or divide `OMV_CSI_DMA_MAX_SIZE` but not both.

You can test it with:
```Python
sensor.set_framesize(sensor.QQVGA)
sensor.skip_frames(time=2000)
sensor.skip_frames(time=2000)
```

About all the extra parentheses put everywhere:

https://en.cppreference.com/w/c/language/operator_precedence.html

iabdalkader (changes requested):

iabdalkader:
Could you also fix this comment: https://github.com/openmv/openmv/blob/master/ports/stm32/omv_csi.c#L87
```C
// Nodes can't be placed in CSI state, as they must remain uncacheable.
```
Also, please move `dma_queue` to `OMV_CSI_PORT_BITS_DCMIPP`.

ports/stm32/omv_csi.c
kwagyeman: Done, but, I still need:

```
if (csi->pixformat == PIXFORMAT_GRAYSCALE && csi->mono_bpp == 2) {
    csi->dma_size /= 2;
}
```

Before starting DMA.

ports/stm32/omv_csi.c
kwagyeman: Done

ports/stm32/omv_csi.c
kwagyeman: Changed, but the test is correct. In `HAL_DCMI_Start_DMA` they do:

```
uint32_t tmp_length = Length;
tmp_length = tmp_length * 4U;
if (tmp_length <= 0xFFFFU) {
// non-circular mode
} else {
// circular mode
}
```

OMV_CSI_DMA_MAX_SIZE is `(0xFFFFU * 4U)`. So to mimick this I have to do:

```
if (csi->dma_size * 4 <= OMV_CSI_DMA_MAX_SIZE / 4)
```

As `csi->dma_size` is `Length` above.

kwagyeman: Simplified things by changing `OMV_CSI_DMA_MAX_SIZE` from `0xFFFFU * 4` to `0xFFFFU`.

PR #2817, opened 2025-08-18
lib/cmsis: Fix cmsis compile time constant folding.

kwagyeman:
Copied what to do from: https://github.com/openmv/openmv/blob/master/lib/cmsis/include/cmsis_gcc.h#L1989

Tested a few functions that depend on these and didn't notice any issues.


iabdalkader:
I found a workaround by excluding CMSIS headers from instrumentation, but then realized we need to do that anyway. I'm still merging this so it doesn't fail again.

iabdalkader:
Actually, the issue is not fixed it fails on RT1060, without filtering the headers.

PR #2818, opened 2025-08-19
common/tinyusb_debug: Fix system lockup in debug task.

kwagyeman:
Interrupts can be disabled by user space when the debug task is called which will result in it spinning forever. It has been modified to handle this now without issue.

Feel free to close the PR and fix in a better way than below. The below appears to work.

### DIFF
--- a/common/tinyusb_debug.c
+++ b/common/tinyusb_debug.c
@@ -37,6 +37,8 @@
 #include "tinyusb_debug.h"
 #include "omv_common.h"
 
+#include "cmsis_gcc.h"
+
 #define USBDBG_BAUDRATE_SLOW     (921600)
 #define USBDBG_BAUDRATE_FAST     (12000000)
 #define DEBUG_EP_SIZE           (TUD_OPT_HIGH_SPEED ? 512 : 64)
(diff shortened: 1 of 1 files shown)


iabdalkader:
Is this for the critical section for I2C target?

common/tinyusb_debug.c
```diff
@@ -188,6 +197,10 @@ void tinyusb_debug_task(mp_sched_node_t *node) {
                 usbdbg_data_out(size, tud_cdc_read);
             }
         }
+        // Exit if interrupts are disabled.
+        if (__get_PRIMASK()) {
```
iabdalkader: The rest of the bits are reserved, means in the future they could mean something so this must be `& 1`

kwagyeman: Done

kwagyeman:
> Is this for the critical section for I2C target?

Yes, for the shadow memory thing we talked about. I started to try to implement something and it caused the MCU to crash.

Doing something like the below causes a crash on the board. The debugger quickly showed the issue was that the CPU was stuck in the tinyusb_debug_task with primask set to 1.

```
        s = machine.disable_irq()
        if self._front[0] & FLAG_LOCK:
            machine.enable_irq(s)
            return
        self._front[:] = b[:]
        machine.enable_irq(s)
```

iabdalkader:
There's a function to disable the scheduler but it's not accessible from Python. Maybe this is a good argument for having that upstream?

kwagyeman:
You shouldn't have to disable the scheduler from Python before you disable_irq()... disable_irq() should disable all interrupts, including the scheduler running, as that's kinda like an interrupt. disable_irq() should be adjusted to probably do that.

Anyway, a fix via disabling the scheduler or via this PR is fine with me. Just need something.

kwagyeman:
https://github.com/openmv/openmv/pull/2820 replaces this.

PR #2820, opened 2025-08-23
common: Refactor TinyUSB debug interface implementation.

iabdalkader:
This commit refactors the TinyUSB debug interface with the following changes:

- Replace global variables with centralized state.
- Add timeout, recovery and validation for more robust transfers (as much as possible given the protocol limits).
- Handle disabled IRQs.

### DIFF
--- a/common/tinyusb_debug.c
+++ b/common/tinyusb_debug.c
@@ -25,7 +25,8 @@
  */
 
 #include "omv_boardconfig.h"
-#if (OMV_TUSBDBG_ENABLE == 1)
+
+#if OMV_TUSBDBG_ENABLE
 #include "py/runtime.h"
 #include "py/stream.h"
 #include "py/mphal.h"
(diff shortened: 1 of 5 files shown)


common/tinyusb_debug.c
```diff
@@ -36,159 +37,177 @@
...
-    usbdbg_cmd_t cmd;
-    tud_cdc_read(&cmd, sizeof(cmd));
+    if (!ctx.length && tud_cdc_available() >= USBDBG_HEADER_SIZE) {
+        uint8_t cmdbuf[USBDBG_HEADER_SIZE];
+        tud_cdc_read(cmdbuf, sizeof(cmdbuf));
+
+        if (cmdbuf[0] == 0x30) {
+            ctx.opcode = cmdbuf[1];
+            ctx.length = *((uint32_t*)(cmdbuf+2)); 
+            ctx.length = *((uint32_t*)(cmdbuf+2));
```
kwagyeman, line 168: Duplicate line?

common/usbdbg.c
```diff
@@ -317,93 +310,6 @@ void usbdbg_data_out(uint32_t size, usbdbg_read_callback_t read_callback) {
...
-            break;
-        }
-
-        case USBDBG_SET_TIME: {
-            // TODO implement
-            uint8_t buffer[32];
-            read_callback(&buffer, sizeof(buffer));
-            cmd = USBDBG_NONE;
-            break;
-        }
```
kwagyeman, line 400: I'd actually like to implement this feature in the future, where the IDE can set the RTC time. It's never been a high priority, though, over everything else. Can we leave it in? I don't have an ETA when I'll be able to work on it, though... which means it may be stuck in an unimplemented state forever...

common/usbdbg.c
```diff
@@ -317,93 +310,6 @@ void usbdbg_data_out(uint32_t size, usbdbg_read_callback_t read_callback) {
...
-
-        case USBDBG_TX_INPUT: {
```
kwagyeman, line 402: Customers have also asked for being able to type into the serial terminal in the IDE to send text to the camera forever. We should implement this one day too. However, it requires micropython to have some type of text input system... which means MP need to be updated to support such a thing... so, also... I guess no idea when this could work either without major effort.

common/usbdbg.c
kwagyeman: Note, the IDE currently send this command on connect since firmware v4.3.2. Removing it will mean that firmware will break with older IDEs...

common/usbdbg.c
kwagyeman: Note, the IDE currently sends this command on typing into the terminal since firmware v4.3.2...

common/tinyusb_debug.c
```diff
@@ -36,159 +37,177 @@
...
+                usbdbg_data_out(bytes, tud_cdc_read);
             }
         }
-    }
+        
+        if (bytes) {
+            ctx.timestamp = mp_hal_ticks_ms();
+        } else if (__get_PRIMASK() & 1) {
+            break;
+        } else if (check_timeout_ms(ctx.timestamp, USBDBG_DATA_TIMEOUT)) {
```
kwagyeman, line 205: Nice!

common/tinyusb_debug.c
```diff
@@ -36,159 +37,177 @@
 #include "usbdbg.h"
 #include "tinyusb_debug.h"
 #include "omv_common.h"
+#include "cmsis_gcc.h"
 
-#define USBDBG_BAUDRATE_SLOW     (921600)
-#define USBDBG_BAUDRATE_FAST     (12000000)
+#define USBDBG_DATA_TIMEOUT     (250)
```
kwagyeman: Set this to 2000-5000. The PC isn't real-time so it can randomly delay for a long time. 250 is in the range where disconnect issues will happen randomly.

common/tinyusb_debug.c
```diff
@@ -36,159 +37,177 @@
...
+#define USBDBG_MAX_RECOVERY     (2048)
```
kwagyeman: What's this for? It does't appear to be used.

common/tinyusb_debug.c
iabdalkader: Yes, good catch!.

common/tinyusb_debug.c
iabdalkader: It gets reset on every read/write in the loop, but will increase to 500ms.

common/tinyusb_debug.c
iabdalkader: Was going to add some recovery code, to search for a command byte, but it's useless. The protocol is just so bad it's hopeless to try to make it robust.

common/tinyusb_debug.c
kwagyeman: I'd be more conservative with this. Please make it a second at least.

common/usbdbg.c
iabdalkader: So we're dropping these old/unsupported commands. Feel free to add any, when you actually implement it.

common/tinyusb_debug.c
kwagyeman: Okay.

kwagyeman:
Please increase the timeout to 1 second. I've had endless issues with timeouts on the USB transport in the IDE when set under 1 second. It's not safe.

iabdalkader:
This is a different type of timeout, again it gets reset after every DATA packet, if it hits 500ms a problem has definitely occurred, but okay, increased to 1000ms.

PR #2829, opened 2025-09-05
sensors/GenX320: Adding script for bias tuning in event streaming mode.

nlyubova-psee:
Adds a simple script for bias setting for GenX320 sensor when streaming in event mode

### DIFF
--- a/scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_biases.py
+++ b/scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_biases.py
@@ -1,0 +1,63 @@
+# This work is licensed under the MIT license.
+# Copyright (c) 2013-2025 OpenMV LLC. All rights reserved.
+# https://github.com/openmv/openmv/blob/master/LICENSE
+#
+# This example shows off using the genx320 event camera from Prophesee
+# using event streaming mode and adjusting the sensor settings such as
+# * biases values
+# * AFK filter
+
+import csi
+import image
+import time
+# https://micropython-ulab.readthedocs.io/en/latest/index.html
+from ulab import numpy as np
+
+# Surface to draw the histogram image on.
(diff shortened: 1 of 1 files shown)


kwagyeman:
Should the bias settings be set to something other than the default? Or maybe list the options in the script? Otherwise, looks fine.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_biases.py
```diff
@@ -0,0 +1,63 @@
...
+#   [5]  Y coordinate 0 to 319 for GENX320
+events = np.zeros((2048, 6), dtype=np.uint16)
+
+# Initialize the sensor.
+csi0 = csi.CSI(cid=csi.GENX320)
+csi0.reset()
+csi0.ioctl(csi.IOCTL_GENX320_SET_MODE, csi.GENX320_MODE_EVENT, events.shape[0])
+
+# Set the sensor biases to their default settings
+csi0.ioctl(csi.IOCTL_GENX320_SET_BIASES, csi.GENX320_BIASES_DEFAULT)
```
kwagyeman: List the options above in a comment and what they do?

nlyubova-psee:
> Should the bias settings be set to something other than the default? Or maybe list the options in the script? Otherwise, looks fine.

The "default" biases here correspond to GenX320 sensor specification and not the default in this driver. Maybe the "default" here is not the best name ... Anyway, it is useful to set the sensor to the settings that correspond to the sensor specification.

scripts/examples/01-Camera/03-Event-Cameras/02-Genx320/genx320_event_mode_biases.py
nlyubova-psee: added the comment.

kwagyeman (approved):

PR #2833, opened 2025-09-11
 lib/imlib: Add an embedded frame header in the streaming buffer. 

iabdalkader:
For use with the new protocol, but there's no reason not to add it now.

### DIFF
--- a/lib/imlib/framebuffer.c
+++ b/lib/imlib/framebuffer.c
@@ -69,6 +69,11 @@ void framebuffer_init(framebuffer_t *fb, void *buff, size_t size, bool dynamic, 
     #if OMV_RAW_PREVIEW_ENABLE
     fb->raw_w = OMV_RAW_PREVIEW_WIDTH;
     fb->raw_h = OMV_RAW_PREVIEW_HEIGHT;
+    fb->raw_enabled = true;
+    #else
+    fb->raw_w = 0;
+    fb->raw_h = 0;
+    fb->raw_enabled = false;
     #endif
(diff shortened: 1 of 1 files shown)


lib/imlib/framebuffer.c
```diff
@@ -315,14 +316,12 @@ void framebuffer_update_preview(image_t *src) {
                 imlib_draw_image(&dst, src, 0, 0, scale, scale, NULL, -1, 255, NULL, NULL,
                                  IMAGE_HINT_BILINEAR | IMAGE_HINT_BLACK_BACKGROUND,
                                  NULL, NULL, NULL, NULL);
-                compress = false;
             }
```
kwagyeman, line 321:
```
else {
    raw_stream = false
}
```

If it doesn't fit still you want to fall back to compression. That said, very unlikely, if at all, to happen.

iabdalkader: You won't be able to set a raw frame size bigger than the max framebuffer size.

iabdalkader: That second check is redundant, if you trust the math not to have rounding errors.

iabdalkader: On second thought, it's probably better to delay this check until the very last moment when we have the actual BPP. Fixed.

kwagyeman: Yeah, it's just like fixing a dangling bug so it won't bite you later for some random reason.

kwagyeman:
LGTM

PR #2840, opened 2025-09-16
stm32: Disable clock gating during sleep for STM32N6 peripherals

iabdalkader:
Remove individual peripheral sleep clock enable/disable calls and instead enable all AHB/APB peripheral clocks during sleep mode. This is needed because MicroPython doesn't manage peripheral sleep clocks individually.


iabdalkader:
FYI @kwagyeman It doesn't seem to affect power consumption.

PR #2861, opened 2025-09-25
ports/stm32: Enable RIF settings for N6 ADC12.

kwagyeman:
This commit is needed to make the ADC work from: https://github.com/micropython/micropython/pull/18009


iabdalkader:
Fixed it here: https://github.com/openmv/openmv/pull/2865
I needed to also fix a couple of minor issues with that table.

PR #2867, opened 2025-10-01
lib/micropython: Update submodule.

iabdalkader:
- stm32/adc: Get ADC working on STM32N6 MCUs.
- alif/machine_spi: Fix init() to only change requested settings.
- alif/boards/OPENMV_AE3: Fix switch name to match OpenMV RT1062 and N6.
- stm32/boards/OPENMV_N6: Define all OpenMV N6 pins.
- stm32/boards: Add PG13 AF7 (USART3_RTS) to N6 AF file.
- stm32/boards/OPENMV_N6: Define RTS pin for UART3.


iabdalkader:
@kwagyeman FYI.

kwagyeman:
Thank you!

PR #2868, opened 2025-10-02
ports/unix: Add Unix port support for desktop development.

andrewleech:
## Summary

Adds Unix port support to OpenMV, enabling desktop development and testing without hardware. This implementation includes VFS abstraction to replace FatFS dependencies and Unix-specific compatibility layer.

### 1. VFS Abstraction (Commit 1: fe3a1414)

Replaced FatFS direct calls with MicroPython VFS abstraction to support multiple filesystem backends:

**New files:**
- `common/file_utils.c` - VFS wrapper functions
- `common/file_utils.h` - VFS API definitions

**Type conversions:**
- `FIL` → `file_t` (VFS file handle wrapper)
- `FRESULT` → exception-based error handling
- `FR_OK` checks → simplified (VFS raises exceptions)

**Function conversions:**
- `f_size()` → `file_size()`
- `f_tell()` → `file_tell()`
- `f_open()` → `file_open()`
- `f_close()` → `file_close()`
- `f_read()` → `file_read()`
- `f_write()` → `file_write()`

**Files modified:**
- `modules/`: py_gif.c, py_mjpeg.c, py_helper.c, py_image.c
- `lib/imlib/`: bmp.c, imlib.c, jpege.c, mjpeg.c, png.c, ppm.c, orb.c, lbp.c
- `boards/UNIX/imlib_config.h`: Fixed define name to IMLIB_ENABLE_IMAGE_FILE_IO
- `modules/micropython.mk`: Added file_utils.c to Unix port build

### 2. Unix Port Support (Commit 2: 45a0566e)

**New directory structure:**
```
boards/UNIX/               - Unix variant configuration
ports/unix/               - Unix-specific implementations
  unix_compat/            - Compatibility headers
    ff.h                  - FatFS compatibility (deprecated/removed after VFS)
    fmath.h               - Fast math functions
    mp_compat.h           - MicroPython compatibility
    omv_force_include.h   - Force-included definitions
```

**Configuration files:**
- `boards/UNIX/mpconfigvariant.h` - MicroPython feature flags
- `boards/UNIX/mpconfigvariant.mk` - Build configuration
- `boards/UNIX/omv_boardconfig.h` - OpenMV board settings
- `boards/UNIX/omv_boardconfig.mk` - OpenMV feature flags
- `boards/UNIX/imlib_config.h` - Image library configuration
- `boards/UNIX/manifest.py` - Frozen modules

**Unix implementations:**
- `ports/unix/omv_gpio.c` - GPIO stubs
- `ports/unix/omv_i2c.c` - I2C stubs
- `ports/unix/omv_spi.c` - SPI stubs
- `ports/unix/sensor.c` - Sensor stubs
- `ports/unix/fb_alloc.c` - Framebuffer allocator

**Build system:**
- `Makefile`: Added `unix` target for building Unix port
- `modules/micropython.mk`: Unix-specific source files and compiler flags

## Features

**Enabled on Unix port:**
- Image module with full file I/O support (BMP, PNG, JPEG, PPM)
- Image processing algorithms (filters, drawing, geometry, etc.)
- Feature detection (ORB, LBP, blobs, lines, etc.)
- QR code, barcode, AprilTag detection
- Color space operations
- Statistical operations

**Disabled on Unix port:**
- Hardware modules (CSI, FIR, IMU, TOF, Audio, Display, TV)
- Machine learning (TensorFlow Lite, STM32 Cube.AI)
- Sensor capture (uses stub implementation)

### Build Testing
```bash
cd lib/openmv
make unix
```
**Result:** Build completes successfully, binary at `lib/micropython/ports/unix/build-openmv/micropython`

### File I/O Testing
```python
import image

# Create 64x64 grayscale image with checkerboard pattern
img = image.Image(64, 64, image.GRAYSCALE)
for y in range(64):
    for x in range(64):
        if (x // 8 + y // 8) % 2:
            img.set_pixel(x, y, 255)
        else:
            img.set_pixel(x, y, 0)

# Save to BMP file
img.save('/tmp/test_pattern.bmp')

# Load from file
img2 = image.Image('/tmp/test_pattern.bmp')

# Verify all 4096 pixels match
for y in range(64):
    for x in range(64):
        expected = 255 if (x // 8 + y // 8) % 2 else 0
        assert img2.get_pixel(x, y) == expected
```
**Result:** SUCCESS - All 4096 pixels verified correctly

### Module Import Testing
```python
import image
print(dir(image))  # Verify all image functions available
```
**Result:** All image module functions accessible

## Build Instructions

```bash
# Clone repository
git clone https://github.com/openmv/openmv.git
cd openmv

# Checkout Unix port branch
git checkout unix-port-support

# Build Unix port
make unix

# Run
cd lib/micropython/ports/unix
./build-openmv/micropython
```

## Use Cases

- Desktop development and testing without OpenMV hardware
- CI/CD integration for automated testing
- Algorithm development and validation
- Image processing pipeline prototyping
- Educational environments without hardware requirements

## Future Work

- Additional testing with various image formats
- Performance benchmarking vs embedded targets
- Integration with existing test suites
- Documentation for Unix port usage


kwagyeman:
Hi Andrew, thanks for this effort!

I took a quick look over the files, and you'll need to try to use exist code as much as possible versus creating shim layers for specific files. For example, with fb_alloc... if you define the linker script variables, a copy of that isn't necessary, same for the frame buffer.

Also, for files with parts that only compile for ARM that need fixes, it's better to make a commit with ifdefs on those files to add the necessary functionality to fallback to the standard C implementation. E.g. for the fast functions.

ndarray support can't be trivially turned off too as that will become more central to the code...

iabdalkader:
I like the first commit (VFS Abstraction) regardless of the Unix port. So maybe we should break this out in a separate PR and merge it first?

> Disabled on Unix port:
>    Hardware modules (CSI, FIR, IMU, TOF, Audio, Display, TV)

I could be missing something, but you might be able to get CSI working if you just enable SoftCSI (see drivers/sensors/softcsi.c), and emulate I2C scan to always return SoftCSI address. This should pull in the rest of the components (framebuffer, queues, fb_alloc etc..). You could take this a step further and define an mjpeg or mpeg file that softcsi could play in a loop (Oh! Since you're on Unix, you might actually be able to read from a real UVC device!).

> CI/CD integration for automated testing

It's very unlikely for the Unix port to be useful for CI, especially that things which typically break will either be disabled, unsupported or replaced by dummy implementation. For CI pipeline, I was thinking about using either QEMU or ARM FVP. I spent some time trying to understand if there's some FVP licensing required for CI usage, but I couldn't. QEMU also looks very promising and it even supports MVE, but FVP would let us emulate Ethos-U.

andrewleech:
> you'll need to try to use exist code as much as possible versus creating shim layers for specific files. For example, with fb_alloc... if you define the linker script variables, a copy of that isn't necessary, same for the frame buffer.

Yeah sorry for the noise, I hadn't meant to raise the PR quite so soon, I intended to continue to clean it up like this. I progressed it quite a lot in this direction today and it's getting closer, but not enough to take it out of draft just yet.

> Also, for files with parts that only compile for ARM that need fixes, it's better to make a commit with ifdefs on those files to add the necessary functionality to fallback to the standard C implementation. E.g. for the fast functions.

Sure thanks. I initially wasn't sure how much conditional support was going to be needed and whether it would be too invasive to have scattered throughout the codebase or better to keep the Unix support in its own area.

> ndarray support can't be trivially turned off too as that will become more central to the code...

Yep definitely didn't mean that to get stubbed out, the submodules weren't updated properly!

andrewleech:
> I like the first commit (VFS Abstraction) regardless of the Unix port. So maybe we should break this out in a separate PR and merge it first?

Good point! This started out as an effort to get a Unix build going again for desktop testing / reprocessing.  While looking at the initial compatibility issues (eg fatfs) I remembered https://github.com/openmv/openmv/pull/400 and rebuilt that.

andrewleech:
> You could take this a step further and define an mjpeg or mpeg file that softcsi could play in a loop.

That definitely adds value particularly for application testing!

> It's very unlikely for the Unix port to be useful for CI, especially that things which typically break will either be disabled, unsupported or replaced by dummy implementation. 

In my past applications using openmv I actually found this very useful in CI to ensure that various steps of the image processing pipeline remained consistent; for a library of known snapshot images I could ensure the correct results are always returned. 

But yes we found it even more beneficial for the image processing pipeline development; capture a set of representative images then can quickly and reliably reprocess them and iterate filter development etc. At one point I had a Jupiter notebook kernel wrapper that let me use the Unix openmv micropython in the notebook for some cells then plot with regular python in others.

iabdalkader:
> > You could take this a step further and define an mjpeg or mpeg file that softcsi could play in a loop.
> 
> That definitely adds value particularly for unit testing!

Since you'll be running on Unix, you could even read the real webcam/uvc. Depending on how much time you're willing to spend on this, you could add a new sensor driver just for Unix (maybe `sensors/uvc.c`) that returns real frames. Note: if you use the new CSI API, you can enable both soft-csi and uvc.

andrewleech:
> > I like the first commit (VFS Abstraction) regardless of the Unix port. So maybe we should break this out in a separate PR and merge it first?
> 
> Good point! This started out as an effort to get a Unix build going again for desktop testing / reprocessing. While looking at the initial compatibility issues (eg fatfs) I remembered #400 and rebuilt that.

Split out to #2870 with some further cleanups, I'll rebase this on top later.

kwagyeman:
@andrewleech - Rebase this?

andrewleech:
Thanks yes I have done locally, been trying to test it a bit more thoroughly before pushing it up again 😁

andrewleech:
I've just pushed a significant refactor addressing review feedback and extending feature coverage. 
Branch now rebased on latest master. 
                                                                                                                                                               
Key Changes From Original PR                                                                                                                                   
                                                                                                                                                               
1. Removed Unnecessary Shim Layers (per @kwagyeman feedback)                                                                                                   
- Deleted ports/unix/fb_alloc.c - now uses common fb_alloc() with GC-based allocation                                                                       
- Deleted ports/unix/omv_gpio.c - GPIO stubs unnecessary (all calls conditionally compiled)                                                                 
- Removed FatFS compatibility shim - VFS now used.
                                                                                                                                                               
2. Added SoftCSI Virtual Camera (per @iabdalkader suggestion)                                                                                                  
- Full sensor module support via drivers/sensors/softcsi.c                                                                                                  
- Synthetic test patterns (checkerboard, gradients) with animation                                                                                          
- Load real images via Python API: sensor.set_source([paths/images])                                                                                        
- Format conversion, scaling, flip/mirror transformations                                                                                                   
- Proper GC safety using static globals
- Not yet connected through to real camera infrastructure, could be added in future (my dev machine doesn't have a camera currently :-) )
                                                                                                                                                               
3. Test Infrastructure                                                                                                                                         
- Official test suite at scripts/unittest/ (82 tests)                                                                                                       
- 78/82 tests passing (95.1% pass rate)                                                                                                                     
- 0 failures, 4 hardware-only tests skipped                                                                                                                 
- CI/CD workflow via GitHub Actions (.github/workflows/unix-port.yml)                                                                                       
                                                                                                                                                               
4. Clean Architecture                                                                                                                                          
- ARM intrinsic compatibility headers in ports/unix/unix_compat/                                                                                            
- Weak symbol overrides for platform differences                                                                                                     
- Uses common CSI infrastructure (common/omv_csi.c)                                                                                                         
- I2C stubs provided for CSI init
                                                                                                                                                            
Working features:                                                                                                                                              
- QR codes, barcodes, AprilTags, data matrices                                                                                                                 
- Blob/line/circle/rectangle detection                                                                                                                         
- Haarcascade, template matching, selective search                                                                                                             
- ORB keypoints, LBP/HOG descriptors                                                                                                                           
- All filters, morphological ops, color conversions                                                                                                            
- Image I/O (BMP, JPEG, PNG, PPM, PGM)                                                                                                                         
- Sensor capture via SoftCSI                                                                                                                                   
                                                                                                                                                               
Skipped (expected):                                                                                                                                            
- Hardware-specific: gamma_correction, get_similarity, stereo_disparity                                                                                        
- ML: tensorflow (module unavailable on Unix)                                                                                                                  
                                                                                                                                                               
Build Size                                                                                                                                                     
                                                                                                                                                               
- Binary: 5.5 MB (1.7 MB text, 106 KB data, 3.7 MB bss)                                                                                                        
- Location: lib/micropython/ports/unix/build-openmv/micropython                                                                                                

---                                                                                                                                                            
Ready for review. All original feedback addressed, test coverage verified, architecture simplified.

common/fb_alloc.c
```diff
@@ -30,8 +30,16 @@
 #include "omv_boardconfig.h"
 #include "omv_common.h"
 
+#if defined(__unix__)
+// On Unix, we can't use linker script symbols, so we use an externally provided pointer
```
iabdalkader: `uintptr_t` aside, No changes should be needed to this file if you just call it `_fb_alloc_end`.

modules/py_csi.c
```diff
@@ -827,6 +827,35 @@ static mp_obj_t py_omv_csi_set_frame_callback(mp_obj_t frame_callback_obj) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_1(py_omv_csi_set_frame_callback_obj, py_omv_csi_set_frame_callback);
 
+#if (OMV_SOFTCSI_ENABLE == 1)
+// Include softcsi_set_source declaration
+extern int softcsi_set_source(omv_csi_t *csi, mp_obj_t source);
+
+static mp_obj_t py_omv_csi_set_source(size_t n_args, const mp_obj_t *args) {
```
iabdalkader: This is not needed, because you can already load images from files into FBs directly, with `copy_to_fb=True`: https://docs.openmv.io/library/omv.image.html#image.Image

It should work the same regardless of the sensor (SoftCSI or otherwise), I haven't tested it though.

common/fb_alloc.c
andrewleech: I've tried playing around with a few ways to handle this with a statically allocated array/struct; `_fb_alloc_end` in embedded is a symbol defined at the end / bottom of the memory region such that the address of it is the bottom of the array. However in C I can't place a symbol at the end, the best I can do is calculate "start + length" to have the end address as a value, but can't place a symbol there to & the address from. Not sure if I've mis-interpreted your suggestion though?

modules/py_csi.c
andrewleech: The sensor / csi stuff ended up being a larger change that I haven't reviewed very well, maybe it'd be better to split it into a separate PR?

My goal with the "fake `sensor` " is to be able to give it an iterable (list etc) of Image and/or file paths that then the sensor would work through each time you call snapshot. I believe I'm already using the existing underlying functionality (in C) from Image's copy_to_fb=True to handle file paths and load them into the snapshot's newly allocated Image object when it gets to a file path. 
I didn't want to have to pre-load them all into image objects beforehand in case you have a long list of images you want to be able to run through. I didn't see any other way to feed a list of images to be used / provided by the fake sensor, but I absolutely may have mis-interpreted how SoftCSI is used here?

kwagyeman: You just do something like this in your code:

```
stuff = ["filename.jpg"]
for s in stuff:
    img = image.Image(s, copy_to_fb=True)
    <do things with the image>
```

Or you can use the ImageIO object which is much faster for this. See the read example under Examples -> Camera - Video Recording.

andrewleech: Oh, my intention for this unix port `sensor` is to be able to use the unix port to unit test applications written for use on a real device, so being able to inject "real" images into the fake sensor means entire imaging pipelines can be tested repeatably.

That being said, yeah maybe this is overcomplicating things. If `sensor` wasn't provided by the unix port we could simply provide a python mock sensor in our ci tests that matched the api and returned the image object, without needing to go through all the C layers.

iabdalkader: > so being able to inject "real" images into the fake sensor

That's what the `copy_to_fb` arg is for: copying/loading some image into the frame buffer. Once an image is set as the frame buffer, the next call runs on that as if it captured that image. Note you could also create images from bytes/ndarray, and many functions, other than the image.Image(), accept this arg, like color conversion, compression, scaling etc..

andrewleech: Sure yes, I thought I understand the `copy_to_fb` arg on the python functions, but that's to copy one source frame into the framebuffer and return / use that wrapped in Image object isn't it?

I was trying to set up a way for multiple calls to sensor.snapshot() to return frames one after another from a pre-configured stream of frames. 
I couldn't see / think of a different way to feed a sequence of frames into the softcsi module.

Though yeah now I'm not sure if there's enough value in providing a C version of this...

iabdalkader: > but that's to copy one source frame into the framebuffer and return / use that wrapped in Image object isn't it?

Yes, and that's the same as calling `snapshot()`: some new image will be in the frame buffer, and you have a "pointer" to that frame buffer.

> for multiple calls to sensor.snapshot() to return frames one after another from a pre-configured stream of frames

As mentioned, you can loop over a file list or use ImageIO. Note the IDE has ffmpeg builtin and convert video files to `.bin`, then you can playback that in loop:
```Python
img = stream.read(copy_to_fb=True, loop=True, pause=True)
```
This is exactly why we made these things all those years ago, for testing stuff.

andrewleech: Yep I'll pull out all the C "fake  sensor" stuff here, it's easier to mock in python for test use. 

I've just pulled up my historic project based on openmv and yes I'd mocked `sensor` module there in python basically doing exactly this which was in the unit/integration test import path so the application code that's expecting to use `sensor` finds and uses the mock just fine. That's much simpler and more flexible than having a C impl here.

kwagyeman:
@andrewleech - If you want this in the v5.0.0 release, you need to finish the PR now.

andrewleech:
Hi @kwagyeman, sorry I went quiet for so long, other work priorities got in the way. After reviewing the substantial changes you've made on master since I last looked, I decided to largely rewrite what I had here rather than try to rebase it. The old branch wasn't really rebaseable anyway with UMA replacing fb_alloc / dma_alloc, the libmicropython.a refactor, the board / port config file renames etc, so a fresh build against current master made more sense.

Branch is now 7 commits, one per area:

```
boards/UNIX:                 board configuration for the Unix port
ports/unix:                  Unix port implementation (UMA pool init, FB init, ARM intrinsic shims)
Makefile:                    unix target for host-side build
modules/micropython.mk:      Unix port build configuration
scripts/unittest:            Unix port test infrastructure
ci:                          GitHub Actions workflow
docs:                        Unix port documentation
```

The approach is pretty different / much simpler than what I had previously:

- No fb_alloc shim, no omv_gpio shim, no separate copy of the framebuffer. A single host-malloc'd UMA pool (default 64MB, overridable at compile time) is registered with flags=0; UMA's fallback selects it for any allocation regardless of attribute, so embedded code that asks for UMA_FAST / UMA_DTCM / UMA_TRANSIENT etc just works.
- Uses MicroPython's own unix variant mechanism rather than building a parallel pipeline. `make unix` invokes `lib/micropython/ports/unix/Makefile` with `VARIANT=openmv`, `VARIANT_DIR=$(pwd)/boards/UNIX`, `USER_C_MODULES=$(pwd)`. The embedded `port_config.mk` / `board_config.mk` includes are skipped for unix via `ifeq ($(filter unix,$(MAKECMDGOALS)),)` guards in the top level Makefile.
- ARM intrinsics on non-ARM hosts handled by three force-included compat shims in `ports/unix/unix_compat/` rather than scattering ifdefs through imlib.
- A GCC constructor in `py_unix_stubs.c` runs UMA / framebuffer / imlib init before `mp_init()`. It also runs a sanity check on `__USAT` / `__USAT_ASR` at startup; the latter is pretty easy to write with the argument order swapped which silently produces 8x wrong filter output, so the trap catches that class of regression before any test touches imlib.
- One real imlib bug fell out along the way; `lib/imlib/imlib.c` does `char magic[4]; magic[0] == 0xFF;` for file magic dispatch which works on ARM (unsigned char by default) but fails on x86 (signed char by default), broke JPEG and PNG load. Fixed in the unix build by adding `-funsigned-char` to `boards/UNIX/mpconfigvariant.mk` so the host matches ARM. No imlib code touched.

Unit tests now share one runner across both targets. The existing `scripts/unittest/run.py` had its three hardcoded paths made overridable via sys.argv, then a ~50 line host wrapper at `scripts/unittest/run_tests.py` invokes the unix binary on it. Same script drives hardware via mpremote and unix via subprocess.

I dropped SoftCSI from this version to keep the diff minimal, can add it back as a follow up if you'd like it.

Test results on x86_64 Linux: 77 pass, 0 fail, 15 skipped (the hardware-only ones, capture / gamma_correction / get_similarity / stereo_disparity / ml_* / tensorflow). CI runs the smoke test then the full suite via the wrapper.

Should be ready for review now.

iabdalkader:
> lib/imlib/imlib.c does char magic[4]; magic[0] == 0xFF

We should just use `uint8_t` here (in separate PR) if it makes it easier.

.github/workflows/unix-port.yml
```diff
@@ -0,0 +1,69 @@
+name: '🐧 Unix Port Build'
```
iabdalkader: Is there any reason this can't be part of the main firmware workflow? Shouldn't `make TARGET=UNIX` just work like any other board?

andrewleech:
Hi @iabdalkader, thanks for both. PR #3155 has merged so I've rebased on top of that and dropped the `-funsigned-char` flag from the unix variant. I also pulled `make TARGET=UNIX` into the standard build pipeline so the unix port flows through `boards/UNIX/board_config.mk` -> `ports/unix/port_config.mk` like any other board, and folded the unix CI matrix entry into `.github/workflows/firmware.yml`. The standalone `unix-port.yml` is gone. `make unix` no longer works on its own, you have to use `make TARGET=UNIX`.

While I was at it I ran a multi-agent review over the branch and addressed the findings. The main ones worth flagging:

- `boards/UNIX/imlib_config.h` was using `#define IMLIB_ENABLE_X (1)` everywhere, but the codebase tests these via `#ifdef`, so anyone setting one to `(0)` to disable a feature would have seen it stay enabled silently. Switched to bare `#define IMLIB_ENABLE_X` matching `boards/OPENMV4/imlib_config.h` convention; commenting out the line is how you disable.
- The `MICROPY_PY_*` flag list in `boards/UNIX/board_config.mk` was largely dead, only `MICROPY_PY_CRC` and `MICROPY_PY_UMALLOC` were being propagated to defines. Extended the conversion list in `modules/micropython.mk` to mirror the set in `common/micropy.mk` so BTREE / SSL / ULAB / etc all actually take effect now.
- `ports/unix/py_clock.h` was a leftover empty stub that was actually shadowing the real `modules/py_clock.h` because `-I$(OMV_PORT_DIR)` was listed before `-I$(OMV_MOD_DIR)` in `CFLAGS_USERMOD`. Deleted the stub and re-ordered the includes.
- Dropped a couple of redundant compat shims (`omv_common.h` wrapper that duplicated `OMV_ALIGN_TO`, the `cmsis_extension.h` macros that already live in `omv_force_include.h`).
- Hardened the GCC constructor in `py_unix_stubs.c`, replaced `abort()` calls with `_exit(1)` for cleaner CI failure surface, added a matching destructor that frees the UMA pool so dlopen / mp_init / mp_deinit cycles don't leak, and tightened the `__USAT` parameter validation to match CMSIS spec.

Public-facing behaviour is unchanged: 7 commits, build clean via `make TARGET=UNIX`, 77 tests pass / 0 fail / 15 skipped (the hardware-only ones). CI runs as the `UNIX` matrix entry alongside the other boards.

kwagyeman:
@andrewleech - You left the PR in draft state. What is remaining from your side?

andrewleech:
Ah sorry I forgot it was still draft - after the last rework I'm pretty happy with it thanks!

PR #2874, opened 2025-10-05
scripts/examples: Add yolov2 and yolov5 template examples.

kwagyeman:
Removes the very bad yolov5 model and the outdated yolov2 model. Adds example scripts for the curious to use if they want to run the networks.


kwagyeman:
Updated

kwagyeman:
Updated to master. Restested YOLOV2 and YOLOV5 on AE3.

PR #2876, opened 2025-10-05
scripts/examples: Add depth estimation example.

kwagyeman:
Depth Estimation. Kinda works. Needs to fix a few features in the firmware before this ready.

https://github.com/user-attachments/assets/a422f06b-5448-4bd4-9989-761eb15572b4


kwagyeman:
Updated with depth re-coloring. Need to fix the AE3 GPU before mergable.

PR #2879, opened 2025-10-05
drivers/boson: Use timespec struct from header.

kwagyeman:
Fixes compile issue with https://github.com/openmv/openmv/pull/2870.

### DIFF
--- a/drivers/boson/include/timeoutLogic.h
+++ b/drivers/boson/include/timeoutLogic.h
@@ -14,9 +14,9 @@
 
 #ifndef _TIMEOUT_LOGIC_H
 #define _TIMEOUT_LOGIC_H
+#include <sys/timespec.h>
 
 #define CLOCK_MONOTONIC 0
-struct timespec { long tv_sec; long tv_nsec; };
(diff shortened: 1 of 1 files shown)


iabdalkader:
Is it okay to use the standard struct? `tv_sec` may be 64 bits.

kwagyeman:
Yes, it's fine. Tested and works, also, the implementation of the logic is this:

```
int clock_gettime(int _ignore, struct timespec *spec)
{
    uint32_t time_ms = mp_hal_ticks_ms();
    spec->tv_sec  = time_ms / 1000L;
    spec->tv_nsec = time_ms % 1000L * 1000000;
    return 0;
}

double difftime(long current, long reference)
{
    return ((double)current - (double)reference);
}

double diff_timespec(struct timespec *current, struct timespec *reference)
{
    double elapsed_sec = difftime(current->tv_sec, reference->tv_sec);
    elapsed_sec += (((double)current->tv_nsec) - ((double)reference->tv_nsec))/1000000000;
    return elapsed_sec;
}
```

So, as long as it's 32-bit at least, there's no issue.

PR #2883, opened 2025-10-08
drivers/sensors: add I3C support for the VD55G1 image sensor

leonardojc90:
Not sure why the OpenMV repo keeps a separate copy of the Alif CMSIS DFP, instead of using the one in the MicroPython submodule (`lib/micropython/lib/alif_ensemble-cmsis-dfp`), but it is relatively outdated and I'm looking forward to use features of a newer version in order to add new functionality to OpenMV

I'm not going straight for updating `lib/alif` to version 2.0.X since there are to many breaking changes that require adapting other parts of the code, but that would to the job for now

I wonder whether it would be better to simply modify the build system in order to link against the SDK libraries shipped with MicroPython, so that OpenMV could get automatically in sync

Edit: I have reduced the scope of the changes to address only the I3C bits


kwagyeman:
Hi, you've got a lot of formatting changes with tabs to spaces in the PR. Remove these. Any PR sent should be a clean diff. Also, it's not so easy to just update the DFP. Doing so will just potentially break everything. We keep a separate copy of the dfp as ours may have modifications in it to fix bugs.

> but it is relatively outdated and I'm looking forward to use features of a newer version in order to add new functionality to OpenMV

What are the newer features you are looking to add?

leonardojc90:
> Hi, you've got a lot of formatting changes with tabs to spaces in the PR. Remove these. Any PR sent should be a clean diff

Sure. I have mostly copied the files from the 1.3.4 release, but expected that some adjustments would be requested

> Also, it's not so easy to just update the DFP. Doing so will just potentially break everything

Which tests would you like to see done successfully in order to consider such an update?

> What are the newer features you are looking to add?

For now, adding I3C implementations for platforms that support it, in order to create drivers to newer camera sensors. The I3C library for the current version of the DFP is lacking some useful functions

I thought about porting just the bare minimum for better I3C support, but that would increase the diff from vanilla DFP releases

kwagyeman:
Need to defer to Ibrahim on tests and etc. for what to do. Probably the bare minimum updates to I3C would be better though as we don't use that in the driver stack... so, it wouldn't break anything at all to update just that.

leonardojc90:
I have reduced the scope of the changes to address only the I3C bits

Based on the e-mail exchange with Jonathan, I'm keeping it as a draft for comments until a proper `omv_i3c` interface is defined by upstream

kwagyeman:
Say the VD55G1 code! When ready for review send me hardware to test with!

iabdalkader:
> I wonder whether it would be better to simply modify the build system in order to link against the SDK libraries shipped with MicroPython, so that OpenMV could get automatically in sync

MicroPython uses the upstream dfp as a submodule. We have custom patches and extra drivers that don't exist there.

iabdalkader:
I'm updating DFP to 1.3.4, see https://github.com/openmv/openmv/pull/2926. I cherry-picked your fixes to `omv_i2c.c`. However, you need to [sign the CLA](https://github.com/openmv/openmv/pull/2926#issuecomment-3619763861) first before I can merge it.

leonardojc90:
Done

Sorry for the delay in finishing the VD55G1 driver. I was stuck with the 1.8v bus not working (even for I2C Fast Mode+) and had to move to different tasks here. But I'll get back to it

iabdalkader:
Thanks! I merged the update to 1.3.4. You may want to revert changes to `omv_i2c.c` before rebasing on master. I see you already did that, so you just need to rebase this.

iabdalkader:
Let us know when this is ready.

kwagyeman:
@leonardojc90 - What's the status on this? The v5.0.0 release is nearing.

leonardojc90:
> @leonardojc90 - What's the status on this? The v5.0.0 release is nearing.

Sorry for the late reply

All that I've wished to implement is done, but due to higher priority tasks here I couldn't dedicate time to testing here yet

iabdalkader:
> All that I've wished to implement is done, but due to higher priority tasks here I couldn't dedicate time to testing here yet

@leonardojc90 No worries, take your time. It's probably best if this waits after 5.0.0 anyway so we can review and test it.

PR #2888, opened 2025-10-12
modules/py_imageio: Fix stream file mode.

iabdalkader:
Open for write should truncate the file if exists.


kwagyeman:
Yeah, that code was emulating an append mode. You might want to add one. I don't use this feature much, so I don't know what demand there is for appending, though. So, okay if you don't.

PR #2895, opened 2025-10-16
docker: Add git worktree support

andrewleech:
## Changes

This PR modifies the Docker build system to support git worktrees by mounting source code at runtime instead of copying it during image build.

## Technical Details

**Previous approach:**
- `COPY . .` embedded source tree into image during `docker build`
- Source was static - required image rebuild for any changes
- Failed with worktrees (`.git` is a pointer, main repo not accessible during build)

**New approach:**
- Source mounted as volumes at runtime: `-v /path/to/repo:/path/to/repo`
- Working directory matches host paths (makes elf file easier to debug from host computer)
- Git worktrees work (both worktree and main repo `.git` mounted)

**Benefits:**
- Faster iteration (no image rebuild needed)
- More reusable image
- Live source updates
- Supports git worktree workflows

## Testing

Verified on git worktree setup:
```bash
cd docker && make TARGET=OPENMV4
```

Build completes successfully with firmware output at `docker/build/OPENMV4/bin/`.

## Open Question

Was the original `COPY . .` approach intentional for other benefits (reproducibility, isolation, CI/CD workflows, etc.)?

Is there a different reason you'd prefer to keep the in-container build in a /workdir path?

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -51,4 +51,100 @@ shell:
         -v $(realpath ../):/workspace \
         $(DOCKER_IMAGE_NAME):$(DOCKER_TAG) bash
 
+# ----------------------------------------------------------------------------
+# Local development targets.
+#
+# These mirror build-firmware / shell but mount the repo (and SDK) at the
+# same absolute path inside the container as on the host. This keeps debug
+# info paths in build artifacts matching the host filesystem (so gdb,
+# addr2line and IDE source navigation work without remapping). Worktrees
+# are supported by also mounting the main repo so the worktree's .git
+# pointer file resolves.
+#
+# Unlike the default targets, build-firmware-dev does not run `make clean`
+# before building, so subsequent invocations are incremental.
+# ----------------------------------------------------------------------------
+
+# Repo root and git layout, resolved on the host.
+DEV_REPO_ROOT := $(realpath ../)
+DEV_GIT_DIR := $(realpath $(shell git -C .. rev-parse --git-dir 2>/dev/null))
+DEV_GIT_COMMON_DIR := $(realpath $(shell git -C .. rev-parse --git-common-dir 2>/dev/null))
+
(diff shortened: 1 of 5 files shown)


iabdalkader:
> Was the original `COPY . .` approach intentional for other benefits (reproducibility, isolation, CI/CD workflows, etc.)?

Yes, mainly isolation. I understand copy is very slow and requires plenty of space, but when running locally, I noticed that mounting the workspace left build artifacts outside the container, patched local files etc.. I don't think CI would be affected by this.

iabdalkader:
> Faster iteration (no image rebuild needed)

FWIW, I tested this and if I make a change to one file, it rebuilds everything from scratch.

kwagyeman:
@andrewleech - Do you plan to fix this PR?

andrewleech:
Sorry for not following this up sooner @kwagyeman. I've completely rebuilt this on current master to support the incremental developer build I was originally aiming for, without breaking the existing `build-firmware` behaviour.

The original commit wasn't really rebaseable - master moved past `COPY . .` already so the PR's initial premise didn't apply. Force-pushed as a fresh single commit instead.

What's there now: `build-firmware` and `shell` are unchanged, with new `build-firmware-dev` / `shell-dev` / `clean-dev` / `install-sdk` targets alongside for local iteration. The dev variant mounts the repo and SDK at matching host paths so debug info in `firmware.elf` resolves against the host filesystem, detects worktrees and mounts the main repo for the `.git` pointer, runs as the host user so artifacts are host-owned, and skips the up-front `make clean` so rebuilds are incremental. fwiw that `clean` was the actual cause of the rebuild-from-scratch you saw before @iabdalkader, the mount changes weren't.

Tested on OPENMV4 from a fresh clone: 2m34s clean, 7.1s incremental after touching one file, all DWARF source paths resolve to the host.

iabdalkader:
@andrewleech I've tested this again and it's working. Thanks for sending this!

PR #2892, opened 2025-10-15
ports: Add QEMU port.

iabdalkader:

PR #2896, opened 2025-10-16
imlib: Fix NULL pointer crashes in GIF/MJPEG finalisers

andrewleech:
## Issue

Hardware testing of PR #2870 revealed additional NULL pointer crashes in GIF and MJPEG modules when garbage collection runs finalisers on objects with failed `file_open()` calls.

### Stack Trace

```
#0  HardFault_Handler () at lib/cmsis/src/mimxrt/startup_MIMXRT1062.S:473
#1  <signal handler called>
#2  mp_get_stream_raise (self_in=<optimized out>, flags=flags@entry=2) at ../../py/stream.c:99
#3  0x6004d542 in file_write_helper (fp=0x20284390, buf=0x7eb7, len=1) at common/file_utils.c:100
#4  0x6004db86 in file_write (fp=fp@entry=0x20284390, data=data@entry=0x7eb7, size=size@entry=1)
#5  0x6004dbb6 in file_write_byte (fp=fp@entry=0x20284390, value=59 ';')
#6  0x60083b32 in gif_close (fp=0x20284390) at lib/imlib/gif.c:129
#7  0x600de05a in py_gif_close (self_in=<optimized out>) at openmv/modules/py_gif.c:113
```

### Root Cause

Both `gif_close()` and `mjpeg_close()` performed file I/O operations **before** calling `file_close()`:

- **gif_close()**: Wrote GIF terminator byte ';' without checking file validity
- **mjpeg_close()**: Called `mjpeg_sync()` (multiple seeks/writes) without checking

**Crash scenario:**
1. Python creates GIF/MJPEG object with finaliser (`mp_obj_malloc_with_finaliser`)
2. `file_open()` fails (invalid path, no space, etc.) → `fp->fp` remains uninitialized
3. Garbage collector runs finaliser → calls `gif_close()` or `mjpeg_close()`
4. Close function attempts I/O on invalid file pointer → **HardFault**

The VFS abstraction fixes in PR #2870 added NULL safety to `file_close()`, but GIF/MJPEG close functions do I/O **before** reaching that safety check.

## Changes

Added NULL checks before file I/O in both close functions:

### lib/imlib/gif.c
```c
void gif_close(file_t *fp) {
    // Check file validity before writing terminator byte
    // Prevents crash if finaliser runs on failed file_open()
    if (fp && fp->fp != MP_OBJ_NULL) {
        file_write_byte(fp, ';');
    }
    file_close(fp);
}
```

### lib/imlib/mjpeg.c
```c
void mjpeg_close(file_t *fp, uint32_t frames, uint32_t bytes, uint32_t us_avg) {
    // Check file validity before syncing headers
    // Prevents crash if finaliser runs on failed file_open()
    if (fp && fp->fp != MP_OBJ_NULL) {
        mjpeg_sync(fp, frames, bytes, us_avg);
    }
    file_close(fp);
}
```

**Lines changed:** 4 lines + comments across 2 files

## Testing

Created test suite to verify finaliser safety: `tests/test_finaliser_safety.py`

### Test Cases
- `test_gif_finaliser_on_failed_open` - Failed open + GC
- `test_mjpeg_finaliser_on_failed_open` - Failed open + GC
- `test_gif_finaliser_on_readonly_filesystem` - Write permission failure + GC
- `test_mjpeg_finaliser_on_readonly_filesystem` - Write permission failure + GC
- `test_multiple_failed_opens_with_gc` - 10 failures with aggressive GC
- `test_exception_in_constructor_before_file_open` - Early constructor failure + GC

### Results

**Without fix (from stack trace):** HardFault crash

**With fix (tested on OpenMV-H7):**
```
Testing GIF finaliser safety...
  Expected OSError: Could not find the path
  ✓ GIF finaliser survived failed file_open()
Testing MJPEG finaliser safety...
  Expected OSError: Could not find the path
  ✓ MJPEG finaliser survived failed file_open()
Testing multiple failures with GC...
  ✓ Multiple failures handled safely
All finaliser safety tests passed!
```

**Hardware:** OpenMV Cam H7 (STM32H743VI)  
**Test execution:** mpremote on-device  
**Result:** All tests passed, no HardFault crashes

## Verification

To reproduce the bug (before fix):
1. Checkout code before this PR
2. Build and flash firmware
3. Run: `import gif; gif.Gif("/invalid/path.gif")` (will throw OSError)
4. Run: `import gc; gc.collect()` → **HardFault**

To verify the fix:
1. Build and flash this branch
2. Run same commands → No crash, proper error handling

## Related

- PR #2870 - VFS abstraction (merged, tested, safe to merge)
- This fix addresses edge case discovered during PR #2870 hardware validation

### DIFF
--- a/lib/imlib/mjpeg.c
+++ b/lib/imlib/mjpeg.c
@@ -251,7 +251,11 @@ void mjpeg_sync(file_t *fp, uint32_t frames, uint32_t bytes, uint32_t us_avg) {
 }
 
 void mjpeg_close(file_t *fp, uint32_t frames, uint32_t bytes, uint32_t us_avg) {
-    mjpeg_sync(fp, frames, bytes, us_avg);
+    // Check file validity before syncing headers
+    // Prevents crash if finaliser runs on failed file_open()
+    if (fp && fp->fp != MP_OBJ_NULL) {
+        mjpeg_sync(fp, frames, bytes, us_avg);
+    }
     file_close(fp);
 }
 
--- a/tests/test_finaliser_safety.py
+++ b/tests/test_finaliser_safety.py
@@ -1,0 +1,104 @@
+# Tests for finaliser NULL safety in GIF and MJPEG modules
+# These tests verify that finalisers don't crash when file_open() fails
+import gc
+
+class TestFinaliserSafety:
+    def setup(self):
+        pass
+
+    def teardown(self):
+        gc.collect()
+
+    def test_gif_finaliser_on_failed_open(self):
+        """Test that GIF finaliser doesn't crash on failed file_open()"""
+        import gif
+
+        # Try to open GIF to invalid path - this will fail
+        try:
+            g = gif.Gif("/nonexistent/invalid/path.gif")
+        except OSError:
+            pass  # Expected to fail
+
+        # Force garbage collection to run finaliser
+        gc.collect()
+
+        # If we get here without HardFault, the fix works
+        assert True, "GIF finaliser survived failed file_open()"
+
+    def test_mjpeg_finaliser_on_failed_open(self):
+        """Test that MJPEG finaliser doesn't crash on failed file_open()"""
+        import mjpeg
+
+        # Try to open MJPEG to invalid path - this will fail
+        try:
+            m = mjpeg.Mjpeg("/nonexistent/invalid/path.mjpeg")
+        except OSError:
+            pass  # Expected to fail
+
+        # Force garbage collection to run finaliser
+        gc.collect()
+
+        # If we get here without HardFault, the fix works
+        assert True, "MJPEG finaliser survived failed file_open()"
+
+    def test_gif_finaliser_on_readonly_filesystem(self):
+        """Test GIF finaliser when trying to write to read-only location"""
+        import gif
+
+        # Try to create GIF in location without write permission
+        try:
+            g = gif.Gif("/flash/readonly_test.gif")
+        except (OSError, PermissionError):
+            pass  # May fail with different errors depending on filesystem
+
+        gc.collect()
+        assert True, "GIF finaliser handled readonly filesystem"
+
+    def test_mjpeg_finaliser_on_readonly_filesystem(self):
+        """Test MJPEG finaliser when trying to write to read-only location"""
+        import mjpeg
+
+        # Try to create MJPEG in location without write permission
+        try:
+            m = mjpeg.Mjpeg("/flash/readonly_test.mjpeg")
+        except (OSError, PermissionError):
+            pass  # May fail with different errors depending on filesystem
+
+        gc.collect()
+        assert True, "MJPEG finaliser handled readonly filesystem"
+
+    def test_multiple_failed_opens_with_gc(self):
+        """Test that multiple failed opens with GC cycles don't leak or crash"""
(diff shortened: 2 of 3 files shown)


iabdalkader:
I think I fixed those issues in followup PRs in https://github.com/openmv/openmv/pull/2890 and https://github.com/openmv/openmv/pull/2893. Passing NULL fp or NULL fp->fp should now be as safe as the old invalid FD.
You can add more unit tests if you like to https://github.com/openmv/openmv/pull/2892 after it's merged.

andrewleech:
## Rebased onto latest master (46961ffe)

The branch has been rebased onto upstream master which includes recent file_utils.c NULL safety improvements (commits ccf6f14e and 34687cda).

### Why This Fix Is Still Potentially useful

Upstream added NULL checks to low-level file functions like `file_write()`, `file_seek()`, etc. These functions now silently return when `fp` or `fp->fp` is NULL.

However, our explicit NULL checks in `gif_close()` and `mjpeg_close()` remain valuable:

1. **Explicit intent** - Clear at call site that validity checking matters
2. **Avoids unnecessary calls** - Don't enter `mjpeg_sync()` for 10+ function calls that would silently fail
3. **Defense-in-depth** - Doesn't rely on lower-level NULL safety
4. **Maintainability** - Code still works if someone removes low-level checks

This is still a double-up of error checking, so quite happy to close this out if you don't think it's adding much value!

lib/imlib/mjpeg.c
```diff
@@ -251,7 +251,11 @@ void mjpeg_sync(file_t *fp, uint32_t frames, uint32_t bytes, uint32_t us_avg) {
 }
 
 void mjpeg_close(file_t *fp, uint32_t frames, uint32_t bytes, uint32_t us_avg) {
-    mjpeg_sync(fp, frames, bytes, us_avg);
+    // Check file validity before syncing headers
+    // Prevents crash if finaliser runs on failed file_open()
+    if (fp && fp->fp != MP_OBJ_NULL) {
+        mjpeg_sync(fp, frames, bytes, us_avg);
+    }
```
iabdalkader:
```suggestion
    // Check file validity before syncing headers
    // Prevents crash if finaliser runs on failed file_open()
    if (fp && fp->fp != MP_OBJ_NULL) {
        mjpeg_sync(fp, frames, bytes, us_avg);
        file_close(fp);
    }
```

tests/test_finaliser_safety.py
```diff
@@ -0,0 +1,104 @@
+# Tests for finaliser NULL safety in GIF and MJPEG modules
```
iabdalkader: If you want to add tests, please add them to `scripts/unittests/tests`. See the other unit tests for the style (basically a single function that returns True/False, also don't catch any exceptions the calling code already catches exceptions).

iabdalkader:
> This is still a double-up of error checking, so quite happy to close this out if you don't think it's adding much value!

@andrewleech `file_t` should be an opaque pointer/type (no other code outside `file_utils.c` should access its internals). This will help if we ever want to change the internals or make `file_t` a file descriptors. This change will fail to build as it checks `fp->fp`. Plus, it's currently safe to call any function with invalid/NULL FP, and this will always be the case, so I don't think this is needed.

PR #2904, opened 2025-10-29
ports/stm32: Enable N6 ethernet.

kwagyeman:
Necessary for https://github.com/micropython/micropython/pull/18161 to work on the OpenMV MicroPython.

`#define MICROPY_HW_ETH_DMA_ATTRIBUTE        __attribute__((aligned(16384), section(".dma_buffer")));`

Causes a memory fault error because MicroPython allocates the ethernet dma buffer via an MPU region in the ETH driver. For ARMV7 on the H7 you get two regions and the lower-priority one wins. On V8 you get a hardfault.

The `#define MICROPY_HW_ETH_DMA_ATTRIBUTE` by not be needed as MicroPython defines that value by default to be `#define MICROPY_HW_ETH_DMA_ATTRIBUTE        __attribute__((aligned(16384)))`. This solves the memory alignment and then an MPU region is allocated for the buffer. Need to check if this can be removed on the H7 Plus... I remember it being needed at the time, though. It's not needed for the n6.


kwagyeman:
Okay, I see why `#define MICROPY_HW_ETH_DMA_ATTRIBUTE        __attribute__((aligned(16384), section(".dma_buffer")));` was needed, it moves the DMA buffer for the H7 ETH to the DMA memory of 32 KB which is already allocated. Without the `.dma_buffer` part, the allocation moves to the DTCM and then we get an overflow... also, the ethernet controller wouldn't be able to access to the DTCM. So, this is absolutely needed for the H7.

Leaving this PR as a draft while issues are resolved upstream.

kwagyeman:
@iabdalkader - This PR is ready. Now just needs Damien to finish ethernet.

iabdalkader:
This needs to be rebased

kwagyeman:
Done and retested on the N6. Ethernet is working fine.

boards/OPENMV_N6/omv_boardconfig.h
```diff
@@ -112,6 +112,9 @@
 #define OMV_RCC_IC12_CLKDIV                 (8)
 
 // Used by MicroPython for slow peripherals.
+#define OMV_RCC_IC12_SOURCE                 (RCC_ICCLKSOURCE_PLL1)
+#define OMV_RCC_IC12_CLKDIV                 (8)
+
```
iabdalkader:
```suggestion
```

Already defined.

ports/stm32/omv_mpconfigport.h
```diff
@@ -50,7 +50,9 @@
 #define MICROPY_BANNER_NAME_AND_VERSION "OpenMV " OPENMV_GIT_TAG "; MicroPython " MICROPY_GIT_TAG
 #define MICROPY_BOARD_FATAL_ERROR           __fatal_error
 #define MICROPY_HW_DMA_ENABLE_AUTO_TURN_OFF (0)
+#if defined(ARDUINO_PORTENTA_H7)
```
iabdalkader, line 53: Should this be something more generic? Like test for F7/H7 series?

boards/OPENMV_N6/omv_boardconfig.h
kwagyeman: Removed.

ports/stm32/omv_mpconfigport.h
kwagyeman: Checking...

kwagyeman: No, we only have to do this because the Portenta's RAM setup places the .data region, where the Ethernet buffer normally ends up, in the DTCM, which is not DMA accessible. So, we have to force it into the DMA buffer region.

On the normal H7/P boards, .data is in SRAM1, which is DMA accessible, so it would not be needed.

So, this is required explicitly just for the Portenta.

iabdalkader (approved):

PR #2909, opened 2025-11-02
lib/stai: Fix STAI NPU memory allocation.

kwagyeman:
The ST Neural ART code generator marks the HyperRAM mpool as cacheable (CACHEABLE_ON), which causes the NPU stream engines to issue cacheable accesses. However, the N6 memory subsystem does not support cacheable accesses to internal memories, only external ones. Because of this, relocating the HyperRAM region to internal memory results in invalid access behavior: only base addresses are updated during relocation, not the cacheable attributes.

In short, buffers generated for HyperRAM must remain in external memory. Marking HyperRAM as non-cacheable is a possible workaround, but may reduce performance if buffers are placed in AXI RAM instead.

Forcing the ext_ram_sz allocation to 1MB ensures that the ext_ram_sz is allocated in SDRAM.

---

I tested that the XIP RAM is okay to be in the internal SRAM. Only the EXT RAM allocation must be in SDRAM.

### DIFF
--- a/lib/stai/stai_backend.c
+++ b/lib/stai/stai_backend.c
@@ -49,6 +49,10 @@
 #include "ll_aton_caches_interface.h"
 #include "ll_aton_reloc_network.h"
 
+// Due to limitations with the STM32N6 NPU design, the ext_ram_sz memory pool
+// must be in external SDRAM and not in internal SRAM. Forcing a minimum of a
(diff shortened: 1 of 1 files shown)


iabdalkader:
LGTM, thanks for the detailed commit message.

PR #2910, opened 2025-11-02
ports/alif: Fully implement Alif GPU driver.

kwagyeman:
Implements all modes the D/AVE GPU supports that we can use. RGB565 destination drawing fully supports all possible modes (except transforms). For grayscale destinations, only drawing grayscale images without alpha/colorPal/alphaPal is supported.

I tried to get a grayscale to grayscale drawing with alpha blending operational. It kinda worked, but the hardware seems to want to blend the two images even at 255 alpha, where one should be picked over the other.

### DIFF
--- a/ports/alif/omv_gpu.c
+++ b/ports/alif/omv_gpu.c
@@ -135,6 +135,22 @@ int omv_gpu_draw_image(image_t *src_img,
                        const uint8_t *alpha_palette,
                        image_hint_t hint,
                        float *transform) {
+    image_t src_copy;
+    image_t dst_copy;
+
+    // BAYER is 8-bpp; the GPU can only do a plain 1:1 crop/copy of it. Any
+    // scale/filter/mirror/flip/blend hint corrupts the raw mosaic, so alias
+    // to GRAYSCALE only for a hintless 1:1 blit; everything else falls to SW.
+    if (src_img->is_bayer && dst_img->is_bayer && hint == 0 &&
+        src_rect->w == dst_rect->w && src_rect->h == dst_rect->h) {
+        src_copy = *src_img;
+        src_copy.pixfmt = PIXFORMAT_GRAYSCALE;
+        src_img = &src_copy;
+        dst_copy = *dst_img;
+        dst_copy.pixfmt = PIXFORMAT_GRAYSCALE;
+        dst_img = &dst_copy;
+    }
+
     // GPU2D can only draw on RGB565/GRAYSCALE buffers.
     if (dst_img->pixfmt != PIXFORMAT_RGB565 && dst_img->pixfmt != PIXFORMAT_GRAYSCALE) {
         return -1;
--- a/ports/alif/omv_gpu.c
+++ b/ports/alif/omv_gpu.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023-2024 OpenMV, LLC.
+ * Copyright (C) 2023-2026 OpenMV, LLC.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -42,6 +42,7 @@
 #include "omv_gpu.h"
 
 static d2_device *dev;
+static d2_color CLUT_BUFFER[256];
 
 extern char _gpu_memory_start;
 extern char _gpu_memory_end;
@@ -89,16 +90,6 @@ int omv_gpu_init() {
         __fatal_error("Failed to init GPU hardware");
     }
 
-#if 0
-    d2_setblendmode(dev, d2_bm_alpha, d2_bm_one_minus_alpha);
-    d2_setalphamode(dev, d2_am_constant); // default in d2_newcontext
-    d2_setlinecap(dev, d2_lc_butt);
-    d2_setantialiasing(dev, 1);
-    d2_setblur(dev, 0);
-    d2_settextureoperation(dev, d2_to_one, d2_to_copy, d2_to_copy, d2_to_copy);
-    d2_setfillmode(dev, d2_fm_texture);
-    d2_settexturemode(dev, d2_tm_filter);
-#endif
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


kwagyeman:
Was going to ask for this in the release. But, then discovered a bug on more testing. So, it can wait.

kwagyeman:
Working on this now...

iabdalkader:

https://forums.openmv.io/t/openmv-ae3-grayscale-issue-with-pag7936-in-hd/11599/2

skemp117-persimia:
Following

kwagyeman:
This should be ready now. Tested cropping with all CSI formats on the AE3, image conversions, and all the various drawing tests that exercise different palettes and scaling.

skemp117-persimia:
The AE3 functions in my use case (BW, crop) are now working with this FW. Great stuff! Thank you

kwagyeman:
Want to check a few more things...

kwagyeman:
This PR is ready now.

ports/alif/omv_gpu.c
```diff
@@ -143,23 +179,48 @@ int omv_gpu_draw_image(image_t *src_img,
...
+    if (hint & IMAGE_HINT_VFLIP) {
+        blit_flags |= d2_bf_mirrorv;
+    }
+
+    if (src_img->pixfmt == PIXFORMAT_GRAYSCALE) {
         blit_flags |= d2_bf_usealpha;
+
+        if (dst_img->pixfmt == PIXFORMAT_RGB565) {
+            omv_gpu_load_clut(color_palette, alpha_palette);
+            err = d2_settexclut(dev, CLUT_BUFFER);
```
iabdalkader:
```suggestion
            err = d2_settexclut(dev, LocalToGlobal(CLUT_BUFFER));
```

ports/alif/omv_gpu.c
```diff
@@ -109,17 +100,30 @@ void omv_gpu_deinit() {
...
-            return d2_mode_alpha1;
-        case PIXFORMAT_BAYER_ANY:
         case PIXFORMAT_GRAYSCALE:
-            return d2_mode_alpha8;
+            return clut ? d2_mode_i8 | d2_mode_clut : d2_mode_alpha8;
         case PIXFORMAT_RGB565:
             return d2_mode_rgb565;
     }
-    mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Pixel format is not supported"));
+    __builtin_unreachable();
```
iabdalkader, line 126: Should default: raise here, what if we add new pix formats?

ports/alif/omv_csi.c
```diff
@@ -369,6 +369,12 @@ int alif_csi_snapshot(omv_csi_t *csi, image_t *dst_image, uint32_t flags) {
         src_cimage.w = csi->resolution[csi->framesize][0];
         src_cimage.h = csi->resolution[csi->framesize][1];
 
+        // GPU only supports GRAYSCALE/RGB565 buffers. BAYER==GRAYSCALE for GPU.
+        if (src_cimage.pixfmt == PIXFORMAT_BAYER) {
+            src_cimage.pixfmt = PIXFORMAT_GRAYSCALE;
+            dst_cimage.pixfmt = PIXFORMAT_GRAYSCALE;
+        }
```
iabdalkader: This should be handled in the GPU driver.

iabdalkader (changes requested):

ports/alif/omv_gpu.c
kwagyeman: The if statements in omv_gpu_draw_image() properly prevent any other pixformats from reaching that. So, it's dead code.

If the GPU can't handle something, it should return -1 for software to take over. Not, throw an error.

ports/alif/omv_gpu.c
kwagyeman: Doing that causes an explicit system crash and isn't required. However, it should be if it were being DMAed.

On testing, CLUT_BUFFER just needs to be persistent, but doesn't have to be in DMA memory. It's managed by d2_settexclut via CPU buffers. Moved to just being static.

ports/alif/omv_csi.c
kwagyeman: Done, the driver can now crop any Bayer image.

ports/alif/omv_gpu.c
iabdalkader: No, it's not just DMA, it's to get a global TCM address for peripherals too, should be a no-op otherwise.

ports/alif/omv_gpu.c
```diff
@@ -143,23 +179,48 @@ int omv_gpu_draw_image(image_t *src_img,
...
+        if (dst_img->pixfmt == PIXFORMAT_RGB565) {
+            omv_gpu_load_clut(color_palette, alpha_palette);
+            err = d2_settexclut(dev, CLUT_BUFFER);
+            OMV_GPU_CHECK_ERROR(err);
+        }
     }
 
-    // Flush source image
+    // Flush destination/source image
+    SCB_CleanInvalidateDCache_by_Addr(dst_img->data, image_size(dst_img));
```
iabdalkader, line 239: Why the cleanInvalidate not just `SCB_InvalidateDCache_by_Addr`?

ports/alif/omv_gpu.c
kwagyeman: Added it in; it doesn't cause a crash anymore now that the CLUT is just a normal static buffer.

ports/alif/omv_gpu.c
kwagyeman: This is the same pattern in the other GPU drivers. You have to clean it because the dst buffer could have been modified by the CPU. So, those cache lines need to be written back.

iabdalkader: But, invalidate would drop them. We don't really need them written back, since we're about to overwrite the buffer anyway.

kwagyeman: No, we support image blending. Also, a draw operation hits multiple cache lines, like drawing a line. That has to be written back.

If you didn't clean the cache whatever changes you did to the image like filtering or etc would all be lost.

kwagyeman: Here's an example:

```
img.mean(1)
img.draw_image(other_img, 10, 20, x_scale=0.5, y_scale=0.5)
```

The dest image needs to be cleaned first because the CPU modified it. GPU drawing doesn't mean we are overwriting all pixels in the image.

iabdalkader: Good point.

iabdalkader (approved):

PR #2911, opened 2025-11-03
ports/alif: Improve PAG7936 Image Quality.

kwagyeman:
Fixes issues with the speed of the AWB algorithm and enables controls to enable/disable it.

### DIFF
--- a/drivers/sensors/pag7936.c
+++ b/drivers/sensors/pag7936.c
@@ -652,6 +652,21 @@ static int get_exposure_us(omv_csi_t *csi, int *exposure_us) {
     return ret;
 }
 
+static int set_auto_whitebal(omv_csi_t *csi, int enable, float r_gain_db, float g_gain_db, float b_gain_db) {
+    csi->stats_enabled = enable;
+    return 0;
+}
+
+static int get_rgb_gain_db(omv_csi_t *csi, float *r_gain_db, float *g_gain_db, float *b_gain_db) {
+    uint32_t r, g, b;
+    omv_csi_get_stats(csi, &r, &g, &b);
+
+    *r_gain_db = 20.0f * log10f(IM_DIV((float) g, r));
+    *g_gain_db = 0.0f;
+    *b_gain_db = 20.0f * log10f(IM_DIV((float) g, b));
+    return 0;
+}
+
 static int set_hmirror(omv_csi_t *csi, int enable) {
     uint8_t reg;
     int ret = omv_i2c_readb2(csi->i2c, csi->slv_addr, TG_FLIP, &reg);
@@ -723,6 +738,8 @@ int pag7936_init(omv_csi_t *csi) {
     csi->get_gain_db = get_gain_db;
     csi->set_auto_exposure = set_auto_exposure;
     csi->get_exposure_us = get_exposure_us;
+    csi->set_auto_whitebal = set_auto_whitebal;
+    csi->get_rgb_gain_db = get_rgb_gain_db;
     csi->set_hmirror = set_hmirror;
     csi->set_vflip = set_vflip;
 
--- a/drivers/sensors/ps5520.c
+++ b/drivers/sensors/ps5520.c
@@ -1442,6 +1442,21 @@ static int get_exposure_us(omv_csi_t *csi, int *exposure_us) {
     return ret;
 }
 
+static int set_auto_whitebal(omv_csi_t *csi, int enable, float r_gain_db, float g_gain_db, float b_gain_db) {
+    csi->stats_enabled = enable;
+    return 0;
+}
+
+static int get_rgb_gain_db(omv_csi_t *csi, float *r_gain_db, float *g_gain_db, float *b_gain_db) {
+    uint32_t r, g, b;
+    omv_csi_get_stats(csi, &r, &g, &b);
+
+    *r_gain_db = 20.0f * log10f(IM_DIV((float) g, r));
+    *g_gain_db = 0.0f;
+    *b_gain_db = 20.0f * log10f(IM_DIV((float) g, b));
+    return 0;
+}
+
 static int set_hmirror(omv_csi_t *csi, int enable) {
     uint8_t reg1, reg2;
     uint16_t hsize;
@@ -1590,6 +1605,8 @@ int ps5520_init(omv_csi_t *csi) {
     csi->get_gain_db = get_gain_db;
     csi->set_auto_exposure = set_auto_exposure;
     csi->get_exposure_us = get_exposure_us;
+    csi->set_auto_whitebal = set_auto_whitebal;
+    csi->get_rgb_gain_db = get_rgb_gain_db;
     csi->set_hmirror = set_hmirror;
     csi->set_vflip = set_vflip;
 


common/average.h
```diff
@@ -0,0 +1,57 @@
...
+#ifndef __AVERAGE_H__
+#define __AVERAGE_H__
+
+#include <stddef.h>
+#include <stdint.h>
+
+typedef struct rgb_moving_average_val {
+    uint32_t r_value;
+    uint32_t g_value;
+    uint32_t b_value;
```
iabdalkader: Why `uint32_t`?

common/average.h
```diff
@@ -0,0 +1,57 @@
...
+    uint32_t timestamp_ms;
+} rgb_moving_average_val_t;
+
+typedef struct rgb_moving_average {
+    size_t size;
+    size_t tail;
+    size_t head;
+    size_t count;
+    uint32_t max_ms;
+    rgb_moving_average_val_t *buffer;
```
iabdalkader: Why a  pointer?
```suggestion
    rgb_buf[OMV_CSI_RGB_AVG_SIZE];
```

common/average.c
```diff
@@ -0,0 +1,81 @@
+/*
```
iabdalkader: I don't like this API, it should be `omv_csi_x`. For example, `omv_csi_stats_t`, and functions moved to `common/omv_csi.c`: `omv_csi_stats_update` etc...

drivers/sensors/pag7936.c
```diff
@@ -652,6 +652,23 @@ static int get_exposure_us(omv_csi_t *csi, int *exposure_us) {
...
 
+static int set_auto_whitebal(omv_csi_t *csi, int enable, float r_gain_db, float g_gain_db, float b_gain_db) {
+    csi->awb_enabled = enable;
+    return 0;
+}
+
+static int get_rgb_gain_db(omv_csi_t *csi, float *r_gain_db, float *g_gain_db, float *b_gain_db) {
+    uint32_t r_avg, g_avg, b_avg;
+    rgb_moving_average_get(&csi->rgb_avg, &r_avg, &g_avg, &b_avg);
+    float g_avg_f = (float) g_avg;
```
iabdalkader:
```suggestion
```

lib/imlib/bayer.c
```diff
@@ -2291,13 +2298,19 @@ static void vdebayer_rggb_to_rgb565_awb_quarter(image_t *src, image_t *dst, uint
...
+                             float brightness, float contrast, float gamma) {
     uint32_t red_gain = IM_DIV(g_out * 32, r_out);
     red_gain = IM_MIN(red_gain, 128U);
 
     uint32_t blue_gain = IM_DIV(g_out * 32, b_out);
     blue_gain = IM_MIN(blue_gain, 128U);
 
+    for (int i = 0; i < 256; i++) {
+        int v = fast_roundf(((fast_powf(i * (1.0f / 255.0f), gamma) * contrast) + brightness) * 255.0f);
+        gamma_lut[i] = IM_CLAMP(v, 0, 255);
```
iabdalkader: Should this be recalculated on every call or when those values change?

ports/alif/omv_portconfig.h
```diff
@@ -137,9 +138,14 @@ struct {                    \
...
-    struct {                    \
-        CPI_Type *base;         \
+#define OMV_CSI_RGB_AVG_SIZE    (60)
+
+#define OMV_CSI_PORT_BITS                                           \
+    struct {                                                        \
+        CPI_Type *base;                                             \
+        bool awb_enabled;                                           \
+        rgb_moving_average_t rgb_avg;                               \
+        rgb_moving_average_val_t rgb_avg_buf[OMV_CSI_RGB_AVG_SIZE]; \
```
iabdalkader: These will be part of the standard omv_csi_t.

common/omv_csi.c
```diff
@@ -322,6 +324,12 @@ __weak int omv_csi_reset(omv_csi_t *csi, bool hard) {
         NULL, NULL
     };
 
+    #if defined(OMV_CSI_RGB_AVG_SIZE)
```
iabdalkader:
```suggestion
    #if defined(OMV_CSI_STATS_ENABLE)
```

iabdalkader (changes requested):
Mostly okay. Main change: Rework the API..

common/average.h
iabdalkader: That was just an example, please find a better name, maybe `samples`.

common/average.h
kwagyeman: The PAG7936 registers are 32-bit accumulators. I could reduce these down like the N6 does beforehand to save RAM. Do it?

common/average.h
kwagyeman: Sure, it was done to decouple the object from the data buffer which needs to be a different size for the AE3 versus N6 as the N6 has a faster max fps on the PAG7936 so I need to make the buffer larger.

common/average.h
iabdalkader: Right, these are averages, never mind.

lib/imlib/bayer.c
kwagyeman: I was going to expose the brightness/contrast controls to the sensor API later on. It only needs to be done when the value changes, however, the lut part is an implementation detail. So, it shouldn't leave bayer.c. I can move the table update to another function call that's only done on value changes.

ports/alif/omv_portconfig.h
kwagyeman: It's only needed for cameras that do the AWB themselves. Recommendation for ifdefing it in omv_csi_t?

lib/imlib/bayer.c
```diff
@@ -559,6 +559,10 @@ static inline v128_t vdebayer_to_cr(vrgb_pixels_t pixels) {
     return pixels.r;
 }
 
+// FIXME: This needs to be placed in the DTCM. .bss is in the DTCM on Alif. But, this is not portable.
+// Move this to a proper DTCM section for other platforms.
+static uint8_t gamma_lut[256];
```
kwagyeman: @iabdalkader - Ideas on how to solve this? It needs to be in the DTCM because that has the quad read port on the AE3.

common/average.h
iabdalkader: It can still be a different size, just not a pointer.

ports/alif/omv_portconfig.h
iabdalkader: `OMV_CSI_STATS_ENABLE` ?

common/average.h
kwagyeman: Done

common/average.c
kwagyeman: Done, renamed it to awb as it's controlling that.

drivers/sensors/pag7936.c
kwagyeman: Done

ports/alif/omv_portconfig.h
kwagyeman: Refactored

common/omv_csi.c
kwagyeman: Done

lib/imlib/bayer.c
kwagyeman: Done

boards/OPENMV_AE3/omv_boardconfig.h
```diff
@@ -273,6 +273,7 @@ extern unsigned char OMV_BOARD_UID_ADDR[12];    // Unique address.
 // Camera interface
 #define OMV_CSI_BASE                    ((CPI_Type *) CPI_BASE)
 #define OMV_CSI_CLK_FREQUENCY           (24000000)
+#define OMV_CSI_AWB_ENABLE
```
iabdalkader:
```suggestion
#define OMV_CSI_STATS_ENABLE   (1)
```
There are two different things here: the CSI collects stats, and the image library does something with the stats (AWB). This enables collecting stats, what you do with those stats is a different story. `omv_csi_awb_get` doesn't make sense.

common/omv_csi.c
```diff
@@ -1635,6 +1645,45 @@ __weak int omv_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
     return ret;
 }
 
+#if defined(OMV_CSI_AWB_ENABLE)
```
iabdalkader:
```suggestion
#if OMV_CSI_STATS_ENABLE
```

common/omv_csi.h
```diff
@@ -313,6 +313,17 @@ typedef struct _omv_clk {
...
+#if defined(OMV_CSI_AWB_ENABLE)
+typedef struct omv_csi_awb {
+    bool enable;
+    bool initialized;
+    uint32_t last_ms;
+    float r_ema;
+    float g_ema;
+    float b_ema;
+} omv_csi_awb_t;
+#endif // defined(OMV_CSI_AWB_ENABLE)
```
iabdalkader:
```suggestion
#if OMV_CSI_STATS_ENABLE
typedef struct omv_csi_stats {
    bool enable;
    bool initialized;
    uint32_t last_ms;
    float r_avg;
    float g_avg;
    float b_avg;
} omv_csi_stats_t;
#endif // OMV_CSI_STATS_ENABLE
```

common/omv_csi.h
```diff
@@ -603,6 +618,14 @@ int omv_csi_copy_line(omv_csi_t *csi, void *dma, uint8_t *src, uint8_t *dst);
 // Default snapshot function.
 int omv_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags);
 
+#if defined(OMV_CSI_AWB_ENABLE)
+// Update RGB moving average struct.
+void omv_csi_awb_update(omv_csi_t *csi, uint32_t *r, uint32_t *g, uint32_t *b, uint32_t ms);
+
+// Get RGB moving average values.
+void omv_csi_awb_get(omv_csi_t *csi, uint32_t *r, uint32_t *g, uint32_t *b);
+#endif // defined(OMV_CSI_AWB_ENABLE)
```
iabdalkader:
```suggestion
#if OMV_CSI_STATS_ENABLE
// Update RGB stats.
void omv_csi_stats_update(omv_csi_t *csi, uint32_t *r, uint32_t *g, uint32_t *b, uint32_t ms);

// Get RGB stats.
void omv_csi_get_stats(omv_csi_t *csi, uint32_t *r, uint32_t *g, uint32_t *b);
#endif // OMV_CSI_STATS_EMA_ALPHA
```

common/omv_csi.c
```diff
@@ -1635,6 +1645,45 @@ __weak int omv_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
...
+        awb->last_ms = ms;
+        awb->r_ema += alpha * (fr - awb->r_ema);
+        awb->g_ema += alpha * (fg - awb->g_ema);
+        awb->b_ema += alpha * (fb - awb->b_ema);
+    }
+
+    // Return the current average.
+    *r = fast_roundf(awb->r_ema);
+    *g = fast_roundf(awb->g_ema);
+    *b = fast_roundf(awb->b_ema);
```
iabdalkader:
```suggestion
```

Don't think this should return the values, just call get.

lib/imlib/bayer.c
iabdalkader: If you want a DTCM section, just add one (e.g, `OMV_FAST_MEMORY`), what's the problem?

iabdalkader: Here's my suggestion: Move this to `imlib.c`, initialize in init0, while we're at it, please rename `imlib_init/deinit_all` to just `imlib_init/deinit`.

**Please send this in a new PR first.**

I'd also consider mapping the input from an easier to work with range, like: `0.0f..1.0f` or `0..255`, to make it easier to pass down from Python. This can also be done in the Python module.

We can later implement set_brightness/contrast/gamma, have them call this function to update the table.

```suggestion
#if ENABLE_GAMMA_LUT
// extern in header
uint8_t gamma_lut[256];
#endif

void imlib_init0() {
#if ENABLE_GAMMA_LUT
   imlib_update_gamma_table(-0.2f, 1.0f, 2.2f);
#endif
}

#if ENABLE_GAMMA_LUT
// brightness: -1.0 to 1.0, contrast: 0.0 to 2.0, gamma: 0.0 to inf.
void imlib_update_gamma_table(float brightness, float contrast, float gamma) {
    for (int i = 0; i < 256; i++) {
        int v = fast_roundf(((fast_powf(i * (1.0f / 255.0f), (1.0f / gamma)) * contrast) + brightness) * 255.0f);
        gamma_lut[i] = IM_CLAMP(v, 0, 255);
    }
}
#endif
```

common/omv_csi.c
iabdalkader: The function below, already does that, just call it.
```suggestion
```

lib/imlib/bayer.c
kwagyeman: Moved to: https://github.com/openmv/openmv/pull/2914

> I'd also consider mapping the input from an easier to work with range, like: 0.0f..1.0f or 0..255, to make it easier to pass down from Python. This can also be done in the Python module.

I tried to match the way arguments are passed to the function [gamma()](https://docs.openmv.io/library/omv.image.html#image.Image.gamma).

> We can later implement set_brightness/contrast/gamma, have them call this function to update the table.

Yes, but, note, the original code with it just recalculates every time was the easiest for handle future updates. This table will need to be moved to the CSI struct after brightness/contrast controls are exposed, as it technically needs to be reset on sensor reset and varies per sensor, since sensors can be controlled independently.

Anyway, the current method is the best for it being static at the moment, but, expect it to be refactored again shortly.

kwagyeman: Moved to the other PR with fast memory.

common/omv_csi.c
kwagyeman: Done

common/omv_csi.h

common/omv_csi.h

boards/OPENMV_AE3/omv_boardconfig.h
kwagyeman: Done

lib/imlib/bayer.c
iabdalkader: > I tried to match the way arguments are passed to the function [gamma()](https://docs.openmv.io/library/omv.image.html#image.Image.gamma).

That function applies gamma correction to an image, right? I think we actually need to match these functions instead https://docs.openmv.io/library/omv.sensor.html#sensor.set_contrast

But the -3 to 3 range is also a really horrible artifact from the past. Perhaps we should refactor those to accept 0..100, or floats.

common/omv_csi.h
```diff
@@ -313,6 +313,17 @@ typedef struct _omv_clk {
...
 } omv_clk_t;
 
+#if defined(OMV_CSI_STATS_ENABLE)
+typedef struct omv_csi_stats {
+    bool enable;
+    bool initialized;
+    uint32_t last_ms;
+    float r_avg;
+    float g_avg;
+    float b_avg;
```
iabdalkader:
```suggestion
    float r_avg;
    float g_avg;
    float b_avg;
    uint32_t last_ms;
```

Can we use r_avg==0 to detect if it's initialized or not? Or bad idea?
Also please add bool stats_enabled to csi_t directly.

lib/imlib/bayer.c
kwagyeman: I like the current float format for [gamma()](https://docs.openmv.io/library/omv.image.html#image.Image.gamma) as it's normalized and can easily map to anything.

> -3 to 3 range is also a really horrible artifact from the past

Agreeded :)

common/omv_csi.h
kwagyeman: > Can we use r_avg==0 to detect if it's initialized or not? Or bad idea?

No, we don't want to do that as once I add black line calibration (which subtracts the value of pixels when seeing pure black, like when the lens cap is on), then it's quite possible to have the image go to all 0 in a dark scene.

> Also please add bool stats_enabled to csi_t directly.

Done

kwagyeman:
Rebased and retested.

common/omv_csi.c
```diff
@@ -1642,6 +1652,38 @@ __weak int omv_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
...
+        uint32_t dt_ms = ms - stats->last_ms;
+        // Continuous-time EMA gain for elapsed dt: alpha = 1 - exp(-dt/τ)
+        float alpha = IM_CLAMP(1.0f - expf(-((float) dt_ms) / OMV_CSI_STATS_TAU_MS), 0.0f, 1.0f);
+        stats->last_ms = ms;
+        stats->r_avg += alpha * (*r - stats->r_avg);
+        stats->g_avg += alpha * (*g - stats->g_avg);
+        stats->b_avg += alpha * (*b - stats->b_avg);
+    }
+
+    omv_csi_get_stats(csi, r, g, b);
```
iabdalkader:
```suggestion
```
Should just call `stats_get` after the update.

common/omv_csi.c
```diff
@@ -322,6 +326,12 @@ __weak int omv_csi_reset(omv_csi_t *csi, bool hard) {
         NULL, NULL
     };
 
+    #if defined(OMV_CSI_STATS_ENABLE)
+    // Setup AWB moving average.
```
iabdalkader:
```suggestion
    // Enable collecting RGB stats.
```

common/omv_csi.c
```diff
@@ -1642,6 +1652,38 @@ __weak int omv_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
     return ret;
 }
 
+#if defined(OMV_CSI_STATS_ENABLE)
```
iabdalkader: Can you move this higher up? maybe after/before omv_csi_get_rgb_gain_db or whitebal.

common/omv_csi.h
```diff
@@ -604,6 +619,14 @@ int omv_csi_copy_line(omv_csi_t *csi, void *dma, uint8_t *src, uint8_t *dst);
 // Default snapshot function.
 int omv_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags);
 
+#if defined(OMV_CSI_STATS_ENABLE)
```
iabdalkader: Should be after `omv_csi_get_rgb_gain_db`.

ports/stm32/stm_isp.c
```diff
@@ -33,24 +33,31 @@
 
 #include <string.h>
 #include "imlib.h"
+#include "py/mphal.h"
 #include "stm_isp.h"
 #include "omv_boardconfig.h"
 
 #ifdef DCMIPP
-float stm_isp_update_awb(DCMIPP_HandleTypeDef *dcmipp, uint32_t pipe, uint32_t n_pixels) {
+float stm_isp_update_awb(omv_csi_t *csi, uint32_t pipe, uint32_t n_pixels) {
```
iabdalkader, line 41: I think we should be consistent here: pass `csi` to `stm_isp_config_pipeline` as well. Maybe call it `stm_isp_init`?

common/omv_csi.c
kwagyeman: Done

common/omv_csi.c
kwagyeman: Done

common/omv_csi.c
kwagyeman: Done

common/omv_csi.h

ports/stm32/stm_isp.c
kwagyeman: Done, and split the commit.

kwagyeman:
Retested changes with AE3 and N6 too! :)

PR #2912, opened 2025-11-03
boards: Increase RAW preview window size.

kwagyeman:
Makes raw preview look good for AI models.

Before:
before

After:
after


kwagyeman:
This doesn't matter after the new protocol is merged. The setting is controlled by the IDE.

PR #2914, opened 2025-11-10
ports: Enable gamma correction.

kwagyeman:
Adds gamma correction to the image for the AE3 and N6, improving the image brightness.

Macbeth Chart color matrix correction needs to be done next to push the image color saturation.

PAG7936-AE3-Before:
image

PAG7936-AE3-Afterwards:
image

PAG7936-N6-Before:
image

PAG7936-N6-Afterwards:
image

PS5520-N6-Before:
image

PS5520-N6-Afterwards:
image

The lighter images represent gamma correction being applied to move the image from the raw RGB color space to the sRGB color space suitable for display. For machine vision processing purposes, the lighter images are more useful, generally for algorithms.

### DIFF
--- a/ports/stm32/stm_isp.c
+++ b/ports/stm32/stm_isp.c
@@ -159,6 +159,14 @@ int stm_isp_config_pipeline(DCMIPP_HandleTypeDef *dcmipp, uint32_t pipe,
         }
     }
 
+    if (HAL_DCMIPP_PIPE_SetISPBadPixelRemovalConfig(dcmipp, pipe, DCMIPP_BAD_PXL_REM_SRENGTH_4) != HAL_OK) {
+        return -1;
+    }
+
+    if (HAL_DCMIPP_PIPE_EnableISPBadPixelRemoval(dcmipp, pipe) != HAL_OK) {
+        return -1;
+    }
+
     return 0;
 }
 


lib/imlib/imlib.h
```diff
@@ -1192,8 +1195,9 @@ typedef struct imlib_draw_row_data {
 typedef void (*imlib_draw_row_callback_t) (int x_start, int x_end, int y_row, imlib_draw_row_data_t *data);
 
 // Library Hardware Init
-void imlib_init_all();
-void imlib_deinit_all();
+void imlib_init();
+void imlib_deinit();
+void imlib_update_gamma_table(float brightness, float contrast, float gamma);
```
iabdalkader: Just move this to one of the groups below, or a new one.

iabdalkader:
> This table will need to be moved to the CSI struct after brightness/contrast controls are exposed, as it technically needs to be reset on sensor reset and varies per sensor, since sensors can be controlled independently.

Yeah, I think you're right. Okay, let's do this then:
* Add the gamma table to `omv_csi_t` (enable with `OMV_CSI_GAMMA_TABLE_ENABLE` instead).
* `imlib_update_gamma_table` can be compiled in unconditionally and will just accept a pointer to the table.
* Init the table using defaults in csi init/reset or wherever you had this originally.
* No need to implement the controls right now, let's keep it simple for now.
 
Note, you will not be able to it to some  section, but it will be in DTCM for AE3 already, and SRAM for N6 (but the N6 doesn't even need this table, right?). I don't see a problem with this for now.

Please keep the imlib renames, still want that.

iabdalkader:
Also, we need to plan ahead for this. If I understand correctly, at some point we will implement set_brightness, set_contrast, and set_gamma for the PAG (for AE3), or override them for N6 (and any future MCU with an ISP), correct? If so, the table will need to be recalculated whenever any of those functions are called, right? If so, then we'll also need to save those three values in the state (for example, you may want to adjust only the brightness).

lib/imlib/imlib.h
kwagyeman: Moved it to the bayer group as it affects that.

kwagyeman:
Okay, I managed to fix the STM32 gamma correction; now, gamma can also be controlled. The STM32 has a second gamma correction step that can be enabled; however, it shouldn't be used. The STM32 ISP contrast table can handle adjustments for brightness, contrast, and gamma.

Anyway, now the AE3 and N6 both support controls for brightness, contrast, and gamma.

Per our discussion, the gamma table for the AE3 is global as we'd only use one bayer input sensor at a time. I implemented the init/reset logic as required for it and the STM32 ISP table.

boards/OPENMV_AE3/omv_boardconfig.h
```diff
@@ -47,6 +47,9 @@ extern unsigned char OMV_BOARD_UID_ADDR[12];    // Unique address.
 // GPU configuration
 #define OMV_GPU_ENABLE                  (CORE_M55_HP)
 
+// Software ISP configuration
+#define OMV_GAMMA_TABLE_ENABLE          (1)
```
iabdalkader: This should now be an IMLIB define: `IMLIB_ENABLE_GAMMA_LUT`

boards/OPENMV_AE3/omv_boardconfig.h
```diff
@@ -112,6 +115,7 @@ extern unsigned char OMV_BOARD_UID_ADDR[12];    // Unique address.
 #define OMV_OPENAMP_SIZE                (64K)
 #define OMV_GC_BLOCK0_MEMORY            SRAM0   // Extra GC block 1
 #define OMV_GC_BLOCK0_SIZE              (4M)
+#define OMV_FAST_MEMORY                 ".data" // Fast memory for quad port reads/writes.
```
iabdalkader: That's not how you add those, if you must, follow `.gpu_memory` (add `.fast_memory` section etc..), but the AE3 does Not even need this, main memory is already in DTCM, so let's just remove it, and add it if/when a board actually needs it.

lib/imlib/imlib.c
```diff
@@ -35,7 +35,15 @@
...
 #include "omv_boardconfig.h"
 
-void imlib_init_all() {
+#if (OMV_GAMMA_TABLE_ENABLE == 1)
+#if defined(OMV_FAST_MEMORY)
+OMV_ATTR_SECTION(uint8_t gamma_table[256], OMV_FAST_MEMORY);
+#else
+uint8_t gamma_table[256];
+#endif // defined(OMV_FAST_MEMORY)
+#endif // (OMV_GAMMA_TABLE_ENABLE == 1)
```
iabdalkader:
```suggestion
uint8_t gamma_table[256];
```

ports/stm32/omv_csi.c
```diff
@@ -764,12 +764,22 @@ static int stm_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
     return 0;
 }
 
+int stm_csi_isp_reset(omv_csi_t *csi) {
+    if (csi->mipi_if) {
+        #if USE_DCMIPP
+        return stm_isp_update_gamma_table(&csi->dcmipp, DCMIPP_PIPE, 0.0f, 1.0f, 2.2f);
+        #endif
+    }
```
iabdalkader:
```suggestion
    #if USE_DCMIPP
    if (csi->mipi_if) {
        return stm_isp_update_gamma_table(&csi->dcmipp, DCMIPP_PIPE, 0.0f, 1.0f, 2.2f);
    }
    #endif
```

ports/stm32/stm_isp.c
```diff
@@ -159,6 +159,39 @@ int stm_isp_config_pipeline(DCMIPP_HandleTypeDef *dcmipp, uint32_t pipe,
...
+        return -1;
+    }
+
+    return 0;
+}
+
+int stm_isp_update_gamma_table(DCMIPP_HandleTypeDef *dcmipp, uint32_t pipe,
+                               float brightness, float contrast, float gamma) {
+    // DCMIPP_ContrastConfTypeDef is not an array of uint8_t, so create an array of the same size.
+    const int g_tab_size = sizeof(DCMIPP_ContrastConfTypeDef) / sizeof(uint8_t);
```
iabdalkader:
```suggestion
    const int g_tab_size = sizeof(DCMIPP_ContrastConfTypeDef) / sizeof(uint8_t);
```
No, this struct is padded of course, so you have an extra 3 elements.

ports/stm32/stm_isp.c
```diff
@@ -159,6 +159,39 @@ int stm_isp_config_pipeline(DCMIPP_HandleTypeDef *dcmipp, uint32_t pipe,
...
+
+    for (int i = 0; i < g_tab_size; i++) {
+        // x goes from 0.0001 to 1.0 (to avoid division by zero)
+        float x = IM_MAX(i / (float) (g_tab_size - 1), 0.0001f);
+        float y = powf(x, 1.0f / gamma) * contrast + brightness;
+        int gain = fast_roundf(y / x * 16.0f);
+        g_tab[i] = IM_CLAMP(gain, 0, 32);
+    }
+
+    if (HAL_DCMIPP_PIPE_SetISPCtrlContrastConfig(dcmipp, pipe, (DCMIPP_ContrastConfTypeDef *) &g_tab) != HAL_OK) {
```
iabdalkader, line 185: Do you need to call `HAL_DCMIPP_PIPE_DisableISPCtrlContrast` first to reconfigure it?

ports/alif/omv_csi.c
```diff
@@ -401,6 +401,13 @@ int alif_csi_snapshot(omv_csi_t *csi, image_t *dst_image, uint32_t flags) {
     return 0;
 }
 
+int alif_csi_isp_reset(omv_csi_t *csi) {
```
iabdalkader: This should be added after `is_active`, same for stm.

boards/OPENMV_AE3/omv_boardconfig.h
kwagyeman: Removed.

lib/imlib/imlib.c
kwagyeman: Done

ports/alif/omv_csi.c
kwagyeman: Done

boards/OPENMV_AE3/omv_boardconfig.h
kwagyeman: Done

ports/stm32/omv_csi.c
kwagyeman: Done

ports/stm32/stm_isp.c
kwagyeman: No, it can be updated live.

ports/stm32/stm_isp.c
kwagyeman: Mmm, it's not though:

```
const int g_tab_size = sizeof(DCMIPP_ContrastConfTypeDef) / sizeof(uint8_t);
printf("g_tab_size: %d\n", g_tab_size);
```

Produces `g_tab_size: 9`.

However, changing to 9 to avoid any uncertainty.

kwagyeman:
PR updated, do you want me to add sensor brightness/contrast/gamma controls just to finish out the feature entirely?

iabdalkader:
> PR updated, do you want me to add sensor brightness/contrast/gamma controls just to finish out the feature entirely?

No, that will wait.

boards/OPENMV_AE3/imlib_config.h
```diff
@@ -184,5 +184,8 @@
 
 // Fast debayer
 #define IMLIB_ENABLE_DEBAYER_OPTIMIZATION
+
+// Software ISP configuration
+#define IMLIB_ENABLE_GAMMA_LUT
```
iabdalkader: Please move to where other LUT options are defined (near the top of the file).

ports/stm32/stm_isp.c
iabdalkader: It should be padded, anyway, yes best to be explicit.

iabdalkader: Ah actually it's because it's all bytes. Anyway, hard-coded is good for now.

boards/OPENMV_AE3/imlib_config.h
kwagyeman: Done

PR #2916, opened 2025-11-11
drivers/sensors: Add bad pixel correction for PAG7936.

kwagyeman:
Add bad pixel correction settings for Pixart Imaging for the pag7936. The N6 has BPC correct in hardware on it's ISP. This is only really needed for the AE3.

Before on N6 without BPC in N6 ISP:
image

After on N6 without BPC in N6 ISP:
image


kwagyeman:
AE3 before this:

image

And after:

image

PR #2918, opened 2025-11-15
misc: USB fixes.

iabdalkader:
- Update bootloader and MicroPython TinyUSB to version 0.19.0-24.
- Update Alif DCD (fixes sof, set_address issues etc..).
- Add DFU flash test utility (supports testing partitions with random data/verifying writes).

This is still missing a fix for the broken `--wrap` with 0.19.x.


kwagyeman (approved):
AE3 tinyusb works fine on windows, same for RT1062. Don't have the setup while traveling to test the AE3 bootloader.

I noticed this in the MP update:

image

Looks like that would affect building for the STM32U5.

iabdalkader:
Not sure how that got there, but it's not important it will be dropped anyway when the fork is updated.

PR #2923, opened 2025-11-28
imlib/fmath: Fix strict-aliasing violation in fast_expf.

andrewleech:
GCC flags fast_expf() as a strict-aliasing violation, which is currently being masked for the entire file. 

I believe this PR fixes it in a clean fashion that should be no change in behaviour / size.


iabdalkader:
LGTM, thanks! Note I'll process all pending PRs after the release.

andrewleech:
Cheers, yep no rush!

PR #2931, opened 2025-12-14
lib/imlib: Fix streaming buffer header alignment.

kwagyeman:
The frame buffer itself must always be cache line aligned so that any DMA masters writing to it can be flushed correctly.


iabdalkader:
This might actually require some protocol fixes, what if the alignment/padding changes in the future? Maybe an ioctl to get the frame header size.

iabdalkader:
Actually let's add the data `offset` to the header `(offset = sizeof(header_t))`, so in this case, it's 32. When parsing, just skip whatever the padding is.

kwagyeman:
Going to bed, modify however you like, the frame buffer data just needs to start on a cache line boundary.

iabdalkader:
Okay, will fix.

PR #2938, opened 2025-12-16
boot: Add bootloader header and bcdDevice version.

iabdalkader:
This change allows the tools to identify the bootloader both at runtime, via a bootloader header embedded in the binary image, and when the bootloader is running, via the bcdDevice version.

Tested AE3, works fine, no functional change.

### DIFF
--- a/boot/src/common/desc.c
+++ b/boot/src/common/desc.c
@@ -32,6 +32,7 @@
  */
 #include <stdint.h>
 #include <stdbool.h>
+#include "header.h"
 #include "omv_bootconfig.h"
 
 #define STR_DESC_MAX_LEN        (32)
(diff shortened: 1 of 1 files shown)


iabdalkader:
The reason for the increased flash is that the header is places at 0x1000, to accommodate all possible ISR table sizes. Note, the offset can actually be tuned per board/series to save some flash, but the bootloader is already quite small.

kwagyeman:
@iabdalkader - I don't have any thoughts about the code changes. Looks fine. Want me to test on a bunch of different boards?

iabdalkader:
> Looks fine. Want me to test on a bunch of different boards?

Sure, that would be helpful.

kwagyeman:
K, will have results later today.

kwagyeman:
F2, F7, H7, and N6 bootloaders work fine. I noticed the F2, F7, H7 bootloaders seem to include the flash fs in their binary file. This should be removed, I think.

PR #2943, opened 2025-12-19
lib/imlib: Fix raw_stream fallback to jpeg.

kwagyeman:
JPEG compression expects the destination image to have the same width and height of the source image and PIXFORMAT_JPEG. These values need to be restored after the raw_stream code runs.

Also, removed modifying the dest image size for raw_stream as this only applies for jpeg images.

### DIFF
--- a/lib/imlib/framebuffer.c
+++ b/lib/imlib/framebuffer.c
@@ -303,9 +303,8 @@ void framebuffer_update_preview(image_t *src) {
 
     if (raw_stream) {
         // Down-scale the frame (if necessary) and send the raw frame.
-        dst.size = src->bpp;
         dst.pixfmt = src->pixfmt;
-        if (src->w <= fb->raw_w && src->h <= fb->raw_h && image_size(&dst) <= available_size) {
+        if (src->w <= fb->raw_w && src->h <= fb->raw_h && image_size(src) <= available_size) {
             memcpy(dst.pixels, src->pixels, image_size(src));
         } else {
             float scale = IM_MIN((fb->raw_w / (float) src->w),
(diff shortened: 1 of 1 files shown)


lib/imlib/framebuffer.c
```diff
@@ -317,6 +316,9 @@ void framebuffer_update_preview(image_t *src) {
                                  IMAGE_HINT_BILINEAR | IMAGE_HINT_BLACK_BACKGROUND,
                                  NULL, NULL, NULL, NULL);
             } else {
+                dst.w = src->w;
+                dst.h = src->h;
+                dst.pixfmt = PIXFORMAT_JPEG;
```
iabdalkader, line 321:
```suggestion
```
But pixfmt here is not changed.

lib/imlib/framebuffer.c
```diff
@@ -303,7 +303,6 @@ void framebuffer_update_preview(image_t *src) {
 
     if (raw_stream) {
         // Down-scale the frame (if necessary) and send the raw frame.
-        dst.size = src->bpp;
```
iabdalkader, line 306: Weird, setting that doesn't make any sense.

iabdalkader (changes requested):
I also noticed that I used `image_size(&dst)`, they're the same but it would probably be less confusing to use `image_size(&src)`.

lib/imlib/framebuffer.c
kwagyeman: It's set to `dst.pixfmt = src->pixfmt;` on line 306.

The source pixformat is mutable, so it's definitely not a JPEG.

lib/imlib/framebuffer.c
kwagyeman: Yeah, wasn't sure why that was there. Removing it saves having to reset the size back to available_size later.

kwagyeman:
> I also noticed that I used image_size(&dst), they're the same but it would probably be less confusing to use image_size(&src).

Update?

lib/imlib/framebuffer.c
iabdalkader: Right.

iabdalkader:
> Update?

Yes, we should update it.

kwagyeman:
Retesting

kwagyeman:
K, it's good.

PR #2948, opened 2025-12-21
boot: Fix DFU detach not exiting main loop on some hosts.

iabdalkader:
On some hosts, if a thread is polling USB, the device remains mounted after dfu-util --reset sends a DFU_DETACH request. This causes the bootloader to stay in the main loop in forced mode instead of resetting.

### DIFF
--- a/boot/src/common/main.c
+++ b/boot/src/common/main.c
@@ -33,10 +33,11 @@
 #include <string.h>
 #include <stdint.h>
 #include <stdbool.h>
-#include "header.h"
 #include "omv_bootconfig.h"
+#include "dfu.h"
+#include "header.h"
 
-bool tud_dfu_detached = true;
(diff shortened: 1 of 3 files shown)


boot/src/common/main.c
```diff
@@ -47,15 +48,8 @@ const omv_boot_header_t omv_boot_header = {
...
-    uint32_t current_ms = port_ticks_ms();
-    if (current_ms >= start_ms) {
-        return current_ms - start_ms;
-    } else {
-        // Handle wraparound
-        return (UINT32_MAX - start_ms) + current_ms + 1;
-    }
-}
+#define check_timeout_ms(start_ms, timeout) \
+    ((port_ticks_ms() - (start_ms)) > (timeout))
```
kwagyeman, line 52: Nice update here.

kwagyeman:
Tested on windows, mac, and linux with multiple boards. No issues.

PR #2951, opened 2025-12-28
modules: Add getting custom framesize.

### LINKED ISSUE #2950: Missing way to get custom frame size resolution passed.

```
...
csi.framesize((320, 320))
print(csi.framesize())
```

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/2950


iabdalkader:
Why not use `width()` and `height()` if you need the dimensions? This breaks the documented API we worked hard to keep clean.

image

iabdalkader:
Alternatively, we could add a new function like `resolution()` -> tuple. Note, I see some deeper related issues here.

iabdalkader:
See #3258

kwagyeman:
Works for me!

PR #2983, opened 2026-03-02
ports/stm32: Fix N6 bootloader.

iabdalkader:
* Revert working external loader.
* Add -align flag to sign tool.


kwagyeman:
@iabdalkader - Tested and works.

PR #2985, opened 2026-03-08
common: Make frame sync pin configurable by camera drivers.

kwagyeman:
Refactors the camera driver fsync trigger so that the camera driver controls which pin the trigger pin appears on.

This is necessary as the FSYNC pin for older legacy cameras is on the SPI_MISO pin for the MT9V034. I don't remember how that happened anymore... but, it is what is.

New camera drivers can now set the trigger pin on init, allowing it to be moved to the FSYNC pin on the DCMI header.

...

This PR was pulled out of https://github.com/openmv/openmv/pull/2564.

...

Tested on the N6, RT1062, H7 Plus, and H7 with the MT9V034 sensor.


iabdalkader:
I don't quite understand the issue, but there should be a default fsync pin, set in init if defined, then sensor drivers can override it, and gpio configured in `omv_csi_probe`. Saves us from defining the same pin for every other sensor.

kwagyeman:
Okay, works for me. The default pin should be P10 but I have to override for the MT9V034.

kwagyeman:
Will process tomorrow.

kwagyeman:
> I don't quite understand the issue, but there should be a default fsync pin, set in init if defined, then sensor drivers can override it, and gpio configured in omv_csi_probe. Saves us from defining the same pin for every other sensor.

Done and retested on the N6, RT1062, H7, and H7P with the MT9V034.

common/omv_csi.c
```diff
@@ -176,9 +176,6 @@ __weak int omv_csi_init() {
     #if defined(OMV_CSI_RESET_PIN)
     omv_gpio_config(OMV_CSI_RESET_PIN, OMV_GPIO_MODE_OUTPUT, OMV_GPIO_PULL_NONE, OMV_GPIO_SPEED_LOW, -1);
```
iabdalkader, line 177: For consistency, should we move other pins to csi_t too?

common/omv_csi.c
```diff
@@ -193,6 +190,9 @@ __weak int omv_csi_init() {
         csi->clk = &csi_clk;
         csi->fb = framebuffer_get(FB_MAINFB_ID);
         csi->color_palette = rainbow_table;
+        #if defined(OMV_CSI_FSYNC_PIN)
+        csi->fsync_pin = OMV_CSI_FSYNC_PIN;
```
iabdalkader, line 194: Why move `omv_gpio_config` to probe?

boards/OPENMV4P/board_config.h
```diff
@@ -231,7 +233,6 @@
 #define OMV_CSI_PXCLK_PIN                     (&omv_pin_A6_DCMI)
 #define OMV_CSI_RESET_PIN                     (&omv_pin_A10_GPIO)
 #define OMV_CSI_POWER_PIN                     (&omv_pin_D7_GPIO)
-#define OMV_CSI_FSYNC_PIN                     (&omv_pin_B4_GPIO)
```
iabdalkader, line 234: Where's the default fsync for these boards then?

iabdalkader (changes requested):

common/omv_csi.c
kwagyeman: We could, but these don't have any legacy issues that require it. So, it would more or less be unused.

But if you still want it, then I'll make an issue for it after v5.0.0 and resolve it in a new PR. Let me know.

boards/OPENMV4P/board_config.h
kwagyeman: They don't physically have a pin connecting to FSYNC. The FSYNC pin on the older boards just connects to the camera module itself. It doesn't actually go to the MCU. This is why the MT9V034's trigger ended up on a SPI bus pin on the camera sensor header.

common/omv_csi.c
kwagyeman: > I don't quite understand the issue, but there should be a default fsync pin, set in init if defined, then sensor drivers can override it, and gpio configured in omv_csi_probe. Saves us from defining the same pin for every other sensor.

You asked me to? I had it in the driver itself. But, moving to probe is a better idea.

iabdalkader: I think the pin config should be where it was, in init.

iabdalkader: Ah right, we need to detect the sensor first.

iabdalkader (approved):

PR #2986, opened 2026-03-08
boards/OPENMV_N6: Add missing fballoc overlay.

kwagyeman:
Provides a massive 3x speedup for algorithms like AprilTags at QVGA resolution by moving some fballoc allocations on-chip versus leaving allocations on SDRAM.


iabdalkader:
Is this the absolute minimum memory needed for this? I wanted to save some spare memory for the future, can't use all of the remaining memory.

kwagyeman:
Reduced it to 400 KB. There's still about 35KB free. Didn't see any performance reductions with april tags.

PR #2987, opened 2026-03-08
protocol: Fix race condition with locking.

kwagyeman:
On our older OpenMV Cams the usb protocol library runs USB interrupts. Which means that stream actions can happen while framebuffer_update_preview() is running in the main thread. On the OpenMV M4, it so happens that the stream checking size() while framebuffer_update_preview() is updating the header happens regularly causing a system crash.

The fix for this is to treat stream->size() as volatile until locked.

...

Did not notice the issue except on the old Cortex M4.

### DIFF
--- a/protocol/omv_protocol_channel_profile.c
+++ b/protocol/omv_protocol_channel_profile.c
@@ -46,8 +46,17 @@ static size_t profile_channel_shape(const omv_protocol_channel_t *channel, size_
 }
 
 static int profile_channel_lock(const omv_protocol_channel_t *channel) {
+    mutex_t *lock = omv_profiler_lock();
+    if (!mutex_try_lock(lock, MUTEX_TID_IDE)) {
+        return -1;
+    }
+    // Size can be checked safetly after acquiring the lock.
     size_t size = channel->size(channel);
-    return size && mutex_try_lock(omv_profiler_lock(), MUTEX_TID_IDE) ? 0 : -1;
+    if (!size) {
+        mutex_unlock(lock, MUTEX_TID_IDE);
+        return -1;
+    }
+    return 0;
 }
 
 static int profile_channel_unlock(const omv_protocol_channel_t *channel) {
--- a/protocol/omv_protocol_channel_stream.c
+++ b/protocol/omv_protocol_channel_stream.c
@@ -40,9 +40,16 @@ static bool stream_channel_poll(const omv_protocol_channel_t *channel) {
 
 static int stream_channel_lock(const omv_protocol_channel_t *channel) {
     framebuffer_t *fb = framebuffer_get(FB_STREAM_ID);
+    if (!mutex_try_lock(&fb->lock, MUTEX_TID_IDE)) {
+        return -1;
+    }
+    // Size can be checked safetly after acquiring the lock.
     size_t size = channel->size(channel);
-    // Attempt locking only if the stream is ready.
-    return size && mutex_try_lock(&fb->lock, MUTEX_TID_IDE) ? 0 : -1;
+    if (!size) {
+        mutex_unlock(&fb->lock, MUTEX_TID_IDE);
+        return -1;
+    }
+    return 0;
 }
 
 static int stream_channel_unlock(const omv_protocol_channel_t *channel) {


iabdalkader:
> it so happens that the stream checking size() while framebuffer_update_preview() is updating the header happens regularly causing a system crash.

Why would this cause a crash? Even if size is invalid, it's only used to check if there might be something to read, Not to do the actual reading. A following `profile_channel_size` should return the actual size, or 0. And even if you pass a wrong size/offset to read it shouldn't crash.

I don't think this fixes anything, the only thing it does is make `lock` slower because now it will always lock/unlock even if there's nothing to read.

kwagyeman:
The firmware doesn't quite crash. This is what I saw on the M4:

1. Using OpenMV IDE would be good initially.
2. Then the video would freeze.
3. Clicking the disable/enable frame buffer button would sometimes restore the video. Sometimes not. Starting or stopping the script would not resolve the issue.
4. Text output continued to work, though.

After seeing this, I turned on debugging in the IDE and traced the issue down to the frame buffer LOCK command returning NACK. I saw that the IDE would get a NACK if it requested more frequently than 10ms. So, added a hold off then to ensure requests are spaced by more than 10ms. Still got a NACK.

After that, I added debug print statements to the camera firmware and saw that the LOCK function in the stream interface was happily working normally... and then all of a sudden the size would always become 0, permanently causing LOCK to fail.

Then, on digging into that, I noticed the size function was returning 0.

... So, after seeing that, I noticed the race condition with LOCK, fixed it, and the issue stopped happening.

iabdalkader:
> I saw that the IDE would get a NACK if it requested more frequently than 10ms.

That's because locking is throttled with a default min interval of 10ms:
https://github.com/openmv/openmv/blob/420102a1d9ac9abdc7512d71ef395f964d05b013/protocol/omv_protocol.c#L779

> ... So, after seeing that, I noticed the race condition with LOCK, fixed it, and the issue stopped happening.

There's no race condition, the lock can't be acquired twice in a row by the same thread, and the protocol clears the header before unlocking. So if the size is set, this means `framebuffer_update_preview` just ran and will not be able to lock again until the protocol reads and clears the frame.

This might be a different issue, this change here only makes every lock slower on all cams, maybe masking the real issue, I think we should revert it.

kwagyeman:
I fixed the IDE already not to violate the 10ms rule and that does nothing. I even traced the code and that error is not being hit after that.

The race condition is that you read the size value which is computed from a function before locking. The size is computed from the header which is updated during the jpeg update. Given this, it shouldn't be read unless you have the lock as it's written as parts which could result in it being partially written if an interrupt happens during its update.

The M4 still uses the old stm32 usb code which is interrupt driven meaning that this could run during the frame buffer header update.

iabdalkader:
> The race condition is that you read the size value which is computed from a function before locking. The size is computed from the header which is updated during the jpeg update.

The size is only used to check if there's a valid frame. If there's a valid frame (size !=0), it means `framebuffer_update_preview` held the lock last (ran twice in a row) so it can't lock again.

One scenario:

USB IRQ -> `framebuffer_update_preview()` -> locks, updates header, unlocks
Protocol -> Read size -> 5K (frame valid)
USB IRQ again -> `framebuffer_update_preview` -> `try_lock_fair` fails because it held it last
Protocol -> try_lock -> channel_size -> 5k

I don't see any problem with this. Can you provide a similar trace that shows how this could potentially fail? Otherwise, I really think we should revert it and try to figure out the root cause. It could be the atomic implementation on older M4.

iabdalkader:
Hmm I think I see the issue now. The channel unlock function unlocks first, then `memset(0)` the header. If it gets interrupted in between those steps this might be a problem, because with the old code `size` will be zero and the channel won't lock again, but `framebuffer_update_preview()` can't lock until the channel reads that lost frame. This is in line with what you've described.

Protocol -> `mutex_unlock()`
USB IRQ -> framebuffer_update_preview() -> locks, updates header, unlocks (size != 0)
Protocol -> `memset(0)` <- frame lost
Now protocol can't lock because `size == 0`, but `framebuffer_update_preview` can't either, because it needs protocol to lock first.

This fix works because you moved the `size` check inside the locking, which unblocks the `framebuffer_update_preview()`.

iabdalkader:
There's still an issue here, a frame is lost occasionally with the right timing. I think the fix is to revert this and bring back that "owned" function you had, and only memset(0) if we held the lock. I don't like that function though, so maybe just set a flag somewhere?

Maybe something like this?
```C
static int stream_channel_lock(const omv_protocol_channel_t *channel) {
    framebuffer_t *fb = framebuffer_get(FB_STREAM_ID);
    size_t size = channel->size(channel);
    // Attempt locking only if the stream is ready.
    channel->locked = size && mutex_try_lock(&fb->lock, MUTEX_TID_IDE);
    return channel->locked ? 0 : -1;
}

static int stream_channel_unlock(const omv_protocol_channel_t *channel) {
    framebuffer_t *fb = framebuffer_get(FB_STREAM_ID);
    if (channel->locked) {
        channel->locked = false;
        memset(fb->raw_base, 0, sizeof(framebuffer_header_t));
        mutex_unlock(&fb->lock, MUTEX_TID_IDE);
    }
    return 0;
}
```

Anyway, @kwagyeman great find!

kwagyeman:
> Hmm I think I see the issue now. The channel unlock function unlocks first, then `memset(0)` the header. If it gets interrupted in between those steps this might be a problem, because with the old code `size` will be zero and the channel won't lock again, but `framebuffer_update_preview()` can't lock until the channel reads that lost frame. This is in line with what you've described.
> 
> Protocol -> `mutex_unlock()` USB IRQ -> framebuffer_update_preview() -> locks, updates header, unlocks (size != 0) Protocol -> `memset(0)` <- frame lost Now protocol can't lock because `size == 0`, but `framebuffer_update_preview` can't either, because it needs protocol to lock first.
> 
> This fix works because you moved the `size` check inside the locking, which unblocks the `framebuffer_update_preview()`.

This seems to be exactly what I was seeing. Size goes to 0 and then everything is stuck.

...

Yeah, your suggestion may work. I can't really test until I get home though as I didn't bring any M4s with me. Alternatively, instead of reverting a new PR could be made to improve this and then i can test when home. At least it doesn't have a deadlock bug at the moment, albeit it may drop valid frames. Folks may run into the issue on the non-M4 cams even though I didn't see it.

kwagyeman:
I haven't forgot this, should have time tommorrow to do what was asked.

iabdalkader:
I'm waiting on a proper fix for this issue, before looking at anything else.

kwagyeman:
@iabdalkader - I'm not able to generate the error with the H7/N6 plus even after adding a small delay as suggested.

I tried your fix also and saw no issue. Maybe just revert this PR for now? When I get home next week, I will test out your fix on the M4. This issue doesn't seem to affect other cameras. Maybe it could be something else related to the M4 arch.

kwagyeman:
Okay, I tried this fix:

```
static int stream_channel_lock(const omv_protocol_channel_t *channel) {
    framebuffer_t *fb = framebuffer_get(FB_STREAM_ID);
    size_t size = channel->size(channel);
    // Attempt locking only if the stream is ready.
    channel->locked = size && mutex_try_lock(&fb->lock, MUTEX_TID_IDE);
    return channel->locked ? 0 : -1;
}

static int stream_channel_unlock(const omv_protocol_channel_t *channel) {
    framebuffer_t *fb = framebuffer_get(FB_STREAM_ID);
    if (channel->locked) {
        channel->locked = false;
        memset(fb->raw_base, 0, sizeof(framebuffer_header_t));
        mutex_unlock(&fb->lock, MUTEX_TID_IDE);
    }
    return 0;
}
```

On the M4 and the error still happened. The changes did nothing from the baseline.

Found the actual issue in: https://github.com/openmv/openmv/pull/3030

PR #2998, opened 2026-03-16
ps5520: Stabilize AEC/AGC control loop.

### LINKED ISSUE #2995: Fix issue with PS5520 Oscillating Still
This was mostly fixed with the AGC smoothing, but, still an issue.

aqsnyder:
Fixes #2995

The `update_agc_aec` control loop oscillates at certain lighting conditions due to three issues in the Pixart-authored code:

1. **No dead zone** — the loop adjusts on any luminance error (unstable with noise)
2. **No output damping** — the full proportional step is applied every frame, causing overshoot
3. **No AEC↔AGC hysteresis** — the loop togles rapidly between exposure and gain adjustment at the gain floor boundary

This patch adds:
- A dead zone threshold (`PS5520_L_AEC_DEADZONE = 4`) to ignore noise
- EMA damping (÷4 per frame) on exposure adjustments to prevent overshoot
- Hysteresis (`PS5520_L_AGC_HYST = 2`) on the AGC→AEC transition

The original `PS5520_L_AEC_DIFF_MUL` is preserved — stability comes from the damping and dead zone, not by changing Pixart's proportional gain.

### DIFF
--- a/drivers/sensors/ps5520.c
+++ b/drivers/sensors/ps5520.c
@@ -93,6 +93,7 @@ static int g_div = 1;
 #define PS5520_L_TARGET         (80)
 #define PS5520_L_AGC_DIFF_DIV   (10)
 #define PS5520_L_AEC_DIFF_MUL   (10)
+#define PS5520_L_AEC_MAX_STEP   (50)
 
 static bool enable_agc = true;
 static int32_t agc_gain = PS5520_DEF_GAIN;
@@ -1364,6 +1365,7 @@ static int set_auto_gain(omv_csi_t *csi, int enable, float gain_db, float gain_d
 
         agc_gain = ((idx - 1) << 4) + ((gain >> (idx - 1)) & 0x0F);
 
+
         ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, REG_BANK, 1, 0x01, 1);
         ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, CMD_GAIN_IDX, 1, agc_gain, 1);
         ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, SENSOR_UPDATE, 1, 0x01, 1);
@@ -1408,6 +1410,7 @@ static int set_auto_exposure(omv_csi_t *csi, int enable, int exposure_us) {
 
         int32_t exposure_line = ConvertT2L(exposure_us);
         aec_exposure = IM_CLAMP(exposure_line, PS5520_MIN_INT, (lpf - 2));
+
 
         ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, CMD_OFFNY1_H, 1, (lpf - aec_exposure) >> 8, 1);
         ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, CMD_OFFNY1_L, 1, (lpf - aec_exposure) & 0xFF, 1);
@@ -1491,7 +1494,17 @@ static int update_agc_aec(omv_csi_t *csi, int luminance) {
     int ret = 0;
     int diff = PS5520_L_TARGET - luminance;
 
-    if (abs(diff) > 0) {
+    // Dead zone: ignore small luminance deviations. With raw (un-smoothed)
+    // luminance, frame-to-frame sensor noise is not filtered, so a small
+    // dead zone prevents noise-driven corrections that appear as flicker.
+    if (abs(diff) > 4) {
+        // If exposure and gain are both floored and scene is still too
+        // bright, skip the adjustment — nothing can change and redundant
+        // I2C writes may cause sensor glitches.
+        if (diff < 0 && aec_exposure <= PS5520_MIN_INT && agc_gain <= PS5520_MIN_GAIN_IDX) {
+            return 0;
+        }
+
         bool aec_exposure_in = ((diff > 0) && (aec_exposure < aec_exposure_ceiling)) ||
                                ((diff < 0) && (agc_gain <= PS5520_MIN_GAIN_IDX));
 
@@ -1504,7 +1517,15 @@ static int update_agc_aec(omv_csi_t *csi, int luminance) {
             bool flg_stall = 0;
             int32_t exposure_line;
 
-            aec_exposure += diff * PS5520_L_AEC_DIFF_MUL;
+            // Continuous step scaling: max_step is proportional to how
+            // far the current luminance is from either saturation extreme
+            // (0 or 255). This avoids step-size discontinuities that cause
+            // oscillation at zone boundaries.
+            int32_t headroom = IM_MIN(luminance, 255 - luminance);
+            int32_t max_step = IM_MAX(1, IM_MIN(PS5520_L_AEC_MAX_STEP, headroom / 10));
+            int32_t step = diff * PS5520_L_AEC_DIFF_MUL;
+            step = IM_CLAMP(step, -max_step, max_step);
+            aec_exposure += step;
             aec_exposure = IM_CLAMP(aec_exposure, PS5520_MIN_INT, aec_exposure_ceiling);
 
             if (aec_exposure > (VTS_5M_30 - 1 - 2)) {
@@ -1552,7 +1573,9 @@ static int update_agc_aec(omv_csi_t *csi, int luminance) {
             }
 
         } else if (enable_agc) {
-            agc_gain += diff / PS5520_L_AGC_DIFF_DIV;
+            int32_t gain_step = diff / PS5520_L_AGC_DIFF_DIV;
+            gain_step = IM_CLAMP(gain_step, -1, 1);
+            agc_gain += gain_step;
             agc_gain = IM_CLAMP(agc_gain, PS5520_MIN_GAIN_IDX, agc_gain_ceiling);
 
             //ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, REG_BANK, 1, 0x01, 1);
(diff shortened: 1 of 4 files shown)


iabdalkader:
Ignore MPS3 on FVP, it's new and it's a bit finicky.

drivers/sensors/ps5520.c
```diff
@@ -1555,7 +1569,6 @@ static int update_agc_aec(omv_csi_t *csi, int luminance) {
             agc_gain += diff / PS5520_L_AGC_DIFF_DIV;
             agc_gain = IM_CLAMP(agc_gain, PS5520_MIN_GAIN_IDX, agc_gain_ceiling);
 
-            //ret |= omv_i2c_write_reg(csi->i2c, csi->slv_addr, REG_BANK, 1, 0x01, 1);
```
kwagyeman, line 1558: Why is this removed?

drivers/sensors/ps5520.c
```diff
@@ -1504,7 +1511,14 @@ static int update_agc_aec(omv_csi_t *csi, int luminance) {
...
 
-            aec_exposure += diff * PS5520_L_AEC_DIFF_MUL;
+            // EMA-damped exposure step: apply only 1/4 of the raw proportional
+            // adjustment per frame to prevent overshoot and oscillation.
+            int32_t raw_step = diff * PS5520_L_AEC_DIFF_MUL;
+            int32_t damped_step = (raw_step + 2) / 4;  // +2 for rounding
+            if (damped_step == 0) {
+                damped_step = (diff > 0) ? 1 : -1;
+            }
+            aec_exposure += damped_step;
```
kwagyeman: Can you implement this like how the EMA filter is implemented for the rest of the AGC code? See https://github.com/openmv/openmv/blob/master/common/omv_csi.c#L1168

There's a very specific mixing factor based on time deltas.

kwagyeman: Note that you'll need to have some history in the driver. So, a variable to keep the last time. Make sure to reset that at the same time the aec_exposure value is reset too.

kwagyeman:
I can review in the afternoon in London tomorrow.

aqsnyder:
with AEC/AGC control
https://github.com/user-attachments/assets/c0dd27c8-3b60-4429-8cf0-fa90ac50c90c

without AEC/AGC control
https://github.com/user-attachments/assets/1c375f14-926d-4f1f-992f-c8c9a567cd72

kwagyeman:
@aqsnyder - can you address my feedback, also, please squash into one commit. I see the videos, results look good.

But, move a time based ema filter as I suggested so that we are not tweaking the gain endlessly.

aqsnyder:
> @aqsnyder - can you address my feedback, also, please squash into one commit. I see the videos, results look good.
> 
> But, move a time based ema filter as I suggested so that we are not tweaking the gain endlessly.

Yes that makes sense. I can do both of these things. I did not see any inline comments about this.

drivers/sensors/ps5520.c
```diff
@@ -1491,9 +1499,30 @@ static int update_agc_aec(omv_csi_t *csi, int luminance) {
...
+    if (!aec_ema_initialized) {
+        aec_ema_initialized = true;
+        aec_ema_last_ms = now_ms;
+        aec_ema_diff = (float) diff;
+    } else {
+        uint32_t dt_ms = now_ms - aec_ema_last_ms;
+        float alpha = IM_CLAMP(1.0f - expf(-((float) dt_ms) / PS5520_L_AEC_TAU_MS), 0.0f, 1.0f);
+        aec_ema_last_ms = now_ms;
+        aec_ema_diff += alpha * ((float) diff - aec_ema_diff);
+    }
```
kwagyeman: Reset these during reset also... and whenever anything related to the aec_expsorue/gain is etting is changed manually.

drivers/sensors/ps5520.c
```diff
@@ -102,6 +105,11 @@ static bool enable_aec = true;
 static int32_t aec_exposure = PS5520_DEF_EXP;
 static int32_t aec_exposure_ceiling = PS5520_MAX_INT;
 
+// Time-based EMA state for AEC smoothing.
+static bool aec_ema_initialized = false;
```
iabdalkader: We should group these variables in a struct, along with the other global variables, accessible via `csi->priv` similar to https://github.com/openmv/openmv/blob/343ec865e33b7d0eacaebce7ea07f58682c76e78/drivers/sensors/genx320.c#L90

For example:
```C
typedef struct ps5520_state {
    bool enable_agc;
    int32_t agc_gain;
    int32_t agc_gain_ceiling;

    bool enable_aec;
    int32_t aec_exposure;
    int32_t aec_exposure_ceiling;

    bool aec_ema_initialized;
    uint32_t aec_ema_last_ms;
    float aec_ema_diff;
} ps5520_state_t;

int ps5520_init(omv_csi_t *csi) {
    //....
    csi->priv = &ps5520;
}

    ps5520_state_t *ps5520 = csi->priv;
    memset(ps5520, 0, sizeof(ps5520_state_t));
   
    ps5520->enable_agc = true;
    ps5520->agc_gain = PS5520_DEF_GAIN;
    ps5520->agc_gain_ceiling = PS5520_DEF_GAINCEILING;

    ps5520->enable_aec = true;
    ps5520->aec_exposure = PS5520_DEF_EXP;
    ps5520->aec_exposure_ceiling = PS5520_DEF_EXP_CEILING;
}
```
Not necessarily in this PR.

kwagyeman: yes!

aqsnyder: okay, that can be done. ignore my last commit. I just needed to build the new ema approach so I could actually test outside while I have bright light.

kwagyeman:
@aqsnyder - I'm back home on Friday. Do you think you will finish with your testing and changes for the PR by then?

aqsnyder:
> @aqsnyder - I'm back home on Friday. Do you think you will finish with your testing and changes for the PR by then?

Yes, I should have time this week to get back to finishing the PR. I will keep you posted.

aqsnyder:
Im starting to think the core issue does not lie in the ps5520 driver. It might be in interaction between the frameworks omv_csi_stats_update() EMA smoothing (τ=250ms) and the AEC control loop. 

I ended up removing the EMA from the ps5520 driver since it is already present in stm_isp_update_awb() → omv_csi_stats_update(), so I was just double smoothing.

I think the EMA introduces enough lag in the luminance measurement that during dark→bright transitions (like my failing outdoor pictures) the lag causes the AEC controller to initially correct in the wrong direction for 2-3 frames. the smoothed luminance still reads below target even though the scene is now very bright, so the controller increases exposure when it should be decreasing it. This amplifies overshoot and creates oscillation between dark and properly-exposed frames.

The current commit adds only a MAX_STEP clamp on the AEC exposure adjustment to limit extreme single-frame jumps. This improves stability somewhat but doesn't fully resolve the outdoor bright-light oscillation.

Should we provide the driver with access to the raw (un-smoothed) luminance for AEC control while keeping the EMA-smoothed values for AWB? Or should the EMA tau be configurable per-sensor? The EMA smoothing is great for AWB stability but works against fast AEC response during rapid scene transitions.

kwagyeman:
Probably the raw unsmoothed value should be available. Go ahead and modify the api to include that in the ioctl. The camera and the awb ema should be independent (the pag7936 AGC and AEC runs on the camera chip for example and doesn't care what awb is doing. The same applies for the ps5520 where we have to do AGC and aec ourselves).

Please handle all paths that need to be updated for this.

aqsnyder:
exp setup:
setup

Original firmware:
https://github.com/user-attachments/assets/8c194da5-11b0-4d77-8099-d11cb781eb00

PR fix:
https://github.com/user-attachments/assets/babe6c35-69de-4417-8845-87da1c3bc1d1

added a raw_luminance out-parameter to stm_isp_update_awb() that captures the luminance before the EMA filter is applied. The AEC ioctl now receives this raw value, while AWB continues using the smoothed stats as before.

On the PS5520 driver side, I added:
* Continuous step scaling based on distance from saturation (smaller steps near extremes, full proportional response in the normal range)
* A small dead zone (4) to filter out raw sensor noise
* Early return when exposure/gain are floored and scene is still too bright (avoids redundant I2C writes)
* AGC step clamped to ±1 per frame

aqsnyder:
> exp setup: setup
> 
> Original firmware: https://github.com/user-attachments/assets/8c194da5-11b0-4d77-8099-d11cb781eb00
> 
> PR fix: https://github.com/user-attachments/assets/babe6c35-69de-4417-8845-87da1c3bc1d1
> 
> added a raw_luminance out-parameter to stm_isp_update_awb() that captures the luminance before the EMA filter is applied. The AEC ioctl now receives this raw value, while AWB continues using the smoothed stats as before.
> 
> On the PS5520 driver side, I added:
> 
> * Continuous step scaling based on distance from saturation (smaller steps near extremes, full proportional response in the normal range)
> * A small dead zone (4) to filter out raw sensor noise
> * Early return when exposure/gain are floored and scene is still too bright (avoids redundant I2C writes)
> * AGC step clamped to ±1 per frame

The above test was a controlled setup. But the real test is using the camera in bright outdoor shots. This is where the heavy oscillations tend to show up. In bright light settings I see 2-3 frame of slight flicker before stabilization.

kwagyeman:
Amazing! So, is this ready for review or do you want to test some more?

aqsnyder:
> Amazing! So, is this ready for review or do you want to test some more?

ready for review

kwagyeman:
Okay, I can do so on Friday. However, Ibrahim and I already both left comments. Please resolve and close those while waiting.

drivers/sensors/ps5520.c
iabdalkader: This is optional, we can clean it up in a follow up PR to minimize the diff.

kwagyeman:
Moved it here and also did the static var cleanup. Works great. https://github.com/openmv/openmv/pull/3043

Issue #3005, opened 2026-03-19
AE3 AprilTags crashes


biasedlogic:
**Describe the bug**
The all-new AE3 just landed, yay!
However, when I try to run my existing code, as well as examples from the IDE, I hit issues.
One is processing AprilTags which I use in my robotics class. 
Whenever I execute img.find_apriltags() I get **MemoryError: Out of fast frame buffer stack memory** and the script crashes.

**To Reproduce**
1. Connect AE3 to your IDE
2. Load Examples > April Tags > find_apriltags.py
3. replace QQVGA with QVGA as the AE3 does not support QQVGA resolution
4. Start the script

**Expected behavior**
The script should run its loop until it detects an aprilTag in the captured image.

**Screenshots**
<img width="740" height="250" alt="Image" src="https://github.com/user-attachments/assets/de5966b7-db44-4730-8f6a-cfa8d3a5ec97" />

**Please complete the following information**
 - IDE version: 4.8.7
 - Firmware version: either:
    - OpenMV ebf04bfa4b; MicroPython 082381af4d; OpenMV-AE3 with AE302F80F55D5AE (release as of 19.MAR.26)
    - OpenMV v4.8.1; MicroPython v1.26.0-77; OpenMV-AE3 with AE302F80F55D5AE (latest 19.MAR.26)

 - Board: AE3
 - OS: Windows 11

kwagyeman:
Fixed. See attached binary. 

[firmware_M55_HP.zip](https://github.com/user-attachments/files/26125504/firmware_M55_HP.zip)

Load with the IDE (after unzipping the binary file) Tools->Load Custom Firmware Image.

...

Also, turning off the PAG7936 auto gain and auto exposure doesn't work right now. Just comment out those lines. that fix will be a bit longer to make.

kwagyeman:
I've added some code to the IDE to upscale QQVGA to QVGA on the AE3 in the future.

biasedlogic:
Testing it now.
Using this code:
```
import sensor
import time
import math

sensor.reset()
sensor.set_pixformat(sensor.RGB565)
sensor.set_framesize(sensor.QVGA)
sensor.skip_frames(time=2000)
sensor.set_auto_gain(False)  # must turn this off to prevent image washout...
sensor.set_auto_whitebal(False)  # must turn this off to prevent image washout...
clock = time.clock()

while True:
    clock.tick()
    img = sensor.snapshot()
    for tag in img.find_apriltags():
[...]

PR #3006, opened 2026-03-19
boards/OPENMV_AE3: Fix Apriltags on the AE3.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/3005


iabdalkader:
I've found the underlying issue here, and it goes way deeper than this. This PR doesn't fix it, but we still need some of the changes here. Please revert DTCM changes and keep the rest if they're important, for example if we can't support the high-res.

PR #3007, opened 2026-03-20
imlib/apriltag: Fix cache corruption.

iabdalkader:
fb_alloc packs allocations tightly, so the stashed 4-byte size field can share a cache line with adjacent allocation data. In this case the data is the temporary image buffers which get passed to GPU and to `SCB_InvalidateDCache_by_Addr`, which aligns the address causing size writes to be lost.

This is one of the issues causing https://github.com/openmv/openmv/issues/3005. Note that currently the GPU buffer is actually in DTCM (uncacheable) but we can't move it to any other SRAM without this fix.


iabdalkader:
Added prefer_size to try to keep the GPU working.

PR #3009, opened 2026-03-20
lib/imlib: Cache align image allocations that are drawn on.

kwagyeman:
These allocations are all drawn on my imlib_draw_image, which may use the GPU to draw the image. As such, the image allocation must be cache-aligned to prevent corruption when the cache is flushed.

I used prefer size for the allocations which are used to create a grayscale image for an algorithm from the frame buffer. Otherwise, I just used cache-align.

### DIFF
--- a/lib/alif/Device/common/source/dcd.c
+++ b/lib/alif/Device/common/source/dcd.c
@@ -51,7 +51,7 @@ static uint32_t  _evnt_buf[1024] CFG_TUSB_MEM_SECTION __attribute__((aligned(409
 static volatile uint32_t* _evnt_tail;
 
 static uint8_t  _ctrl_buf[64] CFG_TUSB_MEM_SECTION __attribute__((aligned(32))); // [TODO] runtime alloc
-static uint32_t _xfer_trb[8][4] CFG_TUSB_MEM_SECTION __attribute__((aligned(32))); // [TODO] runtime alloc
+static uint32_t _xfer_trb[8][8] CFG_TUSB_MEM_SECTION __attribute__((aligned(32))); // [TODO] runtime alloc
 static uint16_t _xfer_bytes[8];
 static bool     _ctrl_long_data = false;
 static bool     _xfer_cfgd = false;


iabdalkader:
I think we should wait for further tests with BKPT in cache invalidate.

kwagyeman:
These ones actually break in the AE3. You get a crash for example from find_line_segments() on the AE3.

I can test with the breakpoint stuff when I get home. The second I add it to the code the AE3 doesn't boot at all anymore. So, I don't have the tools to debug it right now.

iabdalkader:
> These ones actually break in the AE3. You get a crash for example from find_line_segments() on the AE3.

That's why we need to wait until this is tested, there will definitely be other places where unaligned pointers are passed, some may be fine some may need more fixes.

> So, I don't have the tools to debug it right now.

That's why I never leave home without a debugger.

kwagyeman:
So... I found the issue on alif why it's crashing after adding a BKPT on non-32-byte aligned addresses:

`static uint32_t _xfer_trb[8][4] CFG_TUSB_MEM_SECTION __attribute__((aligned(32))); // [TODO] runtime alloc`

So, in the tusb_alf_dcd.c driver, stuff like `RTSS_CleanDCache_by_Addr(_xfer_trb[ep], sizeof(_xfer_trb[ep]));` is happing all the time. However, if ep is 1, then this is a 16-byte-aligned address, and `sizeof(_xfer_trb[ep])` is 16. The code sprinkles RTSS_InvalidateDCache_by_Addr and RTSS_CleanDCache_by_Addr all over... all 32-byte alignment and size requirements.

It's amazing that this code is working. Maybe because only one endpoint is operational at a time, and because we don't use endpoint 7?

The driver needs to be fixed to add __BKPT for non-32-byte-aligned addresses in the SCB functions.

...

On the N6, or any STM32 camera with a uSD card, with the SDIO driver in `sdcard_read_blocks` clean and invalidates the destination where it's writing to... however, that destination was allocated in `fat_vfs_make_new` via `mp_obj_malloc` without any alignment using the `fs_user_mount_t` object. So, that's corrupt too.

Without an SD card the N6 is fine. I can stream video and run Apriltags. Running ML algorithms works too.

...

On the RT1060 with an SD, it's bad too. Same exact error as the N6. A non-cache-aligned block is passed to the uSDHC driver, which then cleans/invalidates the block.

Without an SD card the RT1062 is fine. I can stream video and run Apriltags. Running ML algorithms works too.

...

@iabdalkader - Please let me know how you want to proceed. I think sending the alignment issue to Damien for fat_vfs_make_new makes sense to fix the SD card drivers. For the alif driver, we'd raise this with them?

This PR should be mergeable without waiting on the above. Adding breakpoints on cache cleaning can happen after the above two issues are fixed.

iabdalkader:
* Not sure about Alif DCD, perhaps if we add some padding it would be fine? Yes, please ask them.
* For SD Card, if we align this in `micropython/lib/oofatfs/ff.h`, does it fix it? The fix should be sent to micropython.
```C
    BYTE    win[FF_MAX_SS]; /* Disk access window for Directory, FAT (and file data at tiny cfg) */
} FATFS;
```

Note, I didn't mean to leave the break point in there forever, it was just for debugging. Some unaligned invalidate might be okay if you have enough padding/don't care about the previous line. If we're confident at some point that nothing else invalidates unaligned addresses we can leave it.

kwagyeman:
For the alif dcd issue, just changing the struct to:

`static uint32_t _xfer_trb[8][8] CFG_TUSB_MEM_SECTION __attribute__((aligned(32))); // [TODO] runtime alloc`

Resolves the issue. I emailed Alif to see if this is safe.

For the SD card block size issue, aligning the structure will be challenging because it uses mp_obj_malloc, which is unaligned. They only do cache clean+invalidate for reading blocks and clean for writing blocks. Since there are many structure values before the window buffer in the array, there's no risk of corruption of anything before `win`, but values behind it may be cleaned/invalidated. So... if a malloc were to happen of an image buffer right after the SD card object... corruption will ensue.

32 bytes of padding after `win` will resolve the problem without fixing the alignment issue. However, there's a second issue. The code doesn't do an invalidate after the SD card driver finishes filling the buffer. So, speculative reads by the CPU of the memory address aren't flushed.

I'll create an issue for Damien on this problem and explain it in detail. https://github.com/micropython/micropython/issues/19010

...

Is there anything else you want for this PR? The locations I updated in it were everywhere draw_image is used to write to an image using the GPU. These are known breaking areas.

iabdalkader:
> Is there anything else you want for this PR?

It might be worth waiting for the DCD fix too while we're at it.

kwagyeman:
Okay, hopefully alif gets back to us soon.

kwagyeman:
Done. Also, sent the fix to upstream: https://github.com/micropython/micropython/pull/19017.

iabdalkader (approved):

PR #3061, opened 2026-04-05
modules: Update all py_image methods to mp_arg_parse_all.

### LINKED ISSUE #1862: We need to refactor our argument handling to support MicroPython's argument parser.
We need to refactor our argument handling to support MicroPython's argument parser.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/1862

Refactored using AI, ran sanity checks on everything multiple times. Tested updated functions and example scripts against them.


iabdalkader:
It saved about ~2KBs, sadly wasn't done earlier, could have avoided disabling some features.

modules/py_image.c
```diff
@@ -1315,437 +1337,485 @@ static MP_DEFINE_CONST_FUN_OBJ_1(py_image_flush_obj, py_image_flush);
...
+    PY_ASSERT_TRUE_MSG((0.0f <= floating_threshold) && (floating_threshold <= 1.0f),
                        "Error: 0.0 <= floating_threshold <= 1.0!");
-    int arg_c =
-        py_helper_keyword_color(arg_img, n_args, args, offset + 2, kw_args, -1); // White.
-    bool arg_invert =
-        py_helper_keyword_float(n_args, args, offset + 3, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_invert), false);
-    bool clear_background =
-        py_helper_keyword_float(n_args, args, offset + 4, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_clear_background), false);
-    image_t *arg_msk =
-        py_helper_keyword_to_image(n_args, args, offset + 5, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_mask), NULL);
```
iabdalkader, line 1743: How were these floats?

modules/py_image.c
```diff
@@ -5914,21 +6160,27 @@ static MP_DEFINE_CONST_FUN_OBJ_KW(py_image_find_template_obj, 3, py_image_find_t
...
-    cascade->threshold = py_helper_keyword_float(n_args, args, 2, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_threshold), 0.5f);
-    cascade->scale_factor = py_helper_keyword_float(n_args, args, 3, kw_args, MP_OBJ_NEW_QSTR(MP_QSTR_scale_factor), 1.5f);
-
-    rectangle_t roi;
-    py_helper_keyword_rectangle_roi(arg_img, n_args, args, 4, kw_args, &roi);
+static mp_obj_t py_image_find_features(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    enum { ARG_threshold, ARG_scale_factor, ARG_roi };
+    static const mp_arg_t allowed_args[] = {
+        { MP_QSTR_threshold,    MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_scale_factor, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
```
iabdalkader:
```suggestion
        { MP_QSTR_scale, MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
```
No need to change this, it's also documented as `scale`. You can also rename the C variable, and member to just `scale`.

iabdalkader (changes requested):

modules/py_image.c
kwagyeman: Yeah...

modules/py_image.c
kwagyeman: Done

modules/py_image.c
```diff
@@ -1315,437 +1337,485 @@ static MP_DEFINE_CONST_FUN_OBJ_1(py_image_flush_obj, py_image_flush);
...
-
-    imlib_draw_string(arg_img, arg_x_off, arg_y_off, arg_str,
-                      arg_c, arg_scale, arg_x_spacing, arg_y_spacing, arg_mono_space,
-                      arg_char_rotation, arg_char_hmirror, arg_char_vflip,
-                      arg_string_rotation, arg_string_hmirror, arg_string_vflip);
-    return args[0];
+    uint offset = py_helper_consume_array(n_args, pos_args, 1, 3, &arg_vec);
+    int x_off = mp_obj_get_int(arg_vec[0]);
+    int y_off = mp_obj_get_int(arg_vec[1]);
+    const char *str = mp_obj_str_get_str(arg_vec[2]);
```
iabdalkader: I don't think the intention is to pass (x, y, "text") as a tuple, at least that's not what's documented.

I think we should drop this hacky code and just accept plain args. I scanned scripts and don't actually see anything making use of this pass tuple option.

modules/py_image.c
```diff
@@ -1315,437 +1337,485 @@ static MP_DEFINE_CONST_FUN_OBJ_1(py_image_flush_obj, py_image_flush);
...
-    x1 = mp_obj_get_int(p1[0]);
-    y1 = mp_obj_get_int(p1[1]);
-    x2 = mp_obj_get_int(p2[0]);
-    y2 = mp_obj_get_int(p2[1]);
-    x3 = mp_obj_get_int(p3[0]);
-    y3 = mp_obj_get_int(p3[1]);
+    int x0 = mp_obj_get_int(p0[0]), y0 = mp_obj_get_int(p0[1]);
+    int x1 = mp_obj_get_int(p1[0]), y1 = mp_obj_get_int(p1[1]);
+    int x2 = mp_obj_get_int(p2[0]), y2 = mp_obj_get_int(p2[1]);
+    int x3 = mp_obj_get_int(p3[0]), y3 = mp_obj_get_int(p3[1]);
```
iabdalkader: Bad style, previous was better.

modules/py_image.c
```diff
@@ -1315,437 +1337,485 @@ static MP_DEFINE_CONST_FUN_OBJ_1(py_image_flush_obj, py_image_flush);
...
+    imlib_draw_line(image, x2, y2, x3, y3, color, args[ARG_thickness].u_int);
+    imlib_draw_line(image, x3, y3, x0, y0, color, args[ARG_thickness].u_int);
 
-    if (arg_s >= 1) {
-        imlib_draw_circle(arg_img, x0, y0, arg_s, arg_c, arg_thickness, arg_fill);
-        imlib_draw_circle(arg_img, x1, y1, arg_s, arg_c, arg_thickness, arg_fill);
-        imlib_draw_circle(arg_img, x2, y2, arg_s, arg_c, arg_thickness, arg_fill);
-        imlib_draw_circle(arg_img, x3, y3, arg_s, arg_c, arg_thickness, arg_fill);
+    if (s >= 1) {
+        imlib_draw_circle(image, x0, y0, s, color, args[ARG_thickness].u_int, args[ARG_fill].u_bool);
```
iabdalkader:
```suggestion
        imlib_draw_circle(image, x0, y0, s, color, thickness, fill);
```
One or two times is okay, but this is read 8 times.

iabdalkader (changes requested):

kwagyeman:
Working on update. Will hit all the areas in the code where there are issues expressed with the AI.

modules/py_image.c
kwagyeman: Done

modules/py_image.c
kwagyeman: Done

modules/py_image.c
kwagyeman: Done

kwagyeman:
Updated and searched for other places in the refactor that needed fixing per the comments.

modules/py_image.c
iabdalkader: What I mean is, we could drop code like this and just x, y to args (Not accepts tuples anymore):
```C
    uint offset = py_helper_consume_array(n_args, pos_args, 1, 2, &arg_vec);
    int arg_x = mp_obj_get_int(arg_vec[0]);
    int arg_y = mp_obj_get_int(arg_vec[1]);
```

kwagyeman: The draw functions, like draw_rect, use this quite a bit to draw a rectangle object, which is passed in.

Dropping it on draw_string was fine. But, we should keep it for things like draw_line and draw_rect.

For example:

```
img.draw_rectangle(blob.rect(), color=127)
```

Becomes

```
img.draw_rectangle(blob.rect().x, blob.rect().y, blob.rect().w, blob.rect().h, color=127)
```

Had AI review this with our examples:

## `py_helper_consume_array` usage in examples

### Keep tuple support

- **`draw_line`** — **14 tuple / 18 separate** — tuple form used with `.line()`, `.major_axis_line()`, `.minor_axis_line()` return values across line detection and blob tracking examples, plus tuple/list literals in optical flow and unit tests
- **`draw_rectangle`** — **79 tuple / 9 separate** — overwhelmingly tuple-based; the vast majority of calls pass `.rect()`, `.rect` properties, or detection result objects directly
- **`draw_circle`** — **1 tuple / 7 separate** — only a single tuple call in `tf_object_detection.py`; most examples use separate `c.x(), c.y(), c.r()`
- **`draw_edges`** — **4 tuple / 0 separate** — always called with `blob.min_corners()`; no separate-arg form exists in any example

### Safe to remove tuple support

- **`get_pixel`** / **`set_pixel`** — always called with separate `x, y`
- **`draw_ellipse`** — no examples found
- **`draw_string`** — already removed
- **`draw_cross`** — always separate `x, y`
- **`draw_arrow`** — always separate `x0, y0, x1, y1`
- **`mask_rectangle`** / **`mask_circle`** / **`mask_ellipse`** — no examples found
- **`flood_fill`** — always separate `x, y`

...

So, should I remove support for this except for `draw_line`, `draw_rectangle`, `draw_circle`, and `draw_edges`?

iabdalkader: Yes, I think it's only useful when you have a list of shapes to draw.

scripts/examples/07-Interface-Library/01-Pixy-Emulation/apriltags_pixy_uart_emulation.py
```diff
@@ -225,7 +225,7 @@ def parse_byte(byte):
             if id_map[tag.id] <= max_blocks_per_id:
                 dat_buf += to_object_block_format(tag)
                 img.draw_rectangle(tag.rect)
-                img.draw_cross(tag.cx, tag.cy)
+                img.draw_cross(tag.centroid)
```
iabdalkader, line 228: Is this a property or function? I've seen it used as both. Same for cx/cy.

modules/py_image.c
```diff
@@ -262,6 +262,14 @@ mp_obj_t py_kptmatch_match(mp_obj_t self_in) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_1(py_kptmatch_match_obj, py_kptmatch_match);
 
+mp_obj_t py_kptmatch_centroid(mp_obj_t self_in) {
+    return mp_obj_new_tuple(2, (mp_obj_t []) {
+        ((py_kptmatch_obj_t *) self_in)->cx,
+        ((py_kptmatch_obj_t *) self_in)->cy
+    });
+}
+static MP_DEFINE_CONST_FUN_OBJ_1(py_kptmatch_centroid_obj, py_kptmatch_centroid);
```
iabdalkader, line 271: Is it too bad to just add `()` around the existing calls?

scripts/examples/07-Interface-Library/01-Pixy-Emulation/apriltags_pixy_uart_emulation.py
kwagyeman: For apriltags it is a property; the rest are of the objects like blob and etc. are still functions. Since v5.0.0 is breaking anyway, we might want to move them to properties too.

I'm thinking to do that in another PR after this one is merged? I can also address https://github.com/openmv/openmv/issues/1449 at the same time too.

modules/py_image.c
kwagyeman: It makes sense to add centroid support; it's clearer and makes sense with trying to pass tuples around as much as possible.

iabdalkader: I don't see how this saves memory? You create a new tuple anyway.

scripts/examples/07-Interface-Library/01-Pixy-Emulation/apriltags_pixy_uart_emulation.py
iabdalkader: Okay

modules/py_image.c
kwagyeman: Yeah, it doesn't. I updated my comment. I think it's better than passing x/y separately.

iabdalkader: > it's clearer and makes sense with trying to pass tuples around as much as possible.

Except it's not really clear what it is: `img.draw_cross((tag.cx, tag.cy))` much better.

kwagyeman: Okay, will remove centroids and keep as cx/cy.

iabdalkader (approved):

PR #3081, opened 2026-04-08
scripts/libraries: Remove forced scale to fit in the ML library.

kwagyeman:
Image processing and NMS inside the ML library no longer scale to expand.

Instead, it just stretches. This works well for all YOLO like models and keypoint regressors.

Examples now directly control the w/h of the image passed into the model without any effects from the ML library.

For models like BlazeFace/Palm, these work better with a square crop.

https://github.com/user-attachments/assets/ea209486-b0df-461a-a778-9138db65b166


iabdalkader (approved):

iabdalkader:
This PR broken CI unit tests, it broke `tensorflow.py` (runs on MPS2) because you missed updating it. CI didn't detect it because it doesn't trigger on scripts.

PR #3083, opened 2026-04-09
ports/stm32: Move display buffer into private SDRAM bank.

kwagyeman:
For the longest the OpenMV PT has suffered from display stability issues were the screen would loose lock and flicker randomly. On debugging the issue I determined it was due to a FIFO underflow by the display controller which generated interrupts saying this.

The OpenMV PT becomes unstable due to the additional memory load caused by rendering the thermal color image from the FLIR Lepton on the regular image. Removing this rendering step makes the system stable... but, then there's no thermal overlay.

On further debugging, it was discovered the the display stability could be achieved by pushing the screen buffer allocation into its own SDRAM bank. However, with fballoc, this wasn't fixable in any pretty way.

Now thanks to umalloc, it's possible to create a private data pool for just the display controller so that it can receive the highest possible SDRAM bandwidth by having it's own private bank.

...

The Arduino Giga doesn't show any display stability issues compared to the OpenMV PT. Since it's not dealing with dual cameras and thermal overlay, the memory load is lower.

### DIFF
--- a/boards/OPENMVPT/board_config.h
+++ b/boards/OPENMVPT/board_config.h
@@ -117,11 +117,14 @@
 #define OMV_GC_BLOCK1_MEMORY                    DRAM    // Extra GC block 0.
 #define OMV_GC_BLOCK1_SIZE                      (8M)
 #define OMV_UMA_BLOCK0_MEMORY                   DRAM    // Default UMA pool.
-#define OMV_UMA_BLOCK0_SIZE                     (55M)
+#define OMV_UMA_BLOCK0_SIZE                     (39M)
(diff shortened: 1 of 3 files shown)


kwagyeman:
Closing for further investigation. I had previously managed to fix this issue by pushing the frame buffer allocs into their own bank via fb_allocing 16MB... However, now giving them a private 16MB bank does not appear to resolve the issue. Glitching is reduced heavily, but not 100% gone. Still occurs every minute or so. It may not be fixable on the PT, given all the stuff it's doing.

Current firmware behavior, while not perfect, is acceptable. Will revisit on a complaint given product is EOL. Probably needs more than just moving SDRAM banks.

PR #3089, opened 2026-04-09
framebuffer: Track FPS in stream frame header.

iabdalkader:
Add EMA-based frame time tracking to framebuffer_update_preview() and embed the computed FPS in the stream frame header, so the IDE can read it without requiring script-level clock boilerplate.

### DIFF
--- a/lib/imlib/framebuffer.c
+++ b/lib/imlib/framebuffer.c
@@ -263,6 +263,20 @@ void framebuffer_update_preview(image_t *src) {
     static int overflow_count = 0;
     framebuffer_t *fb = framebuffer_get(FB_STREAM_ID);
 
+    // Update FPS tracking (EMA of frame time in ms).
+    uint32_t now = mp_hal_ticks_ms();
+    if (fb->fps_last_ms) {
+        uint32_t delta = now - fb->fps_last_ms;
+        if (delta > 0) {
+            if (fb->fps_frame_time > 0.0f) {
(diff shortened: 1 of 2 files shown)


lib/imlib/framebuffer.c
```diff
@@ -263,6 +263,20 @@ void framebuffer_update_preview(image_t *src) {
     static int overflow_count = 0;
     framebuffer_t *fb = framebuffer_get(FB_STREAM_ID);
 
+    // Update FPS tracking (EMA of frame time in ms).
+    uint32_t now = mp_hal_ticks_ms();
+    if (fb->fps_last_ms) {
+        uint32_t delta = now - fb->fps_last_ms;
+        if (delta > 0) {
```
kwagyeman: Backwards time check 🙃

iabdalkader: What's backwards?

kwagyeman: No, you are checking if time went backward. That error you saw with MP on Alif. I like the check.

iabdalkader: Ah, no it's actually so that `delta !=0` it can't be less than 0 anyway, it's unsigned. The MP time issue is off by 1 or 2ms, this function can't be called that fast. That said, this test is probably useless for the same reason.

iabdalkader: Correction: off by 1ms* with `ticks_us`

PR #3091, opened 2026-04-10
protocol: Skip MicroPython CDC line state callback during protocol mode.

iabdalkader:
Prevents SOF interrupt flooding when DTR is set while the protocol channel is active. MicroPython enables SOF at 125us intervals on high-speed USB, which overflows the TinyUSB event queue when the main loop is blocked in long-running operations like apriltag.


iabdalkader:
FYI more protocol stability fixes @kwagyeman

PR #3097, opened 2026-04-11
nema: Poll events in nema_wait_irq_cl.

iabdalkader:
nema_wait_irq_cl can block for a while causing the connection to be dropped. Use mp_event_wait_ms to service USB and other events while waiting, and return an error on timeout so the caller can fall back to software rendering.

### DIFF
--- a/ports/stm32/omv_gpu.c
+++ b/ports/stm32/omv_gpu.c
@@ -184,7 +184,7 @@ int omv_gpu_draw_image(image_t *src_img,
     HAL_ICACHE_WaitForInvalidateComplete();
 
     nema_cl_submit(&cl);
-    nema_cl_wait(&cl);
+    int ret = nema_cl_wait(&cl);
 
     #if !OMV_GPU_NEMA_MM_STATIC
     nema_cl_destroy(&cl);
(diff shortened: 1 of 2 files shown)


iabdalkader:
FYI @kwagyeman

kwagyeman:
This commit corrupts the N6 memory. I don't think the GPU stops executing even if you timeout waiting on it.

```
import sensor
import time
import math

sensor.reset()
sensor.skip_frames(time=2000)
clock = time.clock()

while True:
    clock.tick()
    img = sensor.snapshot()
    for tag in img.find_apriltags():
        img.draw_rectangle(tag.rect, color=(255, 0, 0))
        img.draw_cross(tag.cx, tag.cy, color=(0, 255, 0))
        print_args = (tag.name, tag.id, (180 * tag.rotation) / math.pi)
        print("Tag Family %s, Tag ID %d, rotation %f (degrees)" % print_args)
    print(clock.fps())
```

Run this script with RAW buffer streaming, and you'll notice the FPS curiously getting slower over time until the system crashes. You may also get errors about not being able to allocate persistent memory.

...

Everything is fine in https://github.com/openmv/openmv/commit/2ee26b5140f8c2558fee5e1db715df19a32161be right before this commit.

iabdalkader:
It could be one of two things, or both: the small stack threshold you've set that I've warned you about, or the reduced libc heap (used to be 128K).

PR #3100, opened 2026-04-12
protocol: Run tud_task from PendSV dispatch.

iabdalkader:
This PR runs TinyUSB polling from PendSV dispatch scheduled right from USB IRQ (tud_event_cb). PendSV mostly runs right after USB IRQ returns, which makes USB respond almost instantly. The dispatch callback also schedules the protocol to run in VM context where it's safe to raise.

There's still room for improvement here, we could run the protocol polling from the dispatch callback directly, so the flow would be: USB IRQ -> dispatch -> PendSV -> tud_task + protocol_task. However, this is more tricky as we'll need to make sure nothing raises exceptions and we'll need to gate PendSV around critical sections (like gc_ functions).

### DIFF
--- a/protocol/omv_protocol_channel_tinyusb.c
+++ b/protocol/omv_protocol_channel_tinyusb.c
@@ -24,9 +24,7 @@
  * OpenMV Protocol USB Channel.
  */
 #if OMV_USB_STACK_TINYUSB
-#include "py/stream.h"
 #include "py/mphal.h"
-#include "py/ringbuf.h"
 #include "py/runtime.h"
 
 #include "omv_common.h"
@@ -34,25 +32,30 @@
 #include "omv_protocol.h"
 #include "board_config.h"
 #include "tusb.h"
+#include "device/dcd.h"
+#include "pendsv.h"
 
 #ifndef OMV_PROTOCOL_USB_CHANNEL_TIMEOUT_MS
 #define OMV_PROTOCOL_USB_CHANNEL_TIMEOUT_MS (1500)
 #endif
 
-static bool usb_channel_active;
+static volatile bool usb_channel_active;
+static volatile bool usb_channel_masked;
+static volatile bool usb_channel_pending;
+static mp_sched_node_t usb_protocol_node;
+
+static void usb_channel_suspend(const omv_protocol_channel_t *channel);
+static void usb_channel_resume(const omv_protocol_channel_t *channel);
 
 static size_t usb_channel_size(const omv_protocol_channel_t *channel) {
-    if (tud_task_event_ready()) {
-        tud_task_ext(0, false);
-    }
     return tud_cdc_available();
 }
 
 static int usb_channel_flush(const omv_protocol_channel_t *channel) {
-    if (tud_task_event_ready()) {
-        tud_task_ext(0, false);
-    }
-    return tud_cdc_write_flush();
+    usb_channel_suspend(channel);
+    int ret = tud_cdc_write_flush();
+    usb_channel_resume(channel);
+    return ret;
 }
 
 static bool usb_channel_is_active(const omv_protocol_channel_t *channel) {
@@ -64,12 +67,6 @@ static int usb_channel_read(const omv_protocol_channel_t *channel, uint32_t offs
     uint32_t start_ms = mp_hal_ticks_ms();
     while (bytes < size && !check_timeout_ms(start_ms, OMV_PROTOCOL_USB_CHANNEL_TIMEOUT_MS)) {
         bytes += tud_cdc_read((uint8_t *) data + bytes, size - bytes);
-        if (tud_task_event_ready()) {
-            tud_task_ext(0, false);
-        }
-        if (bytes < size) {
-            mp_event_handle_nowait();
-        }
     }
 
     return bytes;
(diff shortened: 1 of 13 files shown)


protocol/omv_protocol_channel_tinyusb.c
```diff
@@ -34,12 +34,35 @@
...
+    soft_timer_static_init(&usb_channel_timer, SOFT_TIMER_MODE_PERIODIC,
+                           OMV_PROTOCOL_USB_CHANNEL_POLL_MS, usb_channel_timer_cb);
+    return 0;
+}
+
+static int usb_channel_deinit(const omv_protocol_channel_t *channel) {
+    usb_channel_active = false;
+    soft_timer_remove(&usb_channel_timer);
+    return 0;
+}
```
kwagyeman:
```suggestion
static int usb_channel_init(const omv_protocol_channel_t *channel) {
    soft_timer_static_init(&usb_channel_timer, SOFT_TIMER_MODE_PERIODIC,
                           OMV_PROTOCOL_USB_CHANNEL_POLL_MS, usb_channel_timer_cb);
    return 0;
}

static int usb_channel_deinit(const omv_protocol_channel_t *channel) {
    soft_timer_remove(&usb_channel_timer);
    return 0;
}
```

The previous behavior left `usb_channel_active` unchanged during a soft reset. This allows you to keep the port open and resync to get back into communication with the camera after a soft reset. Forcing this to false means that the port has to be closed and reopened after a soft reset.

Do we want to change this behavior? It breaks the current IDE on stop. I'm not seeing any issues leaving `usb_channel_active` unchanged.

protocol/omv_protocol_channel_tinyusb.c
```diff
@@ -92,12 +115,18 @@ static int usb_channel_write(const omv_protocol_channel_t *channel, uint32_t off
 }
 
 static void usb_channel_task(mp_sched_node_t *node) {
-    tud_task_ext(0, false);
+    tud_task_ext(0, true);
```
kwagyeman:
```suggestion
    tud_task_ext(0, node == NULL);
```

This way if `tud_task_ext` ever is updated to look at that flag its automatically correct for when `usb_channel_task` is called by `usb_channel_timer_cb` and `mp_sched_schedule_node`.

protocol/omv_protocol_channel_tinyusb.c
```diff
@@ -34,12 +34,35 @@
...
+#include "device/dcd.h"
+#include "shared/runtime/softtimer.h"
 
 #ifndef OMV_PROTOCOL_USB_CHANNEL_TIMEOUT_MS
 #define OMV_PROTOCOL_USB_CHANNEL_TIMEOUT_MS (1500)
 #endif
 
-static bool usb_channel_active;
+#ifndef OMV_PROTOCOL_USB_CHANNEL_POLL_MS
+#define OMV_PROTOCOL_USB_CHANNEL_POLL_MS (50)
```
kwagyeman:
```suggestion
#define OMV_PROTOCOL_USB_CHANNEL_POLL_MS (25)
```

I'm noticing some sluggishness in the fps to the IDE in some algorithms like Apriltags at VGA res on the AE3. Increasing this makes a noticeable difference when the camera is running image library algorithms that don't poll currently at high reses.

kwagyeman:
After testing on the AE3, RT1062, and N6, it's working well. This covers those long-running algorithms generally.

This same soft timer will also help stmusb. I applied the same implementation as how it was done with tinyusb and it makes a noticeable difference in IDE FPS. Can we implement it there, too?

protocol/omv_protocol_channel_tinyusb.c
iabdalkader: Doesn't matter, wrong ISR anyway, I think it's for calling it from USB IRQ.

protocol/omv_protocol_channel_tinyusb.c
iabdalkader: Let's retest before this.

kwagyeman:
Feels great! Every algorithm is running fast! Across the board!

kwagyeman:
Perfect! Merge it!

protocol/omv_protocol_channel_tinyusb.c
```diff
@@ -80,62 +77,84 @@ static int usb_channel_write(const omv_protocol_channel_t *channel, uint32_t off
...
-    if (!omv_protocol_is_active()) {
+    if (!usb_channel_active) {
         __mp_tud_cdc_rx_cb(itf);
     }
 }
 
-void tud_cdc_line_coding_cb(uint8_t itf, cdc_line_coding_t const *coding) {
-    usb_channel_active = (coding->bit_rate == OMV_PROTOCOL_MAGIC_BAUDRATE);
-}
-
```
kwagyeman, line 113: Please leave this in. It appears necessary with some serial port libraries to set usb_channel_active. tud_cdc_line_state_cb is not good enough. Seems to be required for the Mac and Linux Qt drivers when they open the port. Windows is unaffected.

iabdalkader:
I realized this is broken and it's never been safe, at least as it currently stands. Any code can be preempted by PendSV (tud_task -> msc_write) while reading/writing flash. It could be made safe but not without upstream changes.

Closing this, but I did manage to find and fix two related issues:

https://github.com/openmv/openmv/pull/3109
https://github.com/openmv/openmv/pull/3110
https://github.com/openmv/openmv/pull/3111

PR #3104, opened 2026-04-13
modules: Add raw_bpp control from py_csi_ng.

### LINKED ISSUE #3103: Examine if using 1bpp is possible with new CSI FB Arch.
Folks can't use VGA grayscale resolution on the F7/H7 anymore. This is because the frame buffer takes 2BPP always.

It may be possible to pass a flag to initialize the camera CSI object with only 1BPP. However, all paths that depend on the 2BPP behavior must be checked to ensure that they can handle this.

This will enable older cameras to support higher resolutions again.

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/3103

Restores previous larger resolution grayscale image processing on older boards like the OpenMV M4, M7, H7 and Arduino Nicla without SDRAM. You can now do:

```
import csi
import time

csi0 = csi.CSI(raw_bpp=1)
csi0.reset()
csi0.pixformat(csi.GRAYSCALE) # or BAYER
csi0.framesize(csi.VGA)
csi0.snapshot(time=2000)

clock = time.clock()

while True:
    clock.tick()
    img = csi0.snapshot()
    print(clock.fps())
```

To set the frame buffer to support GRAYSCALE/BAYER images only.

### DIFF
--- a/common/omv_csi.c
+++ b/common/omv_csi.c
@@ -1340,10 +1340,12 @@ __weak int omv_csi_set_framebuffers(omv_csi_t *csi, size_t count) {
     // TODO pass this to resize.
     #if OMV_CSI_HW_CROP_ENABLE
     // If hardware cropping is supported, use window size.
-    size_t frame_size = csi->fb->u * csi->fb->v * 2;
+    size_t frame_size = csi->fb->u * csi->fb->v * csi->fb->raw_bpp;
     #else
     // Otherwise, use the real frame size.
-    size_t frame_size = csi->resolution[csi->framesize][0] * csi->resolution[csi->framesize][1] * 2;
+    size_t frame_size = csi->resolution[csi->framesize][0] *
+                        csi->resolution[csi->framesize][1] *
+                        csi->fb->raw_bpp;
     #endif
 
     if (count == -1) {
--- a/lib/imlib/framebuffer.c
+++ b/lib/imlib/framebuffer.c
@@ -44,18 +44,18 @@ void framebuffer_init0() {
     bool enabled = framebuffer_get(FB_STREAM_ID)->enabled;
 
     // Initialize the main framebuffer.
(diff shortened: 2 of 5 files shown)


iabdalkader:

https://forums.openmv.io/t/frame-buffer-overflow-how-to-adjust-the-frame-buffer-without-changing-the-code/11593/4

iabdalkader:
Not very happy with this one, everything assumes 2bpp you're probably missing some corner case here or there, also just use `window()`? Or increase FB if possible? If all fail, buy a new camera, they're cheap.

kwagyeman:
Just trying to restore previous functionality per a second customer's request on the forums. I covered all cases where framebuffer_resize() is used and made that bpp configurable. Other cases with buffer resizing all check the buffer size now and fail if this were set. So turning this on makes a lot of things just not work, like `.to_rgb565()` will throw an error about there not being enough RAM when this is on - as expected.

> Or increase FB if possible?

Doesn't really help for the M7/H7, not enough RAM to store VGA with 2 bpp. Only windowing would work. But, then, this doesn't get you back to the previous capable VGA.

> If all fail, buy a new camera; they're cheap.

That's the other fix... but, kinda lame for the customer.

iabdalkader:
> Just trying to restore previous functionality per a second customer's request on the forums.

Understood, still can't do much about this. It's been like this for a very long time, we moved to 2BPP because it made everything much simpler and some uses cases require it (ex bayer -> rgb565 in alif csi).

> Only windowing would work

Not a bad option, 480x480 should work.

kwagyeman:
To anyone seeing this PR, ping me if you need a custom firmware with a 1-byte-per-pixel frame buffer. Otherwise. This is a won't fix as it's not a problem on new cameras. Just an issue for really old ones with no RAM. To avoid complicating the firmware to support older systems, this is a won't fix.

PR #3113, opened 2026-04-15
boards/OPENMV_N6: Add dedicated standby RAM region.

kwagyeman:
Requires: https://github.com/openmv/micropython/pull/141

1. Creates a new standby region that `_boot_mem` can be assigned to.
2. Carves 8KB of RAM out of the first 80KB  (which are retained during standby) of AXISRAM1 for the standby ram region.
3. Moves the reset RAM function out of the ITCM, freeing up space for the stack to move there, and allowing the UMA fast SRAM pool to be increased by 66KB.

Change 3 is not required for the deep sleep fix. But it seems like an obvious thing to do now.

Deep sleep with gpio wakeup and RTC work as expected.

### DIFF
--- a/boards/OPENMV_N6/board_config.h
+++ b/boards/OPENMV_N6/board_config.h
@@ -144,6 +144,7 @@
 // Linker script constants (see common.ld.S).
 #define OMV_MAIN_MEMORY                     SRAM1  // Data/BSS memory
 #define OMV_STACK_MEMORY                    SRAM1  // stack memory
+#define OMV_STANDBY_MEMORY                  SSRAM
 #define OMV_RAMFUNC_MEMORY                  ITCM
 #define OMV_STACK_SIZE                      (64K)
 #define OMV_HEAP_MEMORY                     SRAM1  // libc/sbrk heap memory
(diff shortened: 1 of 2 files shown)


boards/OPENMV_N6/board_config.h
```diff
@@ -143,8 +143,9 @@
 
 // Linker script constants (see common.ld.S).
 #define OMV_MAIN_MEMORY                     SRAM1  // Data/BSS memory
-#define OMV_STACK_MEMORY                    SRAM1  // stack memory
-#define OMV_RAMFUNC_MEMORY                  ITCM
+#define OMV_STACK_MEMORY                    ITCM   // stack memory
+#define OMV_STANDBY_MEMORY                  SSRAM
+#define OMV_RAMFUNC_MEMORY                  SSRAM
```
iabdalkader: This should remain in ITCM as it was.

iabdalkader (changes requested):

boards/OPENMV_N6/board_config.h
kwagyeman: Removed, the extra 64KB is available, though, if you want it.

iabdalkader: It's should be non-cacheable, the startup code doesn't flush cache.

iabdalkader (approved):

PR #3114, opened 2026-04-16
ports/stm32: Add VENC jpeg compressor.

kwagyeman:
Enables the VENC JPEG compressor on the N6. Easily gets 60 FPS with VGA resolution. Note that it only works for RGB565/YUV images with particular widths and heights. Seems to hit 30 fps with 1280x800. 15 FPS at 1080P.

RGB565/YUV have been tested along with different quality levels and subsampling.

https://github.com/user-attachments/assets/6faf2d45-6175-414a-865d-4c0dee3ef91d

...

PAG7936 QVGA -> Works at 120 FPS
PAG7936 VGA -> Works at 60 FPS
PAG7936 HD -> Works at 25 FPS

PS5520 VGA -> Works at 30 FPS
PS5520 HD -> Works at 30 FPS
PS5520 FHD -> Works at 15 FPS
PS5520 WQXGA2 -> Works at 5 FPS


iabdalkader:

https://forums.openmv.io/t/plans-for-n6-support-h264-encoder/11412/10

kwagyeman:
PR is ready now. ST support got back to me and gave me a fix for the crashing issues... which they use in their cube examples, but is not in the errata sheet. The hardware handshake feature between the DCMIPP and VENC triggers supriously even when not enabled, which can cause the VENC to generate a hardware error and need to be reset.

iabdalkader:
I can't review this for 5.0.0, unless you don't mind waiting another month or so. IMHO py_image + PS fixes + upstream USB fixes should be the last for 5.0.0.

EDIT: And this: https://github.com/openmv/openmv/pull/3061#discussion_r3119178766

Because, like you said the API is changing, so best get this done now.

drivers/vc8000/include/enccommon.h
```diff
@@ -0,0 +1,285 @@
+/*
```
iabdalkader, line 1: Where does this library come from? It might make more sense to add it as a submodule, it's massive.

boards/OPENMV_N6/board_config.h
```diff
@@ -34,6 +34,10 @@
 #define OMV_GPU_NEMA                        (1)
 #define OMV_GPU_NEMA_BUFFER_SIZE            (32 * 1024)
 
+// VENC Configuration
+#define OMV_VENC_ENABLE                     (1)
+#define OMV_VENC_VC8000_JPEG_ENABLE         (1)
```
iabdalkader, line 39:
```suggestion
#define OMV_VENC_CODEC_ENABLE                     (1)
```

For consistency? Also, why two flags?

boards/OPENMV_N6/board_config.mk
```diff
@@ -24,12 +24,12 @@ OMV_PROG_STLDR=MX25UM51245G_STM32N6570-NUCLEO.stldr
 OMV_JLINK_ARGS=--device STM32N657L0
 OMV_STLINK_ARGS=--extload $(OMV_PROG_STLDR)
 OMV_HSE_VALUE=48000000
-CPU_FREQ_HZ=800000000
 OMV_ENABLE_BL=1
 OMV_BOSON_ENABLE=1
 OMV_GENX320_ENABLE=1
 OMV_LEPTON_SDK_ENABLE=1
 OMV_USB_STACK_TINYUSB=1
+OMV_VENC_VC8000_ENABLE=1
```
iabdalkader, line 32:
```suggestion
OMV_VC8000_ENABLE=1
```

iabdalkader (changes requested):

drivers/vc8000/include/basetype.h
```diff
@@ -0,0 +1,129 @@
+/*
```
iabdalkader, line 1: Remove 5MBs pdf file.

ports/stm32/stm_vc8000.c
```diff
@@ -0,0 +1,297 @@
...
+    // could set it and have the store silently clobber it.
+    for (mp_uint_t tickstart = mp_hal_ticks_ms();
+         !atomic_exchange_explicit(&venc_irq_fired, 0, memory_order_acq_rel); ) {
+
+        mp_uint_t elapsed = mp_hal_ticks_ms() - tickstart;
+        if (elapsed > EWL_TIMEOUT) {
+            return EWL_HW_WAIT_TIMEOUT;
+        }
+
+        mp_event_wait_ms(EWL_TIMEOUT - elapsed);
```
iabdalkader, line 266: This could raise, you probably want to call:
```suggestion
        imlib_poll_events_noexec();
```

ports/stm32/stm_vc8000.c
```diff
@@ -0,0 +1,297 @@
...
+
+    if (slicesReady != NULL) {
+        *slicesReady = (VENC_REG(21UL) >> 16) & 0xFFUL;
+    }
+
+    return EWL_OK;
+}
+
+void VENC_IRQHandler(void) {
+    IRQ_ENTER(VENC_IRQn);
```
iabdalkader, line 277:
```suggestion
```
Why?

ports/stm32/stm_vc8000.c
```diff
@@ -0,0 +1,297 @@
...
+    (void) inst;
+    return EWL_OK;
+}
+
+void EWLReleaseHw(const void *inst) {
+    (void) inst;
+}
+
+// ---------------------------------------------------------------------------
+// DMA-coherent linear memory allocation
```
iabdalkader, line 175: Note the memory below is cached.

drivers/vc8000/include/enccommon.h
kwagyeman: Right here: https://github.com/STMicroelectronics/STM32CubeN6/tree/main/Middlewares/Third_Party/VideoEncoder

One of the few libraries they just put on github without hiding the source.

boards/OPENMV_N6/board_config.h
kwagyeman: I was thinking it would eventually end up as:

```
#define OMV_VENC_ENABLE                     (1)
#define OMV_VENC_VC8000_JPEG_ENABLE         (1)
#define OMV_VENC_VC8000_H264_ENABLE         (1)        
```

Was trying to follow how the GPU is enabled:

```
#define OMV_GPU_ENABLE                      (1)
#define OMV_GPU_NEMA                        (1)
```

drivers/vc8000/include/enccommon.h
iabdalkader: Should we add in a new repo and add as a submodule?

kwagyeman:
> I can't review this for 5.0.0, unless you don't mind waiting another month or so. IMHO py_image + PS fixes + upstream USB fixes should be the last for 5.0.0.

Yeah, given you're heading out soon, probably makes sense to cut the release before you leave. So, just bug fixes and API updates. I'll remove feature enhancements from the v5.0.0 tracker.

drivers/vc8000/include/enccommon.h
kwagyeman: It requires some patching. I had to fix some include issues and add cache cleaning for jpeg. I expect to do more for H.264 in the future. But we could treat it the same as Apriltags and move it to a subrepo. However, I don't expect any updates, given that it controls a hardware module, so it would just be a subrepo that never gets any updates. Only patches.

Move to submodule still?

iabdalkader:

https://forums.openmv.io/t/openmv-n6-pag7936-1280x800-120-fps/11615/10

drivers/vc8000/include/enccommon.h
iabdalkader: No, let's keep it as is.

ports/stm32/stm_vc8000.c
kwagyeman: I perform cache maintenance operations as required for JPEG. Will handle them too for H.264 in the future.

iabdalkader:

https://forums.openmv.io/t/ps5520-5mp-support-request-for-pre-release-firmware-pr-3114-3132/11646/1

iabdalkader:

https://forums.openmv.io/t/maximum-n6-image-quality/11683/2

ports/stm32/port_config.mk
```diff
@@ -101,6 +101,7 @@ OMV_CFLAGS += -I$(TOP_DIR)
 OMV_CFLAGS += -I$(TOP_DIR)/modules
 OMV_CFLAGS += -I$(TOP_DIR)/$(LIBPDM_DIR)
 OMV_CFLAGS += -I$(TOP_DIR)/$(NEMA_DIR)/include
+OMV_CFLAGS += -I$(TOP_DIR)/$(VC8000_DIR)/include
```
iabdalkader: Isn't this already added in drivers.mk?

lib/imlib/jpege.c
```diff
@@ -932,7 +932,7 @@ static void jpeg_write_headers(jpeg_buf_t *jpeg_buf, int w, int h, int bpp, jpeg
 
 bool jpeg_compress(image_t *src, image_t *dst, int quality, bool realloc, jpeg_subsampling_t subsampling) {
     if (!dst->data) {
-        uint32_t size = uma_avail(0);
+        uint32_t size = IM_MIN(uma_avail(0), JPEG_MAX_ALLOC_SIZE);
```
iabdalkader, line 935: How about we limit that to raw image size, or `image_size  / 2` and use `uma_realloc`? Software compressor does that already, stm32 hardware compressor does not. Can be done in a follow-up PR.

iabdalkader (commented):
A few comments, otherwise looks good and ready to merge.

lib/imlib/jpege.c
kwagyeman: image_size, or image_size / 2 would be larger than 1MB when talking about a 2592 x 1944 image (5MP) -> 5,168,448 / 2 = 2,584,224. So, that would make the allocation way larger than it needs to be for the largest case.

If we limited all JPEG buffer max allocs to the stream buffer size, that would then make it too small on older cameras, reducing JPEG quality much farther than it needs to be when there's plenty of RAM available via umm_alloc.

I think a constant here works better. Setting to 1MB limits how much ram is reserved to a sane amount while not being artificially too small.

As for using uma_realloc, this won't work for the VENC as you give it the max amount of RAM it's allowed to write to when it is launched and then it largely does it's thing while the processor spins in a busy loop.

ports/stm32/port_config.mk
kwagyeman: Nice catch!

iabdalkader (approved):

iabdalkader:

https://forums.openmv.io/t/is-mp4-audio-video-muxing-possible-on-the-openmv-n6-firmware-v5-0-0/11819/2

PR #3117, opened 2026-04-16
ports/stm32: Fix Nema memory leaks.

iabdalkader:
Free previous ring buffer on re-init to prevent heap leak on soft-reset, and use mp_handle_pending(MP_HANDLE_PENDING_CALLBACKS_ONLY) in the GPU wait loop to avoid raising exceptions that would leak the command list.

### DIFF
--- a/ports/stm32/stm_nema.c
+++ b/ports/stm32/stm_nema.c
@@ -24,6 +24,7 @@
  * STM32 Nema HAL layer.
  */
 #include <stdlib.h>
+#include <stdio.h>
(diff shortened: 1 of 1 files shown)


kwagyeman:
@iabdalkader - Why not just move it to umalloc()? Then the buffers free themselves on a soft reset.

iabdalkader:
> Why not just move it to umalloc()? Then the buffers free themselves on a soft reset.

Because there's no way to reserve memory in a pool.

PR #3132, opened 2026-04-22
ports/stm32: Add image scaler support.

kwagyeman:
Adds image scaler support. Drastically improves the PSS520/PAG7936 image quality. Also allows for the N6 to handle any resolution with the PS5520/PAG7936. @aqsnyder 

Before:
vga-before

After:
vga-after

The image scaler enables the PS5520 to output any resolution now. So, all the extra resolution tables are eliminated except the 2592x1944 table. The image scaler can output down to 40x30. Cropping is also operational.

...

PAG7936 Before (QVGA):

image

PAG7936 After (QVGA):

image

### DIFF
--- a/common/omv_csi.h
+++ b/common/omv_csi.h
@@ -377,6 +377,12 @@ typedef struct _omv_csi {
     // Resolution table
     uint16_t resolution[OMV_CSI_FRAMESIZE_MAX][2];
 
+    #ifdef OMV_CSI_HW_SCALE_ENABLE
+    // Sensor raw resolution output (i.e. before cropping and scaling).
+    uint16_t src_w;
+    uint16_t src_h;
+    #endif // OMV_CSI_HW_SCALE_ENABLE
+
     // Sensor function pointers
     int (*reset) (omv_csi_t *csi);
     int (*sleep) (omv_csi_t *csi, int enable);
--- a/drivers/sensors/pag7936.c
+++ b/drivers/sensors/pag7936.c
@@ -181,6 +181,7 @@
 typedef struct {
     bool gain_auto;
     bool expo_auto;
+    int framerate;
 } pag7936_state_t;
 
 static pag7936_state_t pag7936_state = {};
@@ -498,6 +499,7 @@ static int reset(omv_csi_t *csi) {
     pag7936_state_t *state = csi->priv;
     state->gain_auto = true;
     state->expo_auto = true;
+    state->framerate = PAG7936_HD_FPS_MAX;
     csi->gainceiling = OMV_CSI_GAINCEILING_16X;
     // Write default registers
     for (int i = 0; default_regs[i][0] && ret == 0; i++) {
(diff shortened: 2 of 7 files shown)


iabdalkader:

https://forums.openmv.io/t/ps5520-5mp-support-request-for-pre-release-firmware-pr-3114-3132/11646/1

iabdalkader:

drivers/sensors/ps5520.c
```diff
@@ -1646,6 +846,10 @@ int ps5520_init(omv_csi_t *csi) {
     csi->set_hmirror = set_hmirror;
     csi->set_vflip = set_vflip;
 
+    if (csi->isp_set_src != NULL) {
+        csi->isp_set_src(csi, SENSOR_WIDTH, SENSOR_HEIGHT);
+    }
+
```
iabdalkader: I don't see set_framesize calling `isp_set_src`. What if set_framerate never called? Also, why does set_frame change the image size to begin with and not set_framesize?

drivers/sensors/ps5520.c
```diff
@@ -1241,38 +461,18 @@ static int set_pixformat(omv_csi_t *csi, pixformat_t pixformat) {
 }
 
 static int set_framesize(omv_csi_t *csi, omv_csi_framesize_t framesize) {
-    int ret = 0;
-    const uint8_t(*regs)[2];
+    uint32_t w = csi->resolution[framesize][0];
+    uint32_t h = csi->resolution[framesize][1];
 
-    switch (framesize) {
```
iabdalkader, line 1247: So this driver will never be used without an ISP?

drivers/sensors/pag7936.c
```diff
@@ -543,50 +545,62 @@ static int set_pixformat(omv_csi_t *csi, pixformat_t pixformat) {
...
+        framerate <= PAG7936_VGA_FPS_MAX) {
+        return OMV_CSI_FRAMESIZE_VGA;
     }
 
-    return ret;
+    return OMV_CSI_FRAMESIZE_QVGA;
 }
 
-static int set_framerate(omv_csi_t *csi, int framerate) {
+static int configure(omv_csi_t *csi, omv_csi_framesize_t target, int framerate) {
```
iabdalkader, line 573: I don't like this coupling of frame size and frame rate, the duplicate register writes and and the implicit switch. Perhaps set_framerate should just fail if it's not compatible with the frame size, and vice versa?

common/omv_csi.h
```diff
@@ -410,6 +410,7 @@ typedef struct _omv_csi {
     int (*snapshot) (omv_csi_t *csi, image_t *image, uint32_t flags);
     int (*post_process) (omv_csi_t *csi, image_t *image, uint32_t flags);
     int (*isp_reset) (omv_csi_t *csi);
+    int (*isp_set_src) (omv_csi_t *csi, uint32_t src_w, uint32_t src_h);
```
iabdalkader: Having an ISP scaler is a property of the chip/port not the sensor. I suggest the following API:

common/omv_csi.h:
```c
uint16_t src_w;
uint16_t src_h;
```

Convention: src_w == 0 means the sensor produces exactly resolution[csi->framesize] (the default, every existing driver
unchanged).

Sensor responsibility:
```C
// ps5520_init():
csi->src_w = SENSOR_WIDTH;
csi->src_h = SENSOR_HEIGHT;
```

Port driver responsibility:

Owns the scaler entirely. Reads:
```c
uint32_t src_w = csi->src_w ? csi->src_w : csi->resolution[csi->framesize][0];
uint32_t src_h = csi->src_h ? csi->src_h : csi->resolution[csi->framesize][1];
```
Drives DCMIPP crop/decimate/downsize from (src_w, src_h) to (dst_w, dst_h).

You can gate all of that with `OMV_CSI_HW_SCALE_ENABLE` or something like that.

iabdalkader (changes requested):

drivers/sensors/ps5520.c
kwagyeman: Kept all the resolution tables when OMV_CSI_HW_SCALE_ENABLE is not defined. I looked into refactoring the register sets to reduce the duplication; however, it's not a trivial amount of changes and should be in another PR.

drivers/sensors/pag7936.c
kwagyeman: They need to be coupled in some way. The issue is that you want to generally scale the image from the highest possible resolution down to the target res to get the best image quality.

It's easy enough to bail in set_framesize if you select a resolution that can't be supported by a target frame rate that has already been chosen, like if 470 FPS was set already then fail at selecting anything but QVGA or below. But with set_framerate, I want to scale from HD->target_res if the FPS is 120 or less. If you specify a higher frame rate, then I want to scale from VGA->target_res if the FPS is <= 240 FPS and QVGA if <= 470 FPS.

If I don't do this, then we end up always scaling from HD, which caps the FPS at 120. For example, the user sets the frame size to QVGA without setting the frame rate. The optimal is to select HD->scale_down->QVGA. This gives the best image quality. If they then set the fps to 470, then I want to select QVGA and output that directly without scaling down to honor the speed they asked for at the cost of image quality.

common/omv_csi.h

drivers/sensors/ps5520.c
kwagyeman: I think this is resolved.

ports/stm32/stm_isp.c
```diff
@@ -206,4 +210,124 @@ int stm_isp_update_gamma_table(omv_csi_t *csi, uint32_t pipe,
...
+    };
+
+    if (HAL_DCMIPP_PIPE_SetCropConfig(&csi->dcmipp, pipe, &ccfg) != HAL_OK ||
+        HAL_DCMIPP_PIPE_EnableCrop(&csi->dcmipp, pipe) != HAL_OK) {
+        return -1;
+    }
+
+    DCMIPP_DecimationConfTypeDef dec = {
+        .HRatio = h_dec << DCMIPP_P1DECR_HDEC_Pos,
+            .VRatio = v_dec << DCMIPP_P1DECR_VDEC_Pos,
```
iabdalkader:
```suggestion
        .HRatio = (h_dec << DCMIPP_P1DECR_HDEC_Pos),
        .VRatio = (v_dec << DCMIPP_P1DECR_VDEC_Pos),
```
To fix the indentation.

kwagyeman: Mmm, nice trick. Without the (), the formatter forces that indentation. Will remember.

ports/stm32/omv_csi.c
```diff
@@ -590,15 +590,7 @@ static int stm_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
...
-            // Configure crop
-            DCMIPP_CropConfTypeDef ccfg = {
-                .HStart = fb->x,
-                .VStart = fb->y,
-                .HSize = fb->u,
-                .VSize = fb->v,
-            };
-            if (HAL_DCMIPP_PIPE_SetCropConfig(&csi->dcmipp, DCMIPP_PIPE, &ccfg) != HAL_OK ||
-                HAL_DCMIPP_PIPE_EnableCrop(&csi->dcmipp, DCMIPP_PIPE) != HAL_OK) {
+            if (stm_isp_set_scaler(csi, DCMIPP_PIPE)) {
```
iabdalkader, line 593: Isn't this a lot to call here on every snapshot, in case of 1 FB?

kwagyeman: The user could reconfigure all settings between each frame. It's not a particularly heavy C-wise overhead for it to be rerun, though. The function is bounded in time, so it's not really noticeable overhead.

To fix, I'll need to add some vars to track the applied state. Continue? Seems like a bit of overkill for the single-buffer case, though.

iabdalkader (approved):

PR #3141, opened 2026-04-25
modules: Add x/y position support to polar transforms.

### LINKED ISSUE #1463: linpolar update
Playing with an [OpenMV M7](https://organicmonkeymotion.wordpress.com/2021/10/28/zoom-zoom-zoom/), a bloggie [MHS-PM5 lens mounted with a simple 3D printed adapter and obstacle avoidance code example](https://organicmonkeymotion.wordpress.com/2021/10/17/esp32-v-smt32/) I find linpolar needs a means to move the centre of the polar image, since assumptions the polar image is centred on the sensor centre is problematic.

I can twist the bloggie lens on the mount to physically move the polar centre around the sensor but that can only go so far.

Just a couple of optional parameters for the x,y of the centre would be grand.   So:

`linpolar(reverse=True|False, (x,y))`

Thanks for any consideration.

Cheers,
Baz

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/1463

Ads support to control the linear/log polar center coordinate for the transform.

### DIFF
--- a/modules/py_image.c
+++ b/modules/py_image.c
@@ -2609,8 +2609,10 @@ static MP_DEFINE_CONST_FUN_OBJ_KW(py_image_bilateral_obj, 2, py_image_bilateral)
 
 #ifdef IMLIB_ENABLE_LINPOLAR
 static mp_obj_t py_image_linpolar(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
-    enum { ARG_reverse };
+    enum { ARG_x, ARG_y, ARG_reverse };
     static const mp_arg_t allowed_args[] = {
+        { MP_QSTR_x,       MP_ARG_OBJ | MP_ARG_KW_ONLY, {.u_rom_obj = MP_ROM_NONE} },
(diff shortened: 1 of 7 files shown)


modules/py_image.c
```diff
@@ -2676,9 +2676,11 @@ static MP_DEFINE_CONST_FUN_OBJ_KW(py_image_bilateral_obj, 2, py_image_bilateral)
 
 #ifdef IMLIB_ENABLE_LINPOLAR
 static mp_obj_t py_image_linpolar(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
-    enum { ARG_reverse };
+    enum { ARG_reverse, ARG_x, ARG_y };
```
iabdalkader:
```suggestion
    enum { ARG_x, ARG_y, ARG_reverse };
```

To match the C-level function, and x,y args are typically first.

iabdalkader (changes requested):

iabdalkader (approved):

PR #3143, opened 2026-04-26
modules: Make linear regression always run on a sane image size.

### LINKED ISSUE #886: Robust Linear Regression not stable with horizontal line.
Need to fix this for https://forums.openmv.io/viewtopic.php?f=6&t=2006&p=12209#p12209

kwagyeman:
- Remove the lead-squares linear regression, which isn't particularly useful at all.
- Handle large image sizes passed to get_regression.
- Add a snap feature to fix issues with alignment on horizontal/vertical extremes.

Fixes: https://github.com/openmv/openmv/issues/886


lib/imlib/stats.c
```diff
@@ -879,23 +879,27 @@ bool imlib_get_regression(find_lines_list_lnk_data_t *out,
...
-    if (!robust) {
-        // Least Squares
+    // Theil-Sen Estimator
+    int *x_histogram = uma_calloc(ptr->w * sizeof(int), UMA_DTCM);
+    int *y_histogram = uma_calloc(ptr->h * sizeof(int), UMA_DTCM);
+    long long *x_delta_histogram = uma_calloc((2 * ptr->w) * sizeof(long long), UMA_DTCM);
+    long long *y_delta_histogram = uma_calloc((2 * ptr->h) * sizeof(long long), UMA_DTCM);
+
+    size_t size = uma_avail(0);
+    point_t *points = (point_t *) uma_malloc(size, 0);
```
iabdalkader: It's best if you avoid this pattern in new code. Use `uma_realloc(size, UMA_DTCM | UMA_MAYBE);` with a reasonable initial size. It doesn't map to any other allocator (no libc get me all avail memory), while realloc can map to realloc with no flags.

- If there's not enough memory for the init size, pool gets downgraded to slow memory.
- If subsequent realloc's can grow in place, it will remain in DTCM, if not it will get downgraded
- Stop when realloc fails.
- You'll get fast memory if available.
- Penalty is one memcpy if pointer can't grow, so maybe use FAST.

```C
size = 100;
points = NULL;
points = uma_realloc(points, size, UMA_DTCM | UMA_MAYBE);

if (need_more_points) {
    size *= 2;
    new_points = uma_realloc(points, size, UMA_DTCM | UMA_MAYBE);
    if (!new_points) {
        printf("memory is finite 😢\n");
        uma_free(points);
        return;
    }
    points = new_points;
}
```

kwagyeman: Done, got 10 more fps. Like 60 hz now.

iabdalkader (approved):

PR #3152, opened 2026-04-28
modules/py_image: Use attrtuples for objects.

iabdalkader:
Replace custom struct/print/subscr/getter/dict/type with attrtuples for all result objects that are pure data.

This eliminates ~1900 lines of boilerplate while preserving the same attribute access API (now without parentheses).

Converted to attrtuple: similarity, statistics, percentile, threshold, line, circle, rect, qrcode, apriltag, datamatrix, barcode, displacement, kptmatch.

Kept as attr pattern (lazy compute): blob.
Kept as custom type (methods with args): histogram.

I'll update the examples/test once we finalize this.


modules/py_image.c
```diff
@@ -172,128 +172,11 @@ static MP_DEFINE_CONST_OBJ_TYPE(
...
-    { MP_ROM_QSTR(MP_QSTR_y),       MP_ROM_PTR(&py_kptmatch_y_obj)       },
-    { MP_ROM_QSTR(MP_QSTR_w),       MP_ROM_PTR(&py_kptmatch_w_obj)       },
-    { MP_ROM_QSTR(MP_QSTR_h),       MP_ROM_PTR(&py_kptmatch_h_obj)       },
-    { MP_ROM_QSTR(MP_QSTR_count),   MP_ROM_PTR(&py_kptmatch_count_obj)   },
-    { MP_ROM_QSTR(MP_QSTR_theta),   MP_ROM_PTR(&py_kptmatch_theta_obj)   },
-    { MP_ROM_QSTR(MP_QSTR_match),   MP_ROM_PTR(&py_kptmatch_match_obj)   },
-    { MP_ROM_QSTR(MP_QSTR_rect),    MP_ROM_PTR(&py_kptmatch_rect_obj)    }
+static const qstr kptmatch_fields[] = {
+    MP_QSTR_x, MP_QSTR_y, MP_QSTR_w, MP_QSTR_h, MP_QSTR_cx, MP_QSTR_cy,
+    MP_QSTR_count, MP_QSTR_theta, MP_QSTR_match, MP_QSTR_rect,
```
kwagyeman: This seems rather far away from the first usage in the file, which is thousands of lines away. Otherwise, it looks good!

kwagyeman (approved):

PR #3161, opened 2026-05-04
ports/stm32: Add ccm support.

kwagyeman:
Adds color matrix calibration support for the PS5520 and PAG7936. The color matrices come from values that Pixart tuned for OpenMV using the N6 with either sensor. CCM values are mixed together using Piecewise linear bracketing, which is an industry-standard way to adjust the applied color matrix based on the 

Here are some examples of what this does:

PS5520 before:
FHD-before

PS5520 After:
FHD-after

PAG7936 Before:
HD_before

PAG7936 After:
HD_after

...

The PS5520 appears to work very well with this PR. For the PAG7936, this greatly amplifies the noise from the PAG7936 sensor. The denoise feature of the PAG7936 needs to be activated to reduce this. Additionally, the PAG7936 suffers from limited dynamic range when viewing outdoors and in dark scenes, causing bad image quality.

Improvements to the PAG7936 driver need to be applied before this PR is usable. In particular, the denoise feature of the PAG7936 sensor needs to be enabled to reduce the RGB noise in the image.

### DIFF
--- a/common/omv_csi.h
+++ b/common/omv_csi.h
@@ -270,6 +270,18 @@ typedef enum {
 
 typedef struct _omv_csi omv_csi_t;
 
+typedef struct {
+    float coeffs[3][3];
+    float offsets[3];
+    float avg[3];
+} omv_csi_ccm_entry_t;
+
+// Entries must be sorted warmest to coolest (descending R/B ratio of avg).
+typedef struct {
+    size_t count;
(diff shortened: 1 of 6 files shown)


iabdalkader:

https://forums.openmv.io/t/ps5520-5mp-support-request-for-pre-release-firmware-pr-3114-3132/11646/2

iabdalkader:

PR #3162, opened 2026-05-04
drivers/sensors: Raise PAG7936 Y AEC Target on MIPI.

kwagyeman:
Increases the Y target for the onboard AEC/AGC algorithm on the PAG7936 when used in MIPI mode to compensate for the faster framerate.


kwagyeman:
Not the right fix.

PR #3163, opened 2026-05-04
drivers/sensors: Lower PAG7936 mipi rate to 800 Mb/s max.

kwagyeman:
Not quite sure what was going on here, but the PAG7936 MIPI rate is set for 1200 Mb/s when the datasheet says the max is 800 Mb/s. 

Couldn't find any emails from Pixart that say it does more than 800 Mb/s... The sensor appears to work with the setting changed to 800 Mb/s or 1200 Mb/s.

### DIFF
--- a/drivers/sensors/pag7936.c
+++ b/drivers/sensors/pag7936.c
@@ -782,7 +782,7 @@ int pag7936_init(omv_csi_t *csi) {
     csi->cfa_format = SUBFORMAT_ID_BGGR;
     #if OMV_PAG7936_MIPI_CSI2
     csi->mipi_if = 1;
-    csi->mipi_brate = 1200;
+    csi->mipi_brate = 800;
     #endif
 
     // Initialize csi ops.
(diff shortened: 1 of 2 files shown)


iabdalkader:
Can you still get > 400 FPS with that?

kwagyeman:
Will debug with Pixart and figure it out.

kwagyeman:
Pixart gave me the register settings for updating this. Will finish verification tomorrow.

kwagyeman:
@iabdalkader - 480-ish FPS output works now.

```
import csi
import time

csi0 = csi.CSI()
csi0.reset()
csi0.pixformat(csi.RGB565)
csi0.framesize(csi.QVGA)
csi0.framerate(480)
csi0.snapshot(time=2000)

clock = time.clock()

while True:
    clock.tick()
    img = csi0.snapshot()
    print(clock.fps())
```

The issue was that the real max is 470. Using 480 resulted in the hardware getting a frame time value so low that it started dropping every other frame.

iabdalkader (approved):

PR #3165, opened 2026-05-04
common: Fix green AWB blowout on bright scenes.

kwagyeman:
Fixes the first frame being green on the STM32 (Alif doesn't have this problem) and AWB turning the image green on bright scenes.


iabdalkader:
This fix breaks VGA PAG @kwagyeman reverting it. If you're sending it again, please test it first across all sensors.

kwagyeman:
Okay, it must be the second commit that drops the frame.

PR #3166, opened 2026-05-05
ports/stm32: Raise gamma limit clamp.

kwagyeman:
The PAG7936 on the AE3 and N6 brightness now match.

This appears to be okay inside at night. However, would like to test a bit during the day to make sure this is right to do.


kwagyeman:
This is good.

PR #3176, opened 2026-05-11
csi: Add OMV_CSI_FLAG_NO_UPDATE flag.

### LINKED ISSUE #3157: Double Check GENX320 on RT1062, H7 Plus, and N6.
Noticed glitches on RT1062 in event mode. This was working last time I checked. Investigation.

kwagyeman:
Replace the implicit "always update preview on snapshot when source
matches" behavior with a self-documenting opt-out: by default the
matched CSI's framebuffer is pushed to the IDE stream, and callers that
must bypass it (e.g. raw event reads where the framebuffer is not a
viewable image) pass OMV_CSI_FLAG_NO_UPDATE.

Apply NO_UPDATE in the genx320 READ_EVENTS, READ_EVENTS_RAW, and
CALIBRATE paths so the raw 1024xN event buffer is never pushed to the
preview, which previously raced against the user's img.flush() and
produced scrambled frames in event mode.

The issue with raw event data being shown in the IDE preview is fixed.

Fixes: https://github.com/openmv/openmv/issues/3157

### DIFF
--- a/common/omv_csi.h
+++ b/common/omv_csi.h
@@ -193,8 +193,9 @@ typedef enum {
 } omv_csi_framesize_t;
 
 typedef enum {
-    OMV_CSI_FLAG_NON_BLOCK      = (1 << 0),
-    OMV_CSI_FLAG_NO_POST        = (1 << 1),
+    OMV_CSI_FLAG_NO_UPDATE      = (1 << 0),
+    OMV_CSI_FLAG_NON_BLOCK      = (1 << 1),
+    OMV_CSI_FLAG_NO_POST        = (1 << 2),
     OMV_CSI_FLAG_IOCTL_ABORT    = (1 << 8),
 } omv_csi_flags_t;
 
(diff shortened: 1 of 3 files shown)


common/omv_csi.h
```diff
@@ -193,8 +193,9 @@ typedef enum {
 } omv_csi_framesize_t;
 
 typedef enum {
-    OMV_CSI_FLAG_NON_BLOCK      = (1 << 0),
-    OMV_CSI_FLAG_NO_POST        = (1 << 1),
+    OMV_CSI_FLAG_UPDATE_FB      = (1 << 0),
```
iabdalkader: I think it should be a negative flag, so by default if the CSI is the stream source it will update, and this is just self-documenting: `omv_csi_snapshot(csi, &image, OMV_CSI_FLAG_NO_UPDATE)` or NO_PREVIEW

iabdalkader (approved):

PR #3178, opened 2026-05-14
drvers/sensors: Apply pag7936 denoise settings.

kwagyeman:
Applies a minor denoise effect for HD and VGA. Works as expected. QVGA is not denoised as the max image size would be 316x196 when its enabled.

Before:
hd-before

After:
hd-after

### DIFF
--- a/drivers/sensors/pag7936.c
+++ b/drivers/sensors/pag7936.c
@@ -95,7 +95,9 @@
 #define ISP_EN_H                    (0x0800)
 #define ISP_EN_H_EN                 (0x01)
 #define ISP_TEST_MODE               (0x0801)
+#define DENOISE_EN                  (0x0882)
 #define ISP_TEST_MODE_RAMP          (0x04)
+#define R_RGB1_GRAY0                (0x0E08)
 #define ISP_WOI_EN                  (0x0E10)
 #define ISP_WOI_HSIZE_L             (0x0E11)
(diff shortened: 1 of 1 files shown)


iabdalkader:

drivers/sensors/pag7936.c
```diff
@@ -95,7 +95,9 @@
 #define ISP_EN_H                    (0x0800)
 #define ISP_EN_H_EN                 (0x01)
 #define ISP_TEST_MODE               (0x0801)
+#define Denoise_En                  (0x0882)
```
iabdalkader:
```suggestion
#define DENOISE_EN                  (0x0882)
```
Why the weird constants?

kwagyeman: Fixed

drivers/sensors/pag7936.c
```diff
@@ -95,7 +95,9 @@
 #define ISP_EN_H                    (0x0800)
 #define ISP_EN_H_EN                 (0x01)
 #define ISP_TEST_MODE               (0x0801)
+#define DENOISE_EN                  (0x0882)
 #define ISP_TEST_MODE_RAMP          (0x04)
+#define R_RGB1_Gray0                (0x0E08)
```
iabdalkader: Same comment, weird mixed-case constant.

kwagyeman: Sorry, should have caught that. Fixed.

PR #3180, opened 2026-05-17
tools/modelc: drop redundant npu_driver.py pass.

iabdalkader:
The two-step generate + npu_driver pattern came from pre-release ST scripts that staged the relocatable build manually. The stable stedgeai CLI does everything in one pass when invoked with --relocatable, so the second pass is dead weight (and requires an external make on PATH).


iabdalkader:
@kwagyeman FYI, saves a step, seems slightly faster, remove requirement to know where inner scripts are and a dependency on `make`.

kwagyeman:
@iabdalkader - Updated the IDE to use the relocatable flag. Works correctly.

PR #3182, opened 2026-05-19
lib/microdot: Add support for microdot to modern cameras.

kwagyeman:
Let's you create webservers on OpenMV Cams that work really well. Not sure what to add for examples yet. This PR just adds the base library.

```
# MJPEG streaming web server using microdot.
# Set SSID/KEY, run, open http://<printed-ip>/ in a browser.

import csi
import time
import errno
import network
import asyncio
from microdot import Microdot
from microdot.microdot import MUTED_SOCKET_ERRORS

# This platform reports a client disconnect as EIO (errno 5). Tell microdot to
# treat it as a normal disconnect so closing the browser tab doesn't crash the
# stream task.
if errno.EIO not in MUTED_SOCKET_ERRORS:
    MUTED_SOCKET_ERRORS.append(errno.EIO)

SSID = ""  # Network SSID
KEY = ""   # Network key

running = True  # cleared on shutdown so stream tasks stop themselves

# Init camera.
csi0 = csi.CSI()
csi0.reset()
csi0.pixformat(csi.RGB565)
csi0.framesize(csi.QVGA)

# Connect to WiFi.
wlan = network.WLAN(network.STA_IF)
wlan.active(True)
wlan.connect(SSID, KEY)
while not wlan.isconnected():
    print('Trying to connect to "{:s}"...'.format(SSID))
    time.sleep_ms(1000)
print("WiFi Connected", wlan.ifconfig())

app = Microdot()


@app.route("/")
async def index(request):
    return """<!doctype html>
<html>
  <head><title>OpenMV Stream</title><meta charset="UTF-8"></head>
  <body><h1>OpenMV MJPEG Stream</h1><img src="/video_feed"></body>
</html>""", 200, {"Content-Type": "text/html"}


@app.route("/video_feed")
async def video_feed(request):
    class stream:
        def __aiter__(self):
            return self

        async def __anext__(self):
            await asyncio.sleep(0)
            if not running:                 # shutdown requested
                raise StopAsyncIteration
            try:
                frame = csi0.snapshot().to_jpeg(quality=35).bytearray()
            except OSError:
                raise StopAsyncIteration
            return b"--frame\r\nContent-Type: image/jpeg\r\n\r\n" + \
                frame + b"\r\n"

        async def aclose(self):
            print("Stream closed.")

    return stream(), 200, {
        "Content-Type": "multipart/x-mixed-replace; boundary=frame"
    }


async def main():
    global running
    print("Serving at http://%s/" % wlan.ifconfig()[0])
    server = asyncio.create_task(
        app.start_server(host="0.0.0.0", port=80, debug=True)
    )
    try:
        while True:
            await asyncio.sleep_ms(100)     # finite poll timeout -> Ctrl-C works
    finally:
        running = False                     # tell live stream tasks to stop
        app.shutdown()                      # stop accepting new connections
        for _ in range(10):                 # let in-flight tasks exit (~0.5s)
            await asyncio.sleep_ms(50)


try:
    asyncio.run(main())
except KeyboardInterrupt:
    print("Server stopped.")
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,3 +17,6 @@
 [submodule "lib/apriltag"]
 	path = lib/apriltag
 	url = https://github.com/openmv/apriltag.git
+[submodule "scripts/libraries/microdot"]
+	path = scripts/libraries/microdot
+	url = https://github.com/miguelgrinberg/microdot.git
--- a/boards/OPENMV_AE3/manifest.py
+++ b/boards/OPENMV_AE3/manifest.py
@@ -19,6 +19,7 @@ require("webrepl")
 freeze ("$(OMV_LIB_DIR)/", "rtsp.py")
 freeze ("$(OMV_LIB_DIR)/", "mqtt.py")
 freeze ("$(OMV_LIB_DIR)/", "requests.py")
+require("microdot-lib")
 
 # Utils
 require("time")
--- a/boards/OPENMV_N6/manifest.py
+++ b/boards/OPENMV_N6/manifest.py
@@ -22,6 +22,7 @@ freeze ("$(OMV_LIB_DIR)/", "rpc.py")
 freeze ("$(OMV_LIB_DIR)/", "rtsp.py")
 freeze ("$(OMV_LIB_DIR)/", "mqtt.py")
 freeze ("$(OMV_LIB_DIR)/", "requests.py")
+require("microdot-lib")
 
 # Utils
 require("time")
--- a/boards/OPENMV_RT1060/manifest.py
+++ b/boards/OPENMV_RT1060/manifest.py
@@ -28,6 +28,7 @@ freeze ("$(OMV_LIB_DIR)/", "rpc.py")
 freeze ("$(OMV_LIB_DIR)/", "rtsp.py")
 freeze ("$(OMV_LIB_DIR)/", "mqtt.py")
 freeze ("$(OMV_LIB_DIR)/", "requests.py")
+require("microdot-lib")
 
 # Utils
 require("time")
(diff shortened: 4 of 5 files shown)


kwagyeman:
@iabdalkader - The script above works great until you stop it with the IDE. Then the camera hangs the serial protocol. There might be a bug with the new protocol and asyncio. Clicking stop in the IDE generates server stopped, but the camera locks up after that. This script crashes on the RT1062, N6, and AE3.

iabdalkader:
> The script above works great until you stop it with the IDE. Then the camera hangs the serial protoco

I have no problem connecting, starting & stopping the script multiple times with the Studio.

scripts/libraries/microdot/manifest.py
```diff
@@ -0,0 +1,23 @@
+metadata(version="2.6.2")
+
+# microdot.session uses PyJWT.
+require("pyjwt")
+
+# Freeze the curated set of microdot modules from the submodule source.
+package(
+    "microdot",
+    base_path="../../../lib/microdot/src",
```
iabdalkader: This belongs to scripts/. We could just add manifest to `scripts/libraries/microdot-lib/manifest.py` and move the submodule to `scripts/libraries`?

kwagyeman:
> I have no problem connecting, starting & stopping the script multiple times with the Studio.

Interesting. Okay, good to know.

scripts/libraries/microdot/manifest.py
kwagyeman: Done, will put any future python only libs here too.

PR #3183, opened 2026-05-24
Add pixel-exact blob contour drawing

MadlyFX:
## Summary
- add `Image.draw_contours()` for pixel-exact threshold contour drawing
- support drawing from a separate source image onto an overlay plus optional mask output
- add `seed=(x, y)` to restrict drawing to one connected threshold component
- add unit coverage for direct drawing, overlay compositing, masks, and seeded component selection

## Testing
- `git diff --check upstream/master...HEAD`
- `python -m py_compile scripts/unittest/tests/draw_contours.py`

[contours.bmp](https://github.com/user-attachments/files/28195333/contours.bmp)

### DIFF
--- a/lib/imlib/blob.c
+++ b/lib/imlib/blob.c
@@ -136,6 +136,103 @@ static float calc_roundness(float blob_a, float blob_b, float blob_c) {
     return IM_DIV(roundness_min, roundness_max);
 }
 
+void imlib_draw_contours(image_t *ptr, rectangle_t *roi, list_t *thresholds, bool invert, int color) {
+    if (!list_size(thresholds)) {
+        return;
+    }
+
+    image_t bmp;
+    bmp.w = ptr->w;
+    bmp.h = ptr->h;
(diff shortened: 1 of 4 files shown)


kwagyeman:
@MadlyFX - Thanks for the code. You should edit find_blobs() to accept a new argument which enables the collection of the blob contours as it's being detected. This would be a list of x/y point diffs, which could be 3 bits for x/y.

The 3 bits would encode up/down/left/right (or diagonals) or no change for x/y delta. Then, given an initial condition, storing this would enable a compact representation of the contour. You'd append to the list pixel position diffs each time the perimeter is increased.

Draw_countours would then start with the initial condition, and then draw the blob contour if it was present in the blob object.

Handling blob merge will be tricky. But, this approach would allow for compact contour representation.

kwagyeman:
Thanks, so please structure your PR as a series of commits, adding each feature, and also make sure each commit message passes our commit message checking system and code formatting.

Please see some examples of past PRs. Changes have to be atomic and reviewable. Each commit should be a specific thing added.

PR #3186, opened 2026-06-05
modules: Fix copy_to_fb image stream update. 

kwagyeman:
Fixes the frame buffer flush policy so drawing on loaded images works, per the
review discussion, plus a blocking `flush()`.

* **modules/py_imageio: Update WFI to event polling.** Fixed a leftover
  `__WFI()` in ImageIO.

* **modules/py_helper: Add a frame buffer lookup helper.** Additive only:
  `py_helper_get_framebuffer()` returns the frame buffer that
  `py_helper_set_to_framebuffer()` targets, for use by the following commits.

* **modules: Flush a pending frame buffer image before overwriting it.** The
  pending-flag design from the review: `py_helper_set_to_framebuffer()` marks
  the frame buffer pending and flushes the previous image -- with everything
  drawn on it -- right before the next image overwrites it. All load paths
  (ImageIO.read(), Image(), copy()/scale(), FIR/ToF snapshot with
  `copy_to_fb=True`) go through it, and their old immediate flushes (the
  before/after split) are removed. The camera snapshot path clears the flag
  when it flushes and invalidates the previous frame, so nothing ever
  double-flushes. Pending starts false, so nothing is flushed before the
  first-ever load.

* **modules: Make flush block until the IDE picks up the frame.**
  `img.flush(timeout=1000)` / `csi.flush(timeout=1000)` now block until the
  IDE has picked the frame up or the timeout (ms) expires -- so a flush() at
  the end of a script actually shows the frame. It also retries the write if
  the IDE held the streaming buffer (previously a silent drop). `timeout=0`
  gives the old non-blocking behavior, and with no IDE streaming it returns
  immediately, so headless scripts are unaffected. A manual flush clears the
  pending flag.

* **scripts/examples: Update copy2fb.py for the new flush behavior.** Its
  fixed "sleep so the IDE can read it" delay becomes a blocking `flush()`.

**Coverage**: every `copy_to_fb` load funnels through
`py_helper_set_to_framebuffer()`, and every capture path goes through the
common `omv_csi_snapshot()` (no port overrides the weak function), so those
are the only two places the pending flag lives. All 17 bundled examples that
load images were audited -- only copy2fb.py needed updating; the RPC raw
image transfer example, which previously flushed a blank buffer before the
data arrived, now displays the received frames.

**Testing**: verified on an OpenMV N6 -- the forum repro (ImageIO read +
draw loop), back-to-back constructor loads, load/snapshot interleave, and
flush timings headless (0 ms with no IDE attached). Each commit builds
independently; builds checked for OPENMV_N6, OPENMV_RT1060, OPENMV4P and
OPENMV_AE3.

### DIFF
--- a/lib/imlib/framebuffer.h
+++ b/lib/imlib/framebuffer.h
@@ -81,6 +81,7 @@ typedef struct framebuffer {
     uint8_t enabled;        // Enable/disable framebuffer
     uint8_t quality;        // JPEG compression quality (1-100)
     uint8_t raw_enabled;    // Enable raw streaming
+    uint8_t pending;        // Image loaded into the frame buffer but not previewed yet.
     uint32_t source;        // Stream buffer source ID.
     size_t raw_size;        // Raw buffer size.
     char *raw_base;         // Raw buffer address.
--- a/modules/py_fir.c
+++ b/modules/py_fir.c
@@ -888,6 +888,10 @@ mp_obj_t py_fir_snapshot(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_a
         .pixfmt = args[ARG_pixformat].u_int,
     };
     if (args[ARG_copy_to_fb].u_bool) {
+        framebuffer_t *fb = framebuffer_get(FB_MAINFB_ID);
+        image_t tmp;
+        framebuffer_to_image(fb, &tmp);
+        framebuffer_update_preview(&tmp);
         py_helper_set_to_framebuffer(&dst_img);
     } else {
         image_alloc(&dst_img, image_size(&dst_img));
@@ -968,9 +972,6 @@ mp_obj_t py_fir_snapshot(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_a
 
     uma_free(src_img.data);
 
-    if (args[ARG_copy_to_fb].u_bool) {
-        framebuffer_update_preview(&dst_img);
-    }
     return py_image_from_struct(&dst_img);
 }
 static MP_DEFINE_CONST_FUN_OBJ_KW(py_fir_snapshot_obj, 0, py_fir_snapshot);
--- a/modules/py_fir.c
+++ b/modules/py_fir.c
@@ -968,9 +968,6 @@ mp_obj_t py_fir_snapshot(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_a
 
     uma_free(src_img.data);
 
-    if (args[ARG_copy_to_fb].u_bool) {
-        framebuffer_update_preview(&dst_img);
-    }
     return py_image_from_struct(&dst_img);
 }
 static MP_DEFINE_CONST_FUN_OBJ_KW(py_fir_snapshot_obj, 0, py_fir_snapshot);
--- a/modules/py_helper.c
+++ b/modules/py_helper.c
@@ -298,14 +298,19 @@ void py_helper_update_framebuffer(image_t *img) {
 }
 
 // TODO need to pass a CSI here.
-void py_helper_set_to_framebuffer(image_t *img) {
+framebuffer_t *py_helper_get_framebuffer(void) {
     #if MICROPY_PY_CSI
-    omv_csi_t *csi = omv_csi_get(-1);
-    framebuffer_t *fb = csi->fb;
-
-    omv_csi_abort(csi, true, false);
+    return omv_csi_get(-1)->fb;
     #else
-    framebuffer_t *fb = framebuffer_get(FB_MAINFB_ID);
+    return framebuffer_get(FB_MAINFB_ID);
+    #endif
+}
+
+void py_helper_set_to_framebuffer(image_t *img) {
+    framebuffer_t *fb = py_helper_get_framebuffer();
+
+    #if MICROPY_PY_CSI
+    omv_csi_abort(omv_csi_get(-1), true, false);
     #endif
 
     // Resize the frame buffer to fit the biggest uncompressed image.
--- a/modules/py_helper.c
+++ b/modules/py_helper.c
@@ -313,6 +313,15 @@ void py_helper_set_to_framebuffer(image_t *img) {
     omv_csi_abort(omv_csi_get(-1), true, false);
     #endif
 
+    // Flush an image that was previously loaded into the frame buffer (along
+    // with anything drawn on it) before overwriting it.
+    if (fb->pending) {
+        image_t tmp;
+        framebuffer_to_image(fb, &tmp);
+        framebuffer_update_preview(&tmp);
+        fb->pending = false;
+    }
+
     // Resize the frame buffer to fit the biggest uncompressed image.
     framebuffer_resize(fb, 1, OMV_MAX(image_size(img), fb->u * fb->v * 2));
 
@@ -327,6 +336,10 @@ void py_helper_set_to_framebuffer(image_t *img) {
     img->data = buffer->data;
 
     framebuffer_release(fb, FB_FLAG_FREE);
+
+    // The caller will now load an image into the frame buffer; mark it to be
+    // flushed to the preview by the next frame buffer update.
+    fb->pending = true;
 }
 
 void py_helper_get_array_min_n(mp_obj_t obj, size_t min_n, mp_obj_t **items) {
--- a/common/omv_csi.c
+++ b/common/omv_csi.c
@@ -1672,6 +1672,9 @@ __weak int omv_csi_snapshot(omv_csi_t *csi, image_t *image, uint32_t flags) {
 
         // Release the previous buffer from used queue -> free queue.
         framebuffer_release(csi->fb, FB_FLAG_USED | FB_FLAG_INVALIDATE);
+
+        // The previous image (if any) has been flushed and invalidated.
+        csi->fb->pending = false;
     }
 
     // Toggle FSYNC.
--- a/modules/py_helper.h
+++ b/modules/py_helper.h
@@ -26,6 +26,7 @@
 #ifndef __PY_HELPER_H__
 #define __PY_HELPER_H__
 #include "imlib.h"
+#include "framebuffer.h"
 
 typedef enum py_helper_arg_image_flags {
     ARG_IMAGE_ANY          = (0 << 0),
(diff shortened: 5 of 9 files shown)


iabdalkader:

https://forums.openmv.io/t/issue-drawing-on-an-imageio-stream/11695/4

modules/py_fir.c
```diff
@@ -888,6 +888,10 @@ mp_obj_t py_fir_snapshot(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_a
         .pixfmt = args[ARG_pixformat].u_int,
     };
     if (args[ARG_copy_to_fb].u_bool) {
+        framebuffer_t *fb = framebuffer_get(FB_MAINFB_ID);
+        image_t tmp;
+        framebuffer_to_image(fb, &tmp);
+        framebuffer_update_preview(&tmp);
```
iabdalkader: > Copy the previous image to the stream buffer before grabing the next.

But there's no previous image on the first frame? What does this fix exactly and how to reproduce it?

kwagyeman: This thread: https://forums.openmv.io/t/issue-drawing-on-an-imageio-stream/11695?u=kwagyeman

We previously flushed after loading the frame. However, flushing before loading the frame would make it mimick snapshot() allowing you to draw on the buffer.

This, of course, changes a frame being displayed immediately...

...

Uh, this is actually a policy-level question. Which we might wish to fix. The question is, should we make all functions that modify the frame buffer flush the previous image in it before they load whatever they want? This then allows drawing on the loaded image.

However, it also comes at the cost of needing to use the function in a loop, as something like Image(copy_to_fb=True) would no longer show the frame immediately.

...

We've oscillated on this over the years, but the current way snapshot works seems the least surprising behavior to folks, and to just standardize on that.

iabdalkader: Flush before and after? In fact, could just move this to the helper:

```C
image_t tmp;
framebuffer_t *fb = framebuffer_get(FB_MAINFB_ID);
framebuffer_to_image(fb, &tmp);
framebuffer_update_preview(&tmp);

py_helper_set_to_framebuffer(&image);
```

iabdalkader: Another idea, add `dirty` flag to fb, set it on every draw function, set_pixel etc.. then you can check if FB is dirty -> flush before overwriting (in `py_helper_set_to_framebuffer`?).

kwagyeman: I like the dirty flag idea, but, it doesn't actually solve the flushing policy problem. It just reduces flushing to the minimum required.

Flushing before and after doesn't work, as you'd end up just picking one or the other. It would look glitchy as the fb lock system would just end up allowing before/after flush through randomly.

I think just moving all operations to work like how snapshot works is the right fix. Then, update any examples that depended on the old behavior. This will then be consistent for everyone. Finally, if you need to force a flush... you can always just flush().

iabdalkader: > I think just moving all operations to work like how snapshot works is the right fix. Then, update any examples that depended on the old behavior. This will then be consistent for everyone. Finally, if you need to force a flush... you can always just flush().

This breaks the "load/read image once" behavior though, also the last frame will never be displayed (if not playing in a loop).

>  Finally, if you need to force a flush... you can always just flush().

Same could be said/done to fix this issue without changes, right?

> We've oscillated on this over the years, 

Because flushing before breaks something and flushing after breaks something else. BTW, the code base is not consistent, it's a split, and this PR doesn't fix that either.

kwagyeman: > This breaks the "load/read image once" behavior though, also the last frame will never be displayed (if not playing in a loop).

That was my intention. You'd just have to manually flush instead, and I'd update the examples.

> Same could be said/done to fix this issue without changes, right?

The auto-flush before the image updates, though, is less easy to work around, as the frame buffer locks after you do it. So, trying to flush later after image updates results in flicker and inconsistencies, as the second flush may not get picked up. Typically, in any loop, the flush of the previous frame is what people want. So, it works without needing an explicit flush. Then, for the case where you load a single image and want to see it immediately, that case makes sense for a manual flush right after a load.

> Because flushing before breaks something and flushing after breaks something else. BTW, the code base is not consistent, it's a split, and this PR doesn't fix that either.

I'd update the PR to fix all of these.

Anyway, let me know what you want to do. The original user problem is this:

```
import image
import time

stream = image.ImageIO("test2.bin", "r")

while True:
    clock.tick()

    img = stream.read(copy_to_fb=True, loop=True, pause=True)
    img.draw_string(10, 10, "HELLO", color=(255, 0, 0))
    img.draw_rectangle(0, 0, 100, 100, color=(255, 0, 0))

    print(clock.fps())
```

iabdalkader: > Anyway, let me know what you want to do. The original user problem is this:

I'm still thinking about it, but I think there can't be one policy for everything. Here's what I have so far:
1- For snapshot, and snapshot-like functions: Flush the previous frame before capture.
2- For things that load/copy images to FB:
 - Set a pending flag after the update/copy instead of update_preview.
 - `py_helper_set_to_framebuffer` checks if the flag is set, flushes before overwriting.

Something like this:
```
read N:  set_to_framebuffer -> flush pending (N-1, with drawings) -> write N
         set_pending(fb)                                          -> mark N
*user draws on N*
read N+1:  set_to_framebuffer -> flush pending (N,   with drawings) -> write N+1
```

kwagyeman: Mmm, that basically works. The user still needs to flush to see an immediate update, but it solves the two-flush issue, as the second one won't trigger anymore if you've already flushed.

Okay, let me know when to start updating things if you think the above covers it. Also, I'll add another commit for flush such that it has a wait by default for the IDE to pick up the frame or a 1-second timeout. This then makes it behave as expected (a true blocking flush). Because adding flush() at the end of a script doesn't actually work as the script exits before the IDE picked up the frame.

kwagyeman: Okay, just updated everything.

modules/py_helper.c
```diff
@@ -297,6 +297,14 @@ void py_helper_update_framebuffer(image_t *img) {
     }
 }
 
+framebuffer_t *py_helper_get_framebuffer(void) {
+    #if MICROPY_PY_CSI
```
iabdalkader, line 302: This util function can be called bellow in set_to_framebuffer (remove duplicate code).

modules/py_helper.c
```diff
@@ -308,6 +316,15 @@ void py_helper_set_to_framebuffer(image_t *img) {
...
     #endif
 
+    // Flush an image that was previously loaded into the frame buffer (along
+    // with anything drawn on it) before overwriting it.
+    if (fb->pending) {
+        fb->pending = false;
+        image_t tmp;
+        framebuffer_to_image(fb, &tmp);
+        framebuffer_update_preview(&tmp);
+    }
```
iabdalkader, line 323: Variables first, please.

```suggestion
    if (fb->pending) {
        image_t tmp;
        framebuffer_to_image(fb, &tmp);
        framebuffer_update_preview(&tmp);
        fb->pending = false;
    }
```

modules/py_helper.c
```diff
@@ -330,6 +339,10 @@ void py_helper_set_to_framebuffer(image_t *img) {
     img->data = buffer->data;
 
     framebuffer_release(fb, FB_FLAG_FREE);
+
+    // The caller will now load an image into the frame buffer; mark it to be
+    // flushed to the preview by the next frame buffer update.
+    fb->pending = true;
```
iabdalkader, line 342: Rough idea: instead of requiring an explicit flush after this (for the single frame case), we could flush in any of the protocol functions that get called periodically. Problem is, it could flush before/mid drawing.

lib/imlib/framebuffer.h
```diff
@@ -182,5 +182,12 @@ static inline void framebuffer_set_preview(framebuffer_t *fb, uint16_t width, ui
...
 // and raw preview is disabled, or copy it directly if already compressed.
-void framebuffer_update_preview(image_t *src);
+// Updates the streaming preview buffer with the image; returns true if the
+// frame was written to the streaming buffer (false if it was busy/disabled).
+bool framebuffer_update_preview(image_t *src);
+
+// Like framebuffer_update_preview(), but blocks until the frame has been
+// picked up by the IDE, or the timeout (in milliseconds) expires.  Returns
+// true if the frame was picked up.
+bool framebuffer_update_preview_blocking(image_t *src, uint32_t timeout_ms);
```
iabdalkader: I prefer to add this later, maybe after the release, needs a closer look.

iabdalkader (changes requested):

modules/py_helper.c
kwagyeman: Done -- set_to_framebuffer() now uses it.

modules/py_helper.c
kwagyeman: Done, applied as suggested.

modules/py_helper.c
kwagyeman: Yeah, the mid-drawing flush is the killer -- a periodic flush from the protocol
side could send a half-drawn frame (or tear between two draw calls).  Keeping
the flush points deterministic (next load, or an explicit flush()) avoids that
entirely, and the single-frame case is just one flush() call.

lib/imlib/framebuffer.h
kwagyeman: Dropped it from this PR -- it's a self-contained change on top of this, so
easy to revisit after the release.  The copy2fb example now uses the existing
non-blocking flush() plus the small delay (a flush is required there now,
since loading no longer displays the image by itself).

modules/py_helper.c
iabdalkader: Protocol is scheduled, it can't really run mid-drawing, unless there's a scheduling point there. It only becomes a problem if we switch to real threading in the future. I'm Not saying we should implement this though, just worth considering as it removes any manual steps.

kwagyeman: Uh, well, like, mid-drawing would be between me doing a bunch of stuff. I wouldn't want it to update until the next frame, when I know I am done with all updates.

iabdalkader: Okay, makes sense.

iabdalkader (approved):

PR #3187, opened 2026-06-06
ports/mimxrt: Add missing MIMXRT CAN/Encoder Deinits.

kwagyeman:
Enables CAN on the RT1062, requires https://github.com/openmv/micropython/pull/142, first. Also, fixes the missing encoder deinit so the new Encoder/Counter machine classes work.


ports/mimxrt/main.c
```diff
@@ -196,9 +199,7 @@ int main(void) {
     #if MICROPY_PY_CSI
     omv_csi_abort_all();
     #endif
-    #if MICROPY_PY_MACHINE_CAN
-    machine_can_irq_deinit();
-    #endif
+    machine_can_deinit_all();
```
iabdalkader, line 207:
```suggestion
    #if MICROPY_PY_MACHINE_CAN
    machine_can_deinit_all();
    #endif
```

ports/mimxrt/main.c
```diff
@@ -72,6 +72,9 @@
 #include "mimxrt_hal.h"
 #include "omv_protocol.h"
 
+extern void machine_can_deinit_all(void);
```
iabdalkader, line 76:
```suggestion
#if MICROPY_PY_MACHINE_CAN
extern void machine_can_deinit_all(void);
#endif
```

iabdalkader (changes requested):
Please update the submodule.

iabdalkader (approved):

PR #3190, opened 2026-06-09
modules: Add IMU burst read

chobitsfan:
This PR add `__read_reg_burst(addr: int, buf: bytearray)` to `imu` module. To read LSM6DSM fifo, it is faster and more reliable to use burst read multiple samples at once than multiple `__read_reg`. Tested on AE3.

### DIFF
--- a/modules/py_imu.c
+++ b/modules/py_imu.c
@@ -382,14 +382,24 @@ static mp_obj_t py_imu_write_reg(mp_obj_t addr, mp_obj_t val) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_2(py_imu_write_reg_obj, py_imu_write_reg);
 
-static mp_obj_t py_imu_read_reg(mp_obj_t addr) {
+static mp_obj_t py_imu_read_reg(size_t n_args, const mp_obj_t *args) {
     error_on_not_ready();
 
+    // Burst read into a caller-provided bytearray when buf is given.
+    if (n_args > 1 && args[1] != mp_const_none) {
+        mp_buffer_info_t bufinfo;
+        mp_get_buffer_raise(args[1], &bufinfo, MP_BUFFER_WRITE);
+        if (bufinfo.len > 0) {
+            LSM_FUNC(read_reg) (&dev_ctx, mp_obj_get_int(args[0]), bufinfo.buf, bufinfo.len);
+        }
+        return mp_const_none;
+    }
+
     uint8_t v;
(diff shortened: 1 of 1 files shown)


kwagyeman (approved):

kwagyeman:
Useful feature; maybe just exposing the FIFO would be easier. Anyway, this will wait until after the release.

chobitsfan:
> maybe just exposing the FIFO would be easier. 

Thank you for reviewing. There is other usage for burst read, for example, read timestamp register (0x40 ~ 0x42)

iabdalkader:
I wonder if it would be better to add this as an option arg to read_reg? @kwagyeman

kwagyeman:
That's a better idea. @chobitsfan - Can you do this:

For read, add the buffer object as a second argument. If present, read writes data into that equal to the length and returns the None. If not present, it returns the register int var using a read size of 1. This basically gives you a read_into ability.

Then, for future proofing, make write accept an int or bytearray, if a byte array, then write that much data in it.

chobitsfan:
Hi @iabdalkader and @kwagyeman Thank you for reviewing and suggestions. Yes, that is a better way to do. I will do it and testing. I will update it once it is done.

chobitsfan:
`__read_reg(addr: int, buf: bytearray = None)`. Both  `__read_reg(ADDR_L) __read_reg(ADDR_H)` and `__read_reg(ADDR_L, bytearray(2)` has been tested on AE3

kwagyeman:
Thanks, you need to redo the commits so there's only one and follow the commit message format. Looks good otherwise.

kwagyeman:
Also, is there value in a burst write? If so, please add it; otherwise, it's fine not to add it.

chobitsfan:
Hi @kwagyeman Thank you for reviewing

>  Thanks, you need to redo the commits so there's only one and follow the commit message format. Looks good otherwise.

Done

> Also, is there value in a burst write? If so, please add it; otherwise, it's fine not to add it.

I think we can skip burst writes for now. IMU registers are generally only written during initialization, so optimizing write speeds doesn't provide a noticeable speed boost. Burst reads, on the other hand, are very useful for fetching data.

kwagyeman:
Looks good. Note, the commit message and PR name should be:

`modules: Add IMU burst read.`

chobitsfan:
> Looks good. Note, the commit message and PR name should be:
> 
> `modules: Add IMU burst read.`

Thank you for fix, it is done

PR #3198, opened 2026-06-24
scripts/requests: fix requests library.

kaizhi-singtown:
While testing OpenAI-compatible API requests, I found and fixed several issues in scripts/libraries/requests.py:

- Decode chunked transfer responses.
- Add configurable request timeout.
- Fix json request body serialization.
- Encode string request bodies before sending.
- Handle response headers as bytes and parse them case-insensitively.

Tested on OpenMV Cam N6 with an HTTPS JSON request.

### DIFF
--- a/scripts/libraries/requests.py
+++ b/scripts/libraries/requests.py
@@ -77,7 +77,26 @@ def socket_readall(s):
     return buf
 
 
-def request(method, url, data=None, json=None, files=None, headers={}, auth=None, stream=None):
+def decode_chunked(data):
+    buf = b""
+    pos = 0
+    while True:
+        line_end = data.find(b"\r\n", pos)
+        if line_end < 0:
+            break
+        size_line = data[pos:line_end]
+        if b";" in size_line:
+            size_line = size_line.split(b";", 1)[0]
+        size = int(size_line, 16)
+        pos = line_end + 2
+        if size == 0:
+            break
+        buf += data[pos : pos + size]
+        pos += size + 2
+    return buf
+
+
+def request(method, url, data=None, json=None, files=None, headers={}, auth=None, stream=None, timeout=5.0):
     try:
         proto, dummy, host, path = url.split("/", 3)
     except ValueError:
@@ -104,12 +123,14 @@ def request(method, url, data=None, json=None, files=None, headers={}, auth=None
     resp_code = 0
     resp_reason = None
     resp_headers = []
+    chunked = False
 
     ai = socket.getaddrinfo(host, port)[0]
     s = socket.socket(ai[0], ai[1], ai[2])
     try:
         s.connect(ai[-1])
-        s.settimeout(5.0)
+        if timeout is not None:
+            s.settimeout(timeout)
         if proto == "https:":
             s = ssl.wrap_socket(s, server_hostname=host)
 
@@ -126,9 +147,10 @@ def request(method, url, data=None, json=None, files=None, headers={}, auth=None
             s.write(b"\r\n")
 
         if json is not None:
-            import json
+            json_body = json
+            import json as json_module
 
-            data = json.dumps(json)
+            data = json_module.dumps(json_body)
             s.write(b"Content-Type: application/json\r\n")
 
         if files is not None:
@@ -145,6 +167,9 @@ def request(method, url, data=None, json=None, files=None, headers={}, auth=None
                 data += b"\r\n"
             data += b"\r\n--%s--\r\n" % (boundary)
 
+        if isinstance(data, str):
+            data = data.encode()
+
         if data:
             s.write(b"Content-Length: %d\r\n\r\n" % len(data))
             s.write(data)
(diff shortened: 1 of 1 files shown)


kwagyeman:
Looks fine.

iabdalkader:
@kaizhi-singtown Let me know if you're willing to update this PR and I'll post a review.

scripts/libraries/requests.py
```diff
@@ -77,7 +77,26 @@ def socket_readall(s):
     return buf
 
 
-def request(method, url, data=None, json=None, files=None, headers={}, auth=None, stream=None):
+def decode_chunked(data):
+    buf = b""
```
iabdalkader, line 81: I'd use a bytearray here + memoryview on data and save all those allocs on slices and new bytes object.

```suggestion
    buf = bytearray()
    mv = memoryview(data)
```


While we're at it, please rework this in the same way (also, if there's a recev_into + `mv` would be even better):
```Python
def socket_readall(s):
    buf = bytearray()
    while True:
        try:
            recv = s.recv(512)
        except OSError:
            break
        if not recv:
            break
        buf.extend(recv)
    return buf
```

scripts/libraries/requests.py
```diff
@@ -77,7 +77,26 @@ def socket_readall(s):
...
+        size = int(size_line, 16)
+        pos = line_end + 2
+        if size == 0:
+            break
+        buf += data[pos : pos + size]
+        pos += size + 2
+    return buf
+
+
+def request(method, url, data=None, json=None, files=None, headers={}, auth=None, stream=None, timeout=5.0):
```
iabdalkader: This mutates one shared dict across calls, unrelated bug, should fix.
```suggestion
def request(method, url, data=None, json=None, files=None,
            headers=None, auth=None, stream=None, timeout=5.0):
    headers = dict(headers) if headers else {}
```

scripts/libraries/requests.py
```diff
@@ -77,7 +77,26 @@ def socket_readall(s):
...
+    buf = b""
+    pos = 0
+    while True:
+        line_end = data.find(b"\r\n", pos)
+        if line_end < 0:
+            break
+        size_line = data[pos:line_end]
+        if b";" in size_line:
+            size_line = size_line.split(b";", 1)[0]
+        size = int(size_line, 16)
```
iabdalkader, line 90:
```suggestion
    try:
        size = int(size_line, 16)
    except ValueError:
        break
```

scripts/libraries/requests.py
```diff
@@ -156,19 +181,21 @@ def request(method, url, data=None, json=None, files=None, headers={}, auth=None
             l = response.pop(0).strip()
             if not l or l == b"\r\n":
                 break
-            if l.startswith(b"Transfer-Encoding:"):
```
iabdalkader, line 159: `b"HTTP" in l` can misidentify a header as the status line (e.g. Upgrade: HTTP/2), then crash on int(sline[1]). Since the status line is always the first line, parse it by position instead.

```Python
while response:
    l = response.pop(0).strip()
    if not l:
        break
    if resp_code == 0:
        sline = l.split(None, 2)
        resp_code = int(sline[1])
        resp_reason = sline[2].decode().rstrip() if len(sline) > 2 else ""
        continue
    lower = l.lower()
    if lower.startswith(b"transfer-encoding:"):
        chunked = b"chunked" in lower
    elif lower.startswith(b"location:") and not 200 <= resp_code <= 299:
        raise NotImplementedError("Redirects not yet supported")
    resp_headers.append(l)
```

iabdalkader (changes requested):

kaizhi-singtown:
Pushed 447a9631 with a few cleanups on top of the original fixes:

- Changed the `headers={}` default to `headers=None` (initialized inside the function). With the mutable default, one call with `auth=` would leave the `Authorization` header in the shared default dict and it would be sent on all subsequent requests.
- Moved `settimeout()` before `connect()` so the timeout also covers the connect phase, not just reads/writes.
- Dropped the redundant `json_body` temporary — with `import json as json_module` the `json` parameter is no longer shadowed.
- Simplified the status-line check to `b"HTTP" in l`, since `b"HTTPS"` already contains `b"HTTP"` as a substring.

scripts/libraries/requests.py
```diff
@@ -104,12 +125,14 @@ def request(method, url, data=None, json=None, files=None, headers={}, auth=None
     resp_code = 0
     resp_reason = None
     resp_headers = []
+    chunked = False
 
     ai = socket.getaddrinfo(host, port)[0]
     s = socket.socket(ai[0], ai[1], ai[2])
     try:
+        if timeout is not None:
+            s.settimeout(timeout)
```
iabdalkader, line 134: If I remember correctly, some drivers failed if timeout was set on unconnected socket.

scripts/libraries/requests.py
iabdalkader: Unresolved.

scripts/libraries/requests.py
iabdalkader: Unresolved.

iabdalkader (changes requested):

kwagyeman:
Replaced by: https://github.com/openmv/openmv/pull/3220

PR #3208, opened 2026-07-09
drivers/sensors: Add PAG7936 halt for safe shutdown. 

kwagyeman:
When used with an extension cable of 80mm or 120mm length, the PAG7936 locks up when reset without idling the sensor first. Sleeping the sensor first before reset/pwdn mitigates the issue. See: https://forums.openmv.io/t/n6-ribbon-cable/11808/3?u=kwagyeman

Tested against the N6 with a PAG7936 with no cable, a PAG7936 with a 120mm cable, and the dual PAG7936 + Lepton and GENX320 modules with no cables and cables. Also tested against the AE3.

### DIFF
--- a/common/omv_csi.c
+++ b/common/omv_csi.c
@@ -336,6 +336,12 @@ __weak int omv_csi_reset(omv_csi_t *csi, bool hard) {
     }
 
     if (hard) {
+        // Stop the sensor's activity cleanly.
+        if (csi->power_on && csi->halt_req) {
+            omv_csi_sleep(csi, true);
+            mp_hal_delay_ms(10);
+        }
+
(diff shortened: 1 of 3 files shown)


iabdalkader:
> the PAG7936 locks up when reset without idling the sensor first

Since it's very specific to PAG, and we already have many ops, maybe just always perform that in csi->shutdown for PAG driver and drop the new op?

kwagyeman:
The issue is that I have to perform the PAG7936 output halt before the physical reset or power-down lines are toggled. The sensor-specific ioctl is called afterward.

kwagyeman:
Updated

iabdalkader (approved):

PR #3212, opened 2026-07-20
scripts/libraries: Add waveform channel widget.

kwagyeman:
OpenMV IDE now has support for the CBOR channel widgets. I added a few widgets to make a complete package.

## New widget examples

Each is added to a `CBORChannel` with `ch.add(...)`. Display widgets are
updated with `ch[name] = ...`; control widgets deliver user input to the
channel's `on_write(ch, name, value)` callback.

### Displays

```python
# text - static or rich text (HTML), updatable live
ch.add("Status", type="text", value="<b>Ready.</b>")
ch["Status"] = "Running..."
```

```python
# waveform - a 1D time series plotted live. The value you assign is the raw
# bytes of an array of samples. add() declares the sample rate, the element
# type (typecode, "H"/uint16 by default), and the display range.
from array import array

ch.add("Mic", type="waveform", sample_rate=16000, min=0, max=65535)

mic = array("H", (0 for _ in range(512)))          # 512 uint16 samples
while True:
    for i in range(len(mic)):
        mic[i] = read_mic_sample()                 # fill the chunk
    ch["Mic"] = bytes(mic)                          # assign it as bytes
```

```python
# Multi-series: set series=N and interleave the samples sample-by-sample,
# i.e. [x0, y0, z0, x1, y1, z1, ...]. Here three float32 axes.
from array import array

ch.add("IMU", type="waveform", sample_rate=100, series=3,
       typecode="f", min=-2.0, max=2.0)

imu = array("f", (0.0 for _ in range(128 * 3)))    # 128 samples x 3 axes
while True:
    for i in range(128):
        x, y, z = read_imu()
        imu[(i * 3) + 0] = x
        imu[(i * 3) + 1] = y
        imu[(i * 3) + 2] = z
    ch["IMU"] = bytes(imu)
```

### Controls (values arrive in `on_write`)

```python
# pushbutton - momentary action, writes True on click
ch.add("Calibrate", type="pushbutton")

# spinbox - precise stepped number, also takes a (min, max, value) tuple
ch.add("Gap", type="spinbox", min=0.0, max=15.0, step=0.1, value=1.5, unit="mm")
ch["Gap"] = (0.0, 20.0, 2.5)                             # optionally rescale live

# radio - options rendered as radio buttons
ch.add("Mode", type="radio", options=["Idle", "Track", "Record"], value="Idle")

# lineedit - writable string
ch.add("Device Name", type="lineedit", value="openmv-cam")
```

### DIFF
--- a/scripts/libraries/protocol/protocol/__init__.py
+++ b/scripts/libraries/protocol/protocol/__init__.py
@@ -6,6 +6,9 @@
 # in the board's manifest and CBORChannel will be importable from
 # protocol directly.
 
+import struct
+import time
+
 import cbor2
 from uprotocol import *  # noqa: F401
 
@@ -14,6 +17,9 @@ from uprotocol import *  # noqa: F401
 _N = 0        # name
 _U = 1        # unit
 _V = 2        # numeric value
+_VS = 3       # string value
+_T = 6        # time
+_UT = 7       # update time
 _VD = 8       # data value
 
 # 2D data extension keys.
@@ -33,9 +39,11 @@ _W_OPTS = -34  # select options
 class CBORChannel:
     """A protocol channel backend that serializes named fields to CBOR.
 
-    Supports display widgets (label, depth) and interactive controls
-    (toggle, slider, select) with on_read/on_write callbacks.
-    Slider values can be set as (min, max, value) tuples.
+    Supports display widgets (label, text, depth, waveform) and
+    interactive controls (toggle, pushbutton, slider, spinbox, select,
+    radio, lineedit) with on_read/on_write callbacks. Slider and spinbox
+    values can be set as (min, max, value) tuples; waveform values are
+    set to interleaved sample bytes.
 
     Usage::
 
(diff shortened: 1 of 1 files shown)


iabdalkader:
Please squash this.

iabdalkader:
Possible issues:

1. _ITEMSIZE fallback silently mis-sizes 8-byte typecodes. _ITEMSIZE = {"b":1,"B":1,"h":2,"H":2} with _ITEMSIZE.get(f.get(_VS,"H"), 4). The fallback of 4 is correct for "f", "i", "I", "l", "L" (32-bit target), but "d" (float64) and "q"/"Q" (int64) are 8 bytes and would produce a width that's 2x too large, silently truncating/misplotting. Given this targets 32-bit MicroPython and the docs only advertise "H"/"f"/"h", this is a latent edge case rather than a real bug - but it's cheap to make robust by adding the 4- and 8-byte entries explicitly, or by driving itemsize off array(typecode).itemsize / struct.calcsize.
2. series=0 would divide by zero. f[_W] = len(value) // (itemsize * f[_H]) where f[_H] = series. Default is 1 and nobody sensible passes 0, so this is theoretical - worth a one-line guard only if you care about defensive input handling here (the rest of add() doesn't validate, so skipping it is consistent).
3. Docstring nit: the typecode arg help lists examples "f", "h" but the actual runtime default when omitted is "H" (uint16), which the text does state - just make sure the IDE side agrees that an absent _VS means uint16.

kwagyeman:
Done and retested.

iabdalkader (approved):

PR #3219, opened 2026-08-03
drivers/sensors: Select PAG7936 interface from its I2C address.

### LINKED ISSUE #2858: Remember to Change PAG7936 Driver to map I2C address to MIPI/Parallel Mode
0x15 (0x2A) -> MIPI
0x40 (0x80) -> Parallel

Make this change to the driver for the dual iToF and RGB sensor in the future.

kwagyeman:
The sensor's I2C address now selects whether the MIPI CSI-2 or the parallel interface registers and limits are used, instead of picking them at compile time. OMV_PAG7936_MIPI_CSI2 now only controls whether the MIPI CSI-2 support is built in.

Fixes: https://github.com/openmv/openmv/issues/2858

### DIFF
--- a/common/omv_csi.c
+++ b/common/omv_csi.c
@@ -465,9 +465,10 @@ static size_t omv_csi_detect(omv_i2c_t *i2c, i2c_dev_t *dev_list) {
             #endif // (OMV_FROGEYE2020_ENABLE == 1)
 
             #if (OMV_PAG7920_ENABLE == 1) || (OMV_PAG7936_ENABLE == 1)
-            // PAG720 and PAG7936 share the same I2C address.
-            case PAG7920_SLV_ADDR:
-            case PAG7936_SLV_ADDR_ALT:
+            // PAG7920 and PAG7936 share the same I2C address.
+            // case PAG7920_SLV_ADDR:
+            case PAG7936_SLV_ADDR:
+            case PAG7936_SLV_ADDR_MIPI:
                 omv_i2c_read_reg(i2c, slv_addr, PIXART_CHIP_ID, 2, &chip_id, 2);
                 chip_id = ((chip_id << 8) | (chip_id >> 8)) & 0xFFFF;
                 break;
--- a/drivers/sensors/pag7936.c
+++ b/drivers/sensors/pag7936.c
@@ -175,28 +175,30 @@
 #define PAG7936_EXPOSURE_M(exp)     (((exp) >> 8) & 0xff)
 #define PAG7936_EXPOSURE_L(exp)     ((exp) & 0xff)
 
-#if OMV_PAG7936_MIPI_CSI2
-#define PAG7936_WIDTH_ALIGN         (8)
-#define PAG7936_QVGA_FPS_MAX        (470)
-#define PAG7936_VGA_FPS_MAX         (240)
-#define PAG7936_HD_FPS_MAX          (120)
-#else
-#define PAG7936_WIDTH_ALIGN         (4)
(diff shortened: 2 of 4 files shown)


drivers/sensors/pag7936.c
```diff
@@ -896,17 +913,36 @@ static int ioctl(omv_csi_t *csi, int request, va_list ap) {
...
+        mipi_if = true;
+        pag7936_state.if_regs = mipi_regs;
+        pag7936_state.qvga_fps_max = PAG7936_MIPI_QVGA_FPS_MAX;
+        pag7936_state.vga_fps_max = PAG7936_MIPI_VGA_FPS_MAX;
+        pag7936_state.hd_fps_max = PAG7936_MIPI_HD_FPS_MAX;
+    }
+    #endif // OMV_PAG7936_MIPI_CSI2
+
+    if (!mipi_if) {
+        pag7936_state.if_regs = parallel_regs;
```
iabdalkader, line 924: I think maybe best to remove `OMV_PAG7936_MIPI_CSI2` and compile both sets of regs unconditionally, otherwise a misconfig can leave this unset. All cams that use this have plenty of flash, and it's a small array.

kwagyeman: Done

common/omv_csi.c
```diff
@@ -467,7 +467,7 @@ static size_t omv_csi_detect(omv_i2c_t *i2c, i2c_dev_t *dev_list) {
             #if (OMV_PAG7920_ENABLE == 1) || (OMV_PAG7936_ENABLE == 1)
             // PAG720 and PAG7936 share the same I2C address.
             case PAG7920_SLV_ADDR:
-            case PAG7936_SLV_ADDR_ALT:
+            case PAG7936_SLV_ADDR_MIPI:
```
iabdalkader, line 471: How does this match PAG7936 then for AE3?

kwagyeman: The AE3 is on PAG7920_SLV_ADDR and the N6 is on PAG7936_SLV_ADDR_MIPI. Should I rename PAG7920_SLV_ADDR -> PAG7920_SLV_ADDR_PARALLEL?

iabdalkader: I would actually rename `PAG7920_SLV_ADDR` to `PAG7936_SLV_ADDR`, the 7920 is old and we don't have it on any cam, so less important. Could possibly also drop `PAG7936_SLV_ADDR_MIPI` and just test for `addr != PAG7936_SLV_ADDR` in the driver. Otherwise, looks good.

kwagyeman: > I would actually rename PAG7920_SLV_ADDR to PAG7936_SLV_ADDR, the 7920 is old and we don't have it on any cam, so less important. 

Done

> Could possibly also drop PAG7936_SLV_ADDR_MIPI and just test for addr != PAG7936_SLV_ADDR in the driver. 

I need the PAG7936_SLV_ADDR_MIPI for the case statement.

kwagyeman:
Changes before were lost... fixing.

kwagyeman:
It's ready and tested on the N6 and AE3.

| Board | Interface branch | QVGA | VGA | Frames flat? |
|---|---|---|---|---|
| N6 (openmv-n6-pag7936) | MIPI (addr-selected) | 111.1 fps | 50.0 fps | 0 — all 45 real |
| AE3 (openmv-ae3-pag7936) | parallel (else branch) | 58.8 fps | 47.6 fps | 0 — all 45 real |

iabdalkader (approved):

PR #3220, opened 2026-08-03
scripts/libraries/requests: Fix the requests library.

kwagyeman:
Supersedes #3198, squashed into one commit with @kaizhi-singtown co-authored.
Original fixes are unchanged; this addresses @iabdalkader's review.

### Review comments

| Comment | Resolution |
|---|---|
| `decode_chunked`: use `bytearray` + `memoryview`, save the slice/concat allocs | Done. `buf = bytearray()`, `mv = memoryview(data)`, payload copied via `buf.extend(mv[pos:pos + size])`. |
| Rework `socket_readall` the same way | Done. `bytearray` + `extend`, and `recv(512)` instead of `recv(1)` — no more byte-at-a-time allocation. Bare `except:` narrowed to `except OSError`. |
| `recv_into` + `mv` would be even better | Not done. SSL-wrapped sockets don't reliably implement it, and `recv(512)` already removes the per-byte cost. Happy to revisit. |
| Wrap `int(size_line, 16)` in `try/except ValueError` | Done. A malformed chunk size now breaks out instead of raising. |
| `b"HTTP" in l` can misidentify a header (e.g. `Upgrade: HTTP/2`) and crash on `int(sline[1])` | Done. Status line parsed by position via the `resp_code == 0` guard, as suggested. |
| `headers={}` mutates one shared dict across calls | Done. `headers = dict(headers) if headers else {}` — the caller's dict is no longer modified and `auth` no longer leaks into later requests. |
| Some drivers fail if a timeout is set on an unconnected socket | Done. `settimeout()` moved after `connect()`, with a comment noting why. |

### Testing

Tested on an OpenMV Cam N6 against a local server serving deliberately awkward
responses — 20/20 assertions pass:

1. chunked decoding
2. chunk extensions (`4;name=value`) + `transfer-encoding: CHUNKED`
3. malformed chunk size — no exception, partial body
4. `Upgrade: HTTP/2` header — status stays 200, header preserved
5. plain `Content-Length` response (404/reason/body)
6. 3 s server delay vs `timeout=1` — returns in 1009 ms
7. redirect still raises `NotImplementedError`
8. json body serialization + `Content-Length`
9. `str` body encoded before send
10. `Authorization` does not leak into the next call
11. caller's headers dict unchanged

### Note

On a read timeout the request returns an empty `Response` with `status_code == 0`
rather than raising. That is pre-existing behaviour, left as-is; say the word if
it should raise instead.

### DIFF
--- a/scripts/libraries/requests.py
+++ b/scripts/libraries/requests.py
@@ -64,20 +64,45 @@ def readline(s):
 
 
 def socket_readall(s):
-    buf = b""
+    buf = bytearray()
     while True:
-        recv = b""
         try:
-            recv = s.recv(1)
-        except:
-            pass
-        if len(recv) == 0:
+            recv = s.recv(512)
+        except OSError:
             break
-        buf += recv
+        if not recv:
+            break
+        buf.extend(recv)
     return buf
 
 
-def request(method, url, data=None, json=None, files=None, headers={}, auth=None, stream=None):
+def decode_chunked(data):
+    buf = bytearray()
+    mv = memoryview(data)
+    pos = 0
+    while True:
+        line_end = data.find(b"\r\n", pos)
+        if line_end < 0:
+            break
+        size_line = data[pos:line_end]
+        if b";" in size_line:
+            size_line = size_line.split(b";", 1)[0]
+        try:
+            size = int(size_line, 16)
+        except ValueError:
+            break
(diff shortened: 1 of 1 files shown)


iabdalkader:
Please rebase this.

kwagyeman:
Done, and it's passing!

PR #3228, opened 2026-08-07
modules/py_csi_ng.c: Return custom frame size.

### LINKED ISSUE #2950: Missing way to get custom frame size resolution passed.

Daniiiil1:
## Summary

When CSI.framesize() is queried after configuring a tuple such as (320, 320), return the configured width and height instead of the internal CUSTOM enum.
The existing integer-enum return value remains unchanged for predefined frame sizes.

## Test coverage
The capture unit test now configures the virtual camera with a custom 160 by 120 frame size, verifies that the getter returns the same tuple, and then captures the existing reference frame.
Fixes #2950
## Local validation
- git diff --check\n- python3 -m py_compile scripts/unittest/tests/capture.py
- A complete firmware matrix build is left to project CI because the local checkout does not have the OpenMV SDK and target toolchains installed.

### DIFF
--- a/modules/py_csi_ng.c
+++ b/modules/py_csi_ng.c
@@ -317,6 +317,14 @@ static mp_obj_t py_csi_framesize(size_t n_args, const mp_obj_t *args) {
         if (self->csi->framesize == OMV_CSI_FRAMESIZE_INVALID) {
             omv_csi_raise_error(OMV_CSI_ERROR_INVALID_FRAMESIZE);
         }
+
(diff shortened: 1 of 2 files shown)


kwagyeman:
Already found in https://github.com/openmv/openmv/pull/2951

Daniiiil1:
Thanks for pointing that out — I missed #2951. Closing this duplicate.

PR #3229, opened 2026-08-07
modules/py_csi_ng: Return custom frame size

### LINKED ISSUE #2950: Missing way to get custom frame size resolution passed.

Daniiiil1:
## Summary

- return the configured width and height tuple when a CSI instance uses a custom frame size
- preserve the existing integer enum result for predefined frame sizes
- extend the capture regression test to verify the custom-size getter round trip

Previously, csi.framesize((width, height)) stored the custom resolution, but csi.framesize() exposed the internal custom enum instead of that resolution.

Closes #2950

## Validation

- python3 -m py_compile scripts/unittest/tests/capture.py
- git diff --check

The repository firmware matrix and QEMU/FVP tests are left to CI because they require the OpenMV SDK and target environments.


Daniiiil1:
I overlooked that this was already covered by #2951 and that the earlier duplicate had been closed. Closing this duplicate as well — apologies for the noise.

PR #3230, opened 2026-08-09
lib/micropython: Update submodule.

iabdalkader:
- stm32/mpu: Reuse MPU_ATTRIBUTES_NUMBER0 for all non-cacheable uses.
- stm32: Enable machine.mem_backup via backup SRAM or BKP registers.

### DIFF
--- a/ports/stm32/main.c
+++ b/ports/stm32/main.c
@@ -244,6 +244,9 @@ soft_reset:
     sdcard_init();
     #endif
     rtc_init_start(false);
+    #if MICROPY_PY_MACHINE_MEM_BACKUP
+    machine_mem_backup_init();
+    #endif
     #if MICROPY_PY_LWIP
     // lwIP can only be initialized once, because the system timeout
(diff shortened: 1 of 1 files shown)


kwagyeman:
Will test against user issue by EoD.

kwagyeman:
@iabdalkader - Looks like the PR doesn't compile? Can't test yet.

iabdalkader:
Was missing a commit. Fixed. I had to pull in a couple more for the patch to apply cleanly, easier to rebase later.

kwagyeman:
@iabdalkader This fixes the performance regression - works fine! As for the extra backup memory power draw. I wrote a Python script our customer is using with Niclas (H7) in Bulk to enable this via machine register writes, and they were happy with the power draw on that, so this just enables what customers were already doing. As for the N6, it's not meaningful given the N6's deepsleep current draw is in the 1mA range.

kwagyeman (approved):

PR #3231, opened 2026-08-10
lib/imlib: Update jpeg max size limit to be dynamic.

### LINKED ISSUE #3227: Max 1MB JPG on OpenMV 5.0.0
**Describe the bug**
When the JPEG resolution is set to WQXGA2 (5MP), the img.to_jpeg(quality=quality) fail with an OsError if quality is higher around 60%. 

The root cause is the Commit 1ecfabe

 - IDE version: 5.0.0
 - Firmware version: 5.0.0
 - Board: OpenMV N6 with PS5520 Camera

**To Reproduce**
```python
import csi
import time

csi0 = csi.CSI()
csi0.reset()
csi0.pixformat(csi.RGB565)
csi0.framesize(csi.WQXGA2)
clock = time.clock()

while True:
    clock.tick()
    img = csi0.snapshot()
    img.to_jpeg(quality=90)
    print(clock.fps())
```

kwagyeman:
Fixes: https://github.com/openmv/openmv/issues/3227

1MB is not large enough for a very complex scene for the maximum JPEG size with a 5MP image.

This change only modifies behavior when there's no destination buffer set. This does not affect the stream buffer, which passes in its destination size.

### DIFF
--- a/lib/imlib/imlib.h
+++ b/lib/imlib/imlib.h
@@ -835,7 +835,7 @@ bool image_get_mask_pixel(image_t *ptr, int x, int y);
 #define JPEG_444_YCBCR_MCU_SIZE    ((JPEG_444_GS_MCU_SIZE) * 3)
 #define JPEG_422_YCBCR_MCU_SIZE    ((JPEG_444_GS_MCU_SIZE) * 4)
 #define JPEG_420_YCBCR_MCU_SIZE    ((JPEG_444_GS_MCU_SIZE) * 6)
-#define JPEG_MAX_ALLOC_SIZE        (1024UL * 1024UL) // 1 MB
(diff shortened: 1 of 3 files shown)


iabdalkader:
I have to redo this.

kwagyeman:
Okay

iabdalkader:
Have you ever tried to save a 5MP q=100 image to SD? VENC path and/or file_write doesn't services events at all.

iabdalkader:
Other issues, 1MP cap was never going to fix this, there isn't enough memory left for a 5MP frame anyway, with fb buffers allocated. Also, code just fails silently on overflows, so you get half a jpeg + garbage.

PR #3233, opened 2026-08-13
common: Add HaLow (network.HALOW) driver build integration. 

kwagyeman:
This is the top-level PR that is necessary to integrate https://github.com/openmv/micropython/pull/147 and depends on upstream https://github.com/micropython/micropython/pull/19617. As such, it's expected to be in PR review for a while.

### Summary

This PR builds the `network.HALOW` interface — Morse Micro MM6108/MM8108 802.11ah (Wi-Fi HaLow) sub-GHz Wi-Fi — into the OpenMV firmware and enables it on the **N6, AE3 and RT1060**.

What's in this PR:

- **`common/halow.mk`** — links the prebuilt `morselib` and the transceiver firmware blob into the image, and supplies the exception-unwind bounds morselib's libgcc unwinder references.
- **`common/micropy.mk`** — passes `MICROPY_PY_NETWORK_HALOW` through to the MicroPython build.
- **Port wiring** (stm32 / mimxrt / alif) — each port pulls in the `drivers/halow` build rules and services the transceiver from its network poll, with a PendSV dispatch so a queued transmit doesn't wait for the next tick, tearing it down on soft reset.
- **Board enablement** — `MICROPY_PY_NETWORK_HALOW=1` on OPENMV_N6, OPENMV_AE3 (HP core), and OPENMV_RT1060 (which also sets `HALOW_MORSELIB_CORE=arm-cortex-m7f` and trims the DTCM UMA pool by 16K to make room for morselib's DTCM allocation).

HaLow is only compiled when a board opts in; other boards are unaffected. This PR depends on the `lib/micropython` submodule bump that lands the driver (companion PR above) — the submodule pointer is intentionally left out of these commits and will be updated when that merges.

### Testing

Tested on hardware against a HaLow AP — association plus 20-minute TCP/UDP throughput soaks (hundreds of connect/transfer cycles, no leaks or asserts). All three boards build and run.

Throughput (bench soak means, station→AP, **channel 28, 2 MHz bandwidth**):

| Board | MCU | Transport | UDP&nbsp;up | TCP&nbsp;up | TCP&nbsp;down | Assoc |
|---|---|---|---|---|---|---|
| OPENMV_N6 | STM32N6 (Cortex-M55) | hardware IRQ | **8.1 Mbit/s** | 6.0 Mbit/s | 5.0 Mbit/s | 6.3 s |
| OPENMV_RT1060 | i.MX RT1062 (Cortex-M7) | hardware IRQ | **6.1 Mbit/s** | 2.7 Mbit/s | 3.2 Mbit/s | 5.0 s |
| OPENMV_AE3 | Alif Ensemble E3 (Cortex-M55) | polled | **3.2 Mbit/s** | 2.1 Mbit/s | 2.6 Mbit/s | 3.6 s |

UDP-up is the cleanest capability figure (no retransmit/pacing); ~8 Mbit/s on the N6 is close to the practical ceiling for 2 MHz bandwidth. TCP is lower across the board due to link-level loss/retransmit behaviour. (N6 numbers are from the final image; the AE3 and RT1060 figures are from earlier characterization soaks.)

### Notes

- **`morselib` is a prebuilt binary** (Morse Micro Binary Distribution Licence) that ships in the firmware, carried by the `lib/mm-iot-sdk` submodule under `lib/micropython` — the same model as the other vendor Wi-Fi stack we build (cyw43).
- The build linkage is contained in `common/halow.mk` and the per-port hooks; it costs nothing on boards that don't enable HaLow.


kwagyeman:
[halow_firmware.zip](https://github.com/user-attachments/files/31048566/halow_firmware.zip)

Attached are the firmwares for the RT1062, AE3, and N6 based on the master, with all the latest changes for testing.

Example script:

```
# OpenMV Wi-Fi HaLow MJPEG camera stream (Microdot).
#
# Associates with a HaLow (802.11ah) access point, then serves the camera as an
# MJPEG stream to any browser on the HaLow network.  Open http://<printed-ip>/.
#
# One-time setup on the board:
#     mpremote mip install github:miguelgrinberg/microdot
# (or copy the microdot package into /lib).  Set COUNTRY / SSID / KEY below.

import csi
import time
import errno
import network
import asyncio
from microdot import Microdot
from microdot.microdot import MUTED_SOCKET_ERRORS

# This platform reports a client disconnect as EIO (errno 5). Tell microdot to
# treat it as a normal disconnect so closing the browser tab doesn't crash the
# stream task.
if errno.EIO not in MUTED_SOCKET_ERRORS:
    MUTED_SOCKET_ERRORS.append(errno.EIO)

COUNTRY = "US"                 # your regulatory domain, e.g. "US", "DE", "JP"
SSID = "openmv-halow"
KEY = "your-password"

running = True                 # cleared on shutdown so stream tasks stop themselves

# Init camera.
csi0 = csi.CSI()
csi0.reset()
csi0.pixformat(csi.RGB565)
csi0.framesize(csi.VGA)

# Bring up Wi-Fi HaLow.
network.country(COUNTRY)
wlan = network.HALOW()
wlan.active(True)
wlan.connect(SSID, KEY)
while not wlan.isconnected():
    print('Associating with "{:s}"...'.format(SSID))
    time.sleep_ms(1000)
print("HaLow connected", wlan.ifconfig())

app = Microdot()


@app.route("/")
async def index(request):
    return """<!doctype html>
<html>
  <head><title>OpenMV HaLow Stream</title><meta charset="UTF-8"></head>
  <body><h1>OpenMV camera over Wi-Fi HaLow</h1><img src="/video_feed"></body>
</html>""", 200, {"Content-Type": "text/html"}


@app.route("/video_feed")
async def video_feed(request):
    # Class-based async iterator. An `async def ... yield` generator does NOT
    # stream reliably as a Microdot body under MicroPython (it emits one frame
    # then stops), so drive the frames explicitly here.
    class stream:
        def __aiter__(self):
            return self

        async def __anext__(self):
            await asyncio.sleep(0)
            if not running:                 # shutdown requested
                raise StopAsyncIteration
            try:
                frame = csi0.snapshot().to_jpeg(quality=35).bytearray()
            except OSError:
                raise StopAsyncIteration
            return b"--frame\r\nContent-Type: image/jpeg\r\n\r\n" + \
                frame + b"\r\n"

        async def aclose(self):
            print("Stream closed.")

    return stream(), 200, {
        "Content-Type": "multipart/x-mixed-replace; boundary=frame"
    }


async def main():
    global running
    print("Serving at http://%s/" % wlan.ifconfig()[0])
    server = asyncio.create_task(
        app.start_server(host="0.0.0.0", port=80, debug=True)
    )
    try:
        while True:
            await asyncio.sleep_ms(100)     # finite poll timeout -> Ctrl-C works
    finally:
        running = False                     # tell live stream tasks to stop
        app.shutdown()                      # stop accepting new connections
        for _ in range(10):                 # let in-flight tasks exit (~0.5s)
            await asyncio.sleep_ms(50)


try:
    asyncio.run(main())
except KeyboardInterrupt:
    print("Server stopped.")
```

PR #3236, opened 2026-08-16
common/omv_csi: Add manual control of the stats-based AWB gains.

kwagyeman:
The PAG7936 and PS5520 color gains are computed by the ISP from the RGB statistics. Setting the gains manually now writes the statistics they are derived from, reading them back converts the same values to dB, and the statistics start out neutral instead of black until the first update.

Fixes: https://github.com/openmv/openmv/issues/3050


kwagyeman:
Needs to be rebased on https://github.com/openmv/openmv/pull/3220 to be mergeable after that is merged.

PR #3237, opened 2026-08-16
drivers/sensors: Add PAG7936 strobe mode support.

### LINKED ISSUE #3193: Add PAG7936 Trigger Output
@willothewisp0331 and @chobitsfan 

Issue thread to remember to add support for PAG7936-triggered output from the sensor to the camera (in the opposite direction from what triggered mode normally does).

kwagyeman:
Adds an IOCTL that makes the sensor output its exposure strobe on the trigger pin, in the opposite direction of triggered mode. The MCU stops driving the trigger pin while this mode is on so the sensor can drive it, and resumes driving it when the mode is turned off.

NOTE: There's a 1K resistor between the camera and the MCU pin to handle if both are driving at the same time. So, there's no danger of short circuits.

Fixes: https://github.com/openmv/openmv/issues/3193

### DIFF
--- a/common/omv_csi.h
+++ b/common/omv_csi.h
@@ -242,6 +242,8 @@ typedef enum {
     OMV_CSI_IOCTL_GENX320_CALIBRATE       = 0x25,
     OMV_CSI_IOCTL_GENX320_SET_STC         = 0x26,
     OMV_CSI_IOCTL_GENX320_READ_EVENTS_RAW = 0x27,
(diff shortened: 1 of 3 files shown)


iabdalkader:

https://forums.openmv.io/t/pag7936-exposure-strobe-output-on-p10/11848/5

iabdalkader:

https://forums.openmv.io/t/ae3-trigger-mode-problem/11868/3

PR #3240, opened 2026-08-17
drivers/winc1500: Fix WINC1500 socket receive path and performance.

kwagyeman:
## Summary

Receive on the ATWINC1500 boards had one genuinely broken path, one
semantics quirk, and one large performance limiter — root-caused on the
bench with per-call instrumentation:

1. **`recv()` on UDP sockets never returned any data.** The WINC firmware
   delivers datagrams as `SOCKET_MSG_RECVFROM` events, but a plain `recv()`
   request waits for `SOCKET_MSG_RECV` — and the driver's socket callback
   discards event types it is not waiting for, so every `recv()`/`readinto()`
   on a datagram socket timed out empty (`recvfrom()` worked). Datagram
   sockets now route through the recvfrom path, with the request length
   clamped to the maximum UDP payload since the firmware does not deliver
   against oversized requests.

2. **A peer closing a TCP connection surfaced as `OSError(ECONNABORTED)`**
   instead of the 0/EOF every lwIP-based port returns. TCP receive itself
   works — readers that know the transfer length (HTTP with Content-Length,
   the OTA client) never see this — but plain read-until-EOF loops get an
   exception where other ports end cleanly. Now mapped to EOF.

3. **TCP receive ran at ~1.6Mbit/s** because the per-socket receive buffer
   was 1480 bytes and every `recv()` request costs a full HIF round-trip to
   the module (~7-11ms regardless of size), while the module fills however
   much buffer the request offers. The buffer is now board-configurable
   (`OMV_WINC_SOCKBUF_SIZE`), defaulting to 8192 — the measured knee (16K
   is slower again at ~26ms/trip). It is allocated from the MicroPython
   heap per socket, so OPENMV2 and OPENMV3 cap it at 2048: their free heap
   is only ~40-48KB, where an 8K allocation either fails outright or
   starves the application.

## Measured

Python socket benchmarks against a LAN host, 2.4GHz, multiple runs per
board (Mbit/s, TCP up/down then UDP up/down-ceiling):

| Board | Before | After |
|---|---|---|
| H7 Plus | TCP RX ~1.6, UDP RX **zero at any rate** | TCP 9.1 / 5.4, UDP 20.7 / 6.6 |
| H7 | same | TCP 9.3 / 8.1, UDP 21.0 / 6.6 |
| M7 | same | TCP 10.7 / 2.5, UDP 14.3 / 2.8 (2K cap) |
| M4 | same | TCP 10.3 / 1.7, UDP 12.2 / 1.9 (2K cap) |

Transmit was always functional and is unchanged. The M4/M7 receive figures
follow directly from the 2K buffer x per-trip cost — raising them means
spending heap.

## Testing

Validated on all four ATWINC1500-equipped boards (H7 Plus, H7, M7, M4,
module firmware 19.6.x): TCP and UDP benchmarks in both directions, paced
UDP receive tiers from 5 to 30Mbit/s tracking cleanly where they fit under
each board's ceiling, and TCP downloads ending cleanly at server close.
Build-tested: OPENMV2, OPENMV3, OPENMV4, OPENMV4P.


ports/stm32/modules/py_winc.c
```diff
@@ -493,7 +493,30 @@ static mp_uint_t py_winc_socket_send(mod_network_socket_obj_t *socket, const byt
 }
 
 static mp_uint_t py_winc_socket_recv(mod_network_socket_obj_t *socket, byte *buf, mp_uint_t len, int *_errno) {
-    int ret = winc_socket_recv(socket->fileno, buf, len, socket->_private, socket->timeout);
+    int ret;
+    if (socket->type == MOD_NETWORK_SOCK_DGRAM) {
+        // UDP deliveries always arrive as SOCKET_MSG_RECVFROM events; a plain
+        // recv() request waits for SOCKET_MSG_RECV, which never comes, so the
+        // callback discards the data and recv() times out empty.  Route
+        // datagram sockets through the recvfrom path and discard the address.
```
iabdalkader: No need for this verbose AI comments.

boards/OPENMV3/board_config.mk
```diff
@@ -29,3 +29,6 @@ MICROPY_SSL_MBEDTLS = 1
 MICROPY_PY_NETWORK_CYW43 = 0
 MICROPY_PY_BLUETOOTH = 0
 MICROPY_BLUETOOTH_NIMBLE = 0
+
+# Small heap: cap the per-socket WINC receive buffer (default 8192).
+CFLAGS += -DOMV_WINC_SOCKBUF_SIZE=2048
```
iabdalkader: This belongs to `board_config.h` with other WINC options.

drivers/winc1500/include/winc.h
```diff
@@ -35,9 +35,14 @@
...
 // the async call handler will keep overwriting the user-provided buffer in subsequent callbacks. The socket buffer is used
-// as workaround to this issue to allow receiving partial packets. Note the maximum size of WINC internal socket buffer seems
-// to change with host-driver/firmware updates. It's very important to make sure this value is still valid after an update.
-#define WINC_SOCKBUF_MAX_SIZE   (1480)
+// as workaround to this issue to allow receiving partial packets. Sizing also sets throughput: each recv() request costs
+// a full HIF round-trip and the chip delivers up to the request size per trip, so a larger buffer directly multiplies TCP
+// receive throughput (8K measured as the knee; 16K is slower). One buffer is allocated from the MicroPython heap per
+// socket, so small-RAM boards can override this in their board config.
+#ifndef OMV_WINC_SOCKBUF_SIZE
+#define OMV_WINC_SOCKBUF_SIZE   (8192)
```
iabdalkader, line 42: UDP sockets never really used this, preexisting issue, but now you waste 8K for UDP. Perhaps we could allocate conditionally (don't alloc sockbuf for UDP).

ports/stm32/modules/py_winc.c
```diff
@@ -493,7 +493,30 @@ static mp_uint_t py_winc_socket_send(mod_network_socket_obj_t *socket, const byt
...
+    if (ret == SOCK_ERR_CONN_ABORTED) {
+        // The firmware reports the peer closing the connection as an abort;
+        // every lwIP-based port returns 0 (EOF) there, so match that.  The
+        // distinction from a genuine mid-stream reset is not recoverable
+        // from the firmware's error code.
+        py_winc_socket_close(socket);
+        return 0;
+    }
     if (ret <= 0) {
         // NOTE: 0 return from recv() means connection closed.
```
iabdalkader: Comment here needs to be updated.

ports/stm32/modules/py_winc.c
```diff
@@ -493,7 +493,30 @@ static mp_uint_t py_winc_socket_send(mod_network_socket_obj_t *socket, const byt
...
+        // UDP deliveries always arrive as SOCKET_MSG_RECVFROM events; a plain
+        // recv() request waits for SOCKET_MSG_RECV, which never comes, so the
+        // callback discards the data and recv() times out empty.  Route
+        // datagram sockets through the recvfrom path and discard the address.
+        sockaddr addr;
+        // The firmware does not deliver against oversized requests: clamp to
+        // the maximum UDP payload (a datagram fits in one delivery anyway).
+        if (len > 1480) {
+            len = 1480;
+        }
```
iabdalkader: The clamp here should be moved to `winc_socket_recvfrom`, which didn't clamp either.

ports/stm32/modules/py_winc.c
```diff
@@ -493,7 +493,30 @@ static mp_uint_t py_winc_socket_send(mod_network_socket_obj_t *socket, const byt
...
+        ret = winc_socket_recvfrom(socket->fileno, buf, len, &addr, socket->timeout);
+    } else {
+        ret = winc_socket_recv(socket->fileno, buf, len, socket->_private, socket->timeout);
+    }
+    if (ret == SOCK_ERR_CONN_ABORTED) {
+        // The firmware reports the peer closing the connection as an abort;
+        // every lwIP-based port returns 0 (EOF) there, so match that.  The
```
iabdalkader: lwIP returns b"" for every subsequent call and leaves the socket alive until it gets closed explicitly. Here this closes it, so on first call b"" then on second one an exception. Perhaps we should drop the close, if you really want to match the behavior.

iabdalkader (changes requested):
Fixes here are good and appreciated, but what I'd realy like to see is a firmware update. It's been a while since I did that and the updates could have important security fixes. The process is hard, requires a Windows machine to flash firmware and then dump it using some DOS batch file. I wonder if it could be automated.

drivers/winc1500/include/winc.h
kwagyeman: Only stream sockets allocate it now, datagram sockets get NULL, and close() guards the free.

ports/stm32/modules/py_winc.c
kwagyeman: Dropping the close alone means every following recv() does another HIF round trip on a dead socket. Latched the EOF in the sockbuf instead, so it returns b"" immediately after the first one and the socket stays alive until you close it.

kwagyeman:
All changes done. Making a new PR for the firmware update.

kwagyeman:
Firmware: https://github.com/openmv/openmv/pull/3270

iabdalkader (approved):

PR #3245, opened 2026-08-21
lib/imlib: Fix JPEG software and hardward paths issues.

iabdalkader:
The bug that started this: a 5 MP capture at quality 100 saved a 2.2 MB file with no EOI marker and a blank lower half. On overflow the encoders returned "failed" but left `dst->size` at the allocation size, and `jpeg_write` ignored the return and wrote the whole buffer. All `jpeg_compress` implementations now share one contract, matching `png_compress`:

```
true  == failed  == nothing to clean up, our buffer is freed
false == success == the caller owns a valid buffer
```

### JPEG encoders fixes

- Enable the software encoder's realloc path for buffers it owns. It existed, but every caller disabled it, so overflow was always fatal.
- Grow by half the buffer size instead of a fixed 1 KB, which cost one realloc and one full copy per 1 KB encoded, and keep the `UMA_CACHE` flag across the realloc.
- Free the output buffer on failure and trim it to the encoded size on success, in all three encoders.
- Raise in `jpeg_write` and `mjpeg_write` instead of writing the buffer regardless.
- Replace VC8000's `-1/0/1` return convention, with named error codes, to distinguish buffer overflows from hardware faults.
- Move the `is_compressed` check above the allocation, so that path no longer allocates a buffer only to fail.

### Save-path validation

`.jpg` and `.png` were the only extensions with no format check.

- Saving a **PNG as `.jpg`** reached `jpeg_compress` with a compressed source and wrote up to **1 MB of uninitialised heap to disk**. Now raises.
- Saving a **JPEG as `.png`** wrote an empty file. Now raises.
- An image already in the target format is still copied as-is.
- Replace the per-character extension matching with a table and a suffix match.
- Make the format switch exhaustive, so build fails if a new format is added without validation.
- Check `png_compress`'s return value in `png_write`, and compress before creating the file so a failure doesn't leave an empty one behind.

### VC8000 subsampling

- The driver is built without `JPEGENC_422_MODE_SUPPORTED`, so it forces 4:2:0 and ignores `cfg.codingMode` entirely; an explicit 4:2:2 request was silently honoured as 4:2:0.
- Report anything but 4:2:0 as unsupported and fall back to the HAL codec, as 4:4:4 already did.

### Follow-ups, not in this PR
  - The buffer sizing question from #3231.
  - Uncovered an unrelated USB/protocol/TinyUSB issue.

### DIFF
--- a/lib/imlib/framebuffer.c
+++ b/lib/imlib/framebuffer.c
@@ -343,7 +343,7 @@ void framebuffer_update_preview(image_t *src) {
 
     // Compress the frame if the raw image didn't fit or the format is non-mutable.
     if (!raw_stream) {
-        overflow = jpeg_compress(src, &dst, fb->quality, false, JPEG_SUBSAMPLING_AUTO);
+        overflow = jpeg_compress(src, &dst, fb->quality, JPEG_SUBSAMPLING_AUTO);
     }
 
     if (overflow) {
--- a/lib/imlib/imlib.c
+++ b/lib/imlib/imlib.c
@@ -594,68 +594,88 @@ uint16_t imlib_yuv_to_rgb(uint8_t y, int8_t u, int8_t v) {
 ////////////////////////////////////////////////////////////////////////////////
 
 #if defined(IMLIB_ENABLE_IMAGE_FILE_IO)
-static save_image_format_t imblib_parse_extension(image_t *img, const char *path) {
-    size_t l = strlen(path);
-    const char *p = path + l;
-    if (l >= 5) {
-        if (((p[-1] == 'g') || (p[-1] == 'G'))
-            && ((p[-2] == 'e') || (p[-2] == 'E'))
-            && ((p[-3] == 'p') || (p[-3] == 'P'))
-            && ((p[-4] == 'j') || (p[-4] == 'J'))
-            && ((p[-5] == '.') || (p[-5] == '.'))) {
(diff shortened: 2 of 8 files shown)


iabdalkader:
This partially addresses the JPEG size issue in the software path, as it now can grow and trim the buffer, so we could start with any size.

kwagyeman (approved):
Was concerned at first when I saw the is_compressed() check being move to the start of a function and the quick return, but, then noticed the contract change and from there everything tracks.

Read all the files multiple times and didn't see any issues.

iabdalkader:
I see a couple of issues here, so far. They're not too critical, but should fix them while I'm at it:

1- The software encoder reallocates buffer if it owns it || realloc arg is true. The hardware path ignores `realloc` completely and only reallocs owned buffers. Of course hw path can't grow the buffer, but it can at least trim it.
2- These prototypes are really old, they returned `bool` for some reason, while everything else returns -1/0. We should align them with the rest of the API.

iabdalkader:
@kwagyeman I think we should drop the `realloc` arg too. Nothing calls with a buffer and `realloc=true`. It's also dangerous if caller passes GC heap pointer.

kwagyeman:
@iabdalkader - I don't have any issue with that. I don't use it for the hardware stuff. Please make all changes you need and then merge. I'd like to get the v5.0.1 release out before the end of the month.

PR #3246, opened 2026-08-21
modules/py_csi_ng: Dispatch frame/vsync callbacks via mp_irq_dispatch.

kwagyeman:
The frame and vsync callbacks ran Python in IRQ context via a bare mp_call_function call with no NLR handler, so any exception raised by user code (e.g. an IndexError) had nowhere to unwind to and hard-hung the device with no traceback and dead USB.

Dispatch them through mp_irq_dispatch like pin IRQ callbacks: ISR stack limit, scheduler and GC locked, and any exception is printed and the callback disabled instead of hanging the device. A C trampoline strips the argument mp_irq_dispatch passes so the frame_callback signature is unchanged.

### DIFF
--- a/modules/py_csi_ng.c
+++ b/modules/py_csi_ng.c
@@ -34,6 +34,7 @@
 #include <stdio.h>
 #include "py/mphal.h"
 #include "py/runtime.h"
+#include "shared/runtime/mpirq.h"
 
 #if MICROPY_PY_CSI_NG
 
@@ -831,7 +832,12 @@ static void omv_csi_vsync_callback(void *data) {
         #ifdef OMV_CSI_VSYNC_PIN
         vsync_state = omv_gpio_read(OMV_CSI_VSYNC_PIN);
         #endif
-        mp_call_function_1(self->vsync_cb, mp_obj_new_int(vsync_state));
+        // Dispatch as a hard IRQ so an exception cannot propagate into the IRQ handler.
+        if (mp_irq_dispatch(self->vsync_cb, mp_obj_new_int(vsync_state), true) < 0) {
+            // Uncaught exception; disable the callback so that it doesn't run again.
+            self->vsync_cb = mp_const_none;
+            omv_csi_set_vsync_callback(self->csi, (omv_csi_cb_t) { NULL, NULL });
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


modules/py_csi_ng.c
```diff
@@ -860,11 +866,24 @@ static mp_obj_t py_csi_vsync_callback(size_t n_args, const mp_obj_t *args) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(py_csi_vsync_callback_obj, 1, 2, py_csi_vsync_callback);
 
+// Strips the argument mp_irq_dispatch passes so frame_cb keeps its zero-arg signature.
+static mp_obj_t py_csi_frame_cb_trampoline(mp_obj_t self_in) {
+    py_csi_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    return mp_call_function_0(self->frame_cb);
```
iabdalkader, line 872: Maybe make this accept an arg too, to match the other callback and reserve it for the future. But this is a breaking API change, so we'll need to bump major.

kwagyeman: Here's some research from Claude on what things we could pass as args...

---

Current contract:
- `frame_callback` has been zero-arg since 2021 (`94462ef74`). v5.0.0 carried it over unchanged, with a `set_frame_callback` → `frame_callback` migration row and `Callable[[], None]` in `csi.pyi`.
- `scripts/libraries/sensor.py` forwards `set_frame_callback` to it, so a signature change breaks both APIs.
- No in-tree script uses `frame_callback` (four use `vsync_callback`). Blast radius is external users only.
- There is no clean way to accept both shapes in IRQ context, so it is a hard break.
- The arg version is the smaller diff: `mp_irq_dispatch` already delivers one positional arg. The trampoline (the whole +88 bytes) exists only to preserve the zero-arg shape.

Candidates for what the arg would carry:

| Candidate | Assessment |
|---|---|
| The `CSI` object | MicroPython convention: `Pin.irq`, `Timer`, `UART.irq`, `Counter`, `Encoder` all pass the object. STM32 HAL does the same (`HAL_DCMI_FrameEventCallback(hdcmi)`). Addresses an existing gap: `OMV_CSI_MAX_DEVICES` is 3 on N6 and 2 on RT1060/PT, and one handler cannot tell which sensor fired. Any later data becomes a method on the object with no further break. |
| Status/flags int | A real signal exists: in double-buffer mode `framebuffer_release` discards the frame (returns NULL) but the callback still fires; `VB_FLAG_OVERFLOW` exists on stm32/mimxrt. Needs plumbing through `omv_csi_cb_t` (`void *` only today), and a second field breaks again unless it is a bitmask. |
| Timestamp | Nothing captures a per-frame timestamp in the ISR; `last_frame_ms` only updates in the framerate limiter (`omv_csi.c:907-915`). The callback can read `ticks_us()` itself with µs jitter. |
| Sequence / drop count | Not tracked anywhere today; would need a counter added to `omv_csi_t`. |
| Image object | Not possible: GC is locked under `mp_irq_dispatch`, and the frame is consumed by `snapshot()`. |

If the signature changes, the candidate with the most value is the `CSI` object, `cb(csi)`, and nothing else. Under `mp_irq_dispatch(..., true)` the callback runs synchronously inside the ISR, so any state read through `csi` inside the callback is the state of this exact frame. That makes every other candidate a method rather than an argument:
- Sequence / drop count: a counter in `omv_csi_t` read inside the callback is exact, since no other frame can complete before the callback returns.
- Timestamp: `time.ticks_us()` inside the callback is within a few µs of the DMA interrupt.
- Status: the double-buffer discard is set before the callback fires, so it can be read from the object too. It only occurs with 2 buffers and a slow consumer; 3 buffers or FIFO mode is the fix for a user who cares.
A second positional arg would reintroduce the problem under discussion: the next datum needs another break, a method never does.

Implementation: `mp_irq_dispatch(self->frame_cb, MP_OBJ_FROM_PTR(self), true)`, trampoline removed, plus docs, stubs, and the `sensor.py` shim.

Methods to add: none required in the first cut. The object already gives identity (`cid()`, `devices()`), `readable()`, `width()`/`height()`/`pixformat()`. Two are cheap if wanted, with real signals at the ISR sites:
- `frame_count()`: uint32 in `omv_csi_t`, incremented in each port's frame-event handler before the callback.
- a drop counter: incremented at the three existing drop sites (framerate limiter `drop_frame`, double-buffer discard in `framebuffer_release`, `VB_FLAG_OVERFLOW` on stm32/mimxrt).
A capture timestamp is a separate design: with multiple buffers it must live per `vbuffer_t` and surface on the image `snapshot()` returns, not on the CSI object.

`vsync_callback`: no change recommended.
- It already carries the event datum, which is OpenMV's existing convention: audio passes the PCM buffer, ft5x06 the touch point, tfp410 the connected bool, CEC the source address and frame. `frame_callback` was zero-arg only because it has no datum.
- The state cannot be recovered from `csi` after the fact: there is no vsync accessor, and vsync is stm32-only (registered directly on the EXTI pin at `ports/stm32/omv_csi.c:400`; other ports take the weak default and never fire it). A `vsync()` method would exist only to move an int from an argument to a call.
- It is the only one of the two with in-tree users: four examples and the docs use `lambda state, led=led: led.value(state)`.
If MicroPython symmetry is preferred over the datum convention, the consistent form is `cb(csi, state)`, and it should ride the same major bump as `frame_callback` so it is one break, not two.

---

So, if we change it, we should probably add the above so it's maximally useful. Otherwise, I would leave it unchanged.

PR #3254, opened 2026-08-30
lib/imlib: Drop bayer/YUV support from gif_add_frame().

### LINKED ISSUE #3250: out-of-bounds stack read in gif_add_frame
**Describe the bug**

Adding certain Bayer or YUV images to a GIF causes an out-of-bounds stack read in `gif_add_frame()`.

For the final GIF block, the function allocates:

```c
uint16_t pixels[block_size];
```

but later unconditionally reads:

```c
uint16_t pixel = pixels[2];
```

If the final block contains only one or two pixels, `pixels[2]` is outside the VLA.

**To Reproduce**

A minimal Bayer example is an `11 x 23` image. It contains 253 pixels:

```text
253 = (2 * 126) + 1
```

so the final GIF block contains one pixel.

```python
import image
import gif

img = image.Image(11, 23, image.BAYER)

g = gif.Gif("test.gif", width=11, height=23, color=True)
g.add_frame(img)
g.close()
```

The affected code in `lib/imlib/gif.c` is:

```c
int block_size = IM_MIN(BLOCK_SIZE, bytes - (y * BLOCK_SIZE));

uint16_t pixels[block_size];

if (img->is_bayer) {
    imlib_debayer_line(0, block_size, y, pixels, PIXFORMAT_RGB565, img);
} else {
    imlib_deyuv_line(0, block_size, y, pixels, PIXFORMAT_RGB565, img);
}

for (int x = 0; x < block_size; x++) {
    uint16_t pixel = pixels[2];
    ...
}
```

For the final block in the example:

```text
block_size = 1
pixels     = uint16_t[1]
access     = pixels[2]
```

resulting in an out-of-bounds stack read.

Any Bayer/YUV image where:

```text
(width * height) % 126 == 1
```

or:

```text
(width * height) % 126 == 2
```

can reach the same condition.

**Expected behavior**

`Gif.add_frame()` should encode valid Bayer/YUV images without accessing memory outside the temporary pixel buffer.

**Screenshots**

Not applicable.

* IDE version: Not applicable
* Firmware version: Current `master` / head revision as of 2026-08-28
* Board: Not hardware-specific
* OS: Not applicable

iabdalkader:
py_gif_add_frame() only accepts mutable images, so this branch was never reachable. Converting line by line here would also be much slower than setting the format to RGB565 in the first place.

Fixes #3250.

### DIFF
--- a/lib/imlib/gif.c
+++ b/lib/imlib/gif.c
@@ -101,26 +101,6 @@ void gif_add_frame(file_t *fp, image_t *img, uint16_t delay) {
             }
             file_write(fp, blk_buf, 2 + block_size);
         }
-    } else if (img->is_bayer || img->is_yuv) {
-        for (int y = 0; y < blocks; y++) {
-            int block_size = IM_MIN(BLOCK_SIZE, bytes - (y * BLOCK_SIZE));
-            blk_buf[0] = 1 + block_size;
-            blk_buf[1] = 0x80; // clear code
-            uint16_t pixels[block_size];
-            if (img->is_bayer) {
-                imlib_debayer_line(0, block_size, y, pixels, PIXFORMAT_RGB565, img);
-            } else {
-                imlib_deyuv_line(0, block_size, y, pixels, PIXFORMAT_RGB565, img);
-            }
-            for (int x = 0; x < block_size; x++) {
-                uint16_t pixel = pixels[2];
-                uint16_t r = COLOR_RGB565_TO_R5(pixel) >> 3;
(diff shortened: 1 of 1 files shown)


iabdalkader:
Note: Code can be fixed, but not sure if it's worth it.

kwagyeman (approved):
Yeah, dropping it is fine. Realastically, the old gif module should be removed at some point once all the h.264 stuff is done. I have some good python only avi code that replaces it and mjpeg really and would have audio recording support.

iabdalkader:
Maybe move to an example. Never should have been built-in really.

PR #3258, opened 2026-09-02
modules/py_csi_ng: Export custom resolution symbol.

### LINKED ISSUE #2950: Missing way to get custom frame size resolution passed.

iabdalkader:
Getting the custom resolution back already works with `csi.width()` and `csi.height()`, there was just no export to compare against. Additionally, fix setting custom framesize twice.

Fixes: https://github.com/openmv/openmv/issues/2950


kwagyeman:
Works for me!

kwagyeman (approved):

PR #3259, opened 2026-09-03
Prod

amrdeepsaini:


iabdalkader:
@amrdeepsaini 

One of those files contains private keys plus your fleet's device UIDs and node addresses. That's on a public repo, and since your uploaded images are encrypted using the same keys, anyone with this diff can decrypt them and impersonate your nodes. Please rotate all of those keys.  Also worth mentioning your upload endpoint (api.vyomiq.io/watchmen-detect/) is called with no auth header in the firmware. Anyone can post events or images into it with a made-up machine_id.

PR #3269, opened 2026-09-15
drivers/winc1500: Update host driver to Microchip's 19.7.x driver.

kwagyeman:
Opening this for the record and then closing it. It works, it is fully bench-verified, and it should not be merged — measurements say the driver swap costs us performance for no functional gain. The firmware half goes in separately as a 3-file PR.

## What this does

Replaces the in-tree ASF 19.6.1 host driver with Microchip's Harmony driver (wireless_wifi, driver/winc/winc1500_19.7.7), which their release notes pair with firmware 19.7.7/19.7.10/19.7.11, and adds firmware 19.7.11.

Vendor sources are unmodified where possible. The things Harmony supplies through its own framework are shimmed instead: OSAL (mutexes and semaphores are no-ops — the driver is only ever entered from the MicroPython context, the WINC interrupt is deferred to PendSV which cannot preempt a driver call, so there is nothing to serialize and blocking there would deadlock), SPI, GPIO and debug.

Three vendor edits were unavoidable and are marked in place. The interesting one is nmbus.c: Harmony ships static nm_bus_init/nm_bus_deinit stubs there because its WDRV layer owns the SPI. Left alone they shadow our bus wrapper, gc-sections drops the real one, the SPI peripheral is never initialized, and the board resets with no output at all. The tell is in the symbol table — a 16-byte local `t nm_bus_init` instead of our global `T`.

API deltas on our side: close() is now shutdown(), m2m_wifi_handle_events() takes no argument, the wifi callback payload is const, gethostbyname() takes a char *, and the ASF sint8/uint8 typedefs are gone in favour of stdint.

## Why it isn't worth merging

Measured on M4, M7, H7 and H7 Plus with ATWINC1500 shields, using 512KB TCP transfers and 800 x 1024-byte UDP datagrams. Medians in Mbit/s, on the H7:

| Config | TCP down | TCP up | UDP down | UDP up |
| --- | --- | --- | --- | --- |
| master driver + 19.7.11 | 7.53 | 4.68 | 5.88 | 15.51 |
| Harmony driver + 19.7.11 | 5.48 | 4.78 | 3.81 | 15.64 |

That pattern holds across boards: the Harmony driver costs roughly 22-30% on TCP receive and 25-36% on UDP receive. Transmit is unaffected in both directions.

I isolated it by running the old driver against the new firmware, which is legal — the image declares a minimum required driver of 19.3.0 and the driver enforces that gate. The firmware is not the problem; the driver is. Suspected cause: Harmony's nmspi.c splits every bus transaction into separate write-then-read CS frames where the ASF driver used one combined transfer.

Since the existing driver runs 19.7.11 unchanged and faster, there is no reason to take the 55-file vendor drop.

## Testing

Everything below was run on real hardware before deciding to close this.

- Scan, associate, DHCP and TCP/UDP transfer on all four boards.
- 30-transfer soak per board on the Harmony driver: 120/120, no failures.
- No leak: on the H7 Plus, free heap after gc.collect() is -48 bytes from baseline and flat across 40 transfers.
- A soak including disconnect() + connect() scores 17/21 on every board, but master scores exactly 17/21 too, on the same iterations with the same byte counts. Re-association after an explicit disconnect is a pre-existing bug, not from this change, and is worth its own issue.

Superseded by the firmware-only PR.


kwagyeman:
Replaced by https://github.com/openmv/openmv/pull/3270

PR #3271, opened 2026-09-16
drivers/winc1500: Fix reconnecting after disconnect().

kwagyeman:
`disconnect()` followed by `connect()` left the interface down — `isconnected()` stayed false and every socket call afterwards failed with EAGAIN, even though `connect()` had returned success.

`disconnect()` asks the chip to disconnect but never pumps the event handler, so the resulting `M2M_WIFI_RESP_CON_STATE_CHANGED` is still queued when the next `connect()` runs. `connect()` waited on the generic `async_request_done` flag, and the handler sets that flag for the disconnected state as well — so the stale event from the *previous* connection ended the wait immediately, and `connect()` returned success having just cleared `wlan_connected` and `ip_obtained`.

The same confusion made a failed connection look like a successful one: an association that is rejected (a wrong key, say) reports the same state change as a disconnection, so `connect()` returned 0 for that too.

`connect()` now drains anything left over before it starts, then waits on the connection state itself rather than on any event arriving. Since a rejected association never reaches that state, the wait is bounded by a timeout and reports failure to the caller, which is what the Python layer already expects. `disconnect()` also handles its own event now.

## Testing

OpenMV M4, M7, H7 and H7 Plus with ATWINC1500 shields. A soak of 21 transfers with a disconnect/reconnect every seventh iteration:

| Board | Before | After | Run time before | after |
| --- | --- | --- | --- | --- |
| M4 | 17/21 | 21/21 | 178s | 39s |
| M7 | 17/21 | 21/21 | 177s | 21s |
| H7 | 17/21 | 21/21 | 172s | 10s |
| H7 Plus | 17/21 | 21/21 | 171s | 13s |

The time drop is because it no longer waits out a 40s timeout on every reconnect. Connecting with a wrong key now raises after the timeout instead of returning success, and a correct key still connects in ~230ms immediately afterwards.


drivers/winc1500/src/winc.c
```diff
@@ -576,11 +583,18 @@ int winc_connect(const char *ssid, uint8_t security, const char *key, uint16_t c
...
+    // Wait for the connection to come all the way up. Note this waits on the
+    // connection state, not on any event arriving: a failed connection reports
+    // the same state change as a disconnection, and would otherwise be reported
+    // to the caller as success.
+    uint32_t tick_start = HAL_GetTick();
+    while (winc_isconnected() == 0) {
         __WFI();
         // Handle pending events from network controller.
         m2m_wifi_handle_events(NULL);
+        if ((HAL_GetTick() - tick_start) >= WINC_CONNECT_TIMEOUT) {
```
iabdalkader: Why not use MP ticks functions and event handling instead of WFI?

iabdalkader (approved):

drivers/winc1500/src/winc.c
kwagyeman: It was just the AI trying not to pull in the MP headers. Cool.

PR #3274, opened 2026-09-20
modules/py_tof: Make VL53LX init and ranging more robust.

kwagyeman:
- Gate init on the device ID: poll is_alive for up to 100 ms after the power-cycle instead of ignoring the result and pushing the firmware into a sensor that has not come up yet.
- Power-cycle the sensor before the one retry, instead of only pulsing SCL, so a half-loaded sensor starts from a clean state.
- Report which init step failed in the RuntimeError message.
- Raise the default read_depth/snapshot timeout from 100 ms to 500 ms. At 15 Hz a corrupted-frame retry needs up to ~140 ms, and frames measured up to 104 ms apart on the bench, so 100 ms could time out on a healthy sensor and trigger a 2.5 s recover for nothing.

Bench (OPENMV_AE3): 10/10 init cycles, 906 reads/60 s with 0 failures, re-init over a ranging sensor 8/8, power cut mid-stream raises once and the next read succeeds.


modules/py_tof.c
```diff
@@ -107,36 +107,57 @@ static void tof_fill_image_float_obj(image_t *img, mp_obj_t *data, float min, fl
...
+    // Load the firmware and initialize the sensor.
+    if (vl53lx_init(vl53lx_dev) != 0) {
+        return 2;
+    }
 
     // Set resolution (number of zones).
     // NOTE: This function must be called before updating the ranging frequency.
-    error |= vl53lx_set_resolution(vl53lx_dev, VL53LX_RESOLUTION_8X8);
+    if (vl53lx_set_resolution(vl53lx_dev, VL53LX_RESOLUTION_8X8) != 0) {
+        return 3;
```
iabdalkader: What difference does it make which step failed? I don't think this is useful.

iabdalkader (changes requested):
So tests say ~104ms? Why 500ms delay? Recovery should be fast.
Also, perhaps there's room for improvement here, we could retry GO2 more times (2-3) with a backoff/random delay to get the driver/sensor back in sync. Or (I haven't checked) but there might be a flag or something that says sensor is writing/busy.

modules/py_tof.c
kwagyeman: Dropped the number.

kwagyeman:
Had Claude take in your comments from the PR and the forum post. Here's what it did:

> Dropped the step number. On the timeout: the 104 ms was a plain wait for a frame, not a retry. A 3 minute histogram on the bench AE3 shows 2718/2719 reads under 70 ms and one between 100 and 110 ms, so a 100 ms default trips on a healthy sensor about once every 3 minutes and pays a 2.5 s firmware reload each time. I set it to 200 ms (two 15 Hz periods plus margin) and made it bound each wait for a frame rather than the whole read, so a stuck sensor is still caught in 200 ms.

> Corrupted frames now retry up to three times. No backoff needed: after a corrupted read the next data-ready is a fresh update, so the retry read starts at the beginning of the period and cannot overlap the sensor's write. GO2 errors are still polled through within the timeout. I checked the ULD for a busy flag: the 4 byte header only exposes data-ready (byte 3 bit 4) and GO2 error (byte 3 bit 7), so the header/footer id check is the only overlap detection available.

> Bench: 10/10 init cycles, 2719 reads over 180 s with 0 failures, re-init over a ranging sensor 8/8, power cut mid-stream raises once and the next read succeeds.

modules/py_tof.c
```diff
@@ -149,25 +169,32 @@ static int tof_vl53lx_recover(vl53lx_dev_t *vl53lx_dev) {
 }
 
 static void tof_vl53lx_get_depth(vl53lx_dev_t *vl53lx_dev, float *frame, int timeout) {
-    uint8_t frame_ready = 0;
     // Note depending on the config in platform.h, this struct can be too big to alloc on the stack.
     vl53lx_data_t ranging_data;
 
-    for (mp_uint_t start = mp_hal_ticks_ms(); ; mp_hal_delay_ms(1)) {
-        uint8_t status = vl53lx_check_data_ready(vl53lx_dev, &frame_ready);
+    // The sensor's output buffer is not double-buffered, so a read that overlaps the sensor's
```
iabdalkader: We could avoid more nesting here, and minimize the diff, with something like:
```diff
@@ -169,7 +169,10 @@ static void tof_vl53lx_get_depth(vl53lx_dev_t *vl53lx_dev, float *frame, int tim
         // - GO2 errors (other non-zero status) are transient, keep polling.
         if ((timeout > 0) && (mp_hal_ticks_ms() - start) >= timeout) {
             tof_vl53lx_recover(vl53lx_dev);
-            mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("VL53LX ranging failed"));
+            if (--retry == 0) {
+                mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("VL53LX ranging failed"));
+            }
+            start = mp_hal_ticks_ms();
         }
     }
```

iabdalkader (approved):

PR #3276, opened 2026-09-20
modules/py_ml: Allocate NPU models from external memory.

### LINKED ISSUE #3272: ml.Model() fails with "Failed to load network" for any model under ~1 MB loaded from a file (image allocated in the SRAM2 GC block)
**Describe the bug**
On the OpenMV N6, `ml.Model(path)` fails for a file-backed model whenever the model is small enough to be allocated in the first GC block (SRAM2, 1 MB). The identical bytes load fine from ROMFS (zero-copy path) and load fine from the same file once SRAM2 is already occupied, so the failure is purely about where `py_ml.c` places the image. Every model in the stock ROMFS except `palm_detection_full_192` and `hand_landmarks_full_224` (the only two over 1 MB) is affected when loaded from `/flash` or an SD card.

It has stayed hidden because the bundled models take the `mp_get_buffer()` zero-copy path and never touch the allocator. It appears as soon as a model is loaded from the filesystem — e.g. after recompiling one with `stedgeai` and copying it to the SD card — and the error message gives no hint of the cause.

**To Reproduce**
Steps to reproduce the behavior — on a stock N6, nothing needed but `/rom` and `/flash`:

1. Copy the board's own `/rom/fomo_face_detection.tflite` to `/flash` byte-for-byte.
2. `ml.Model("/rom/fomo_face_detection.tflite")` → loads (`model_addr 0x7080…`, ROMFS zero-copy).
3. `ml.Model("/flash/repro.tflite")` → **`RuntimeError: Failed to load network`**.
4. Allocate 1 MB of ballast (64 × 16 KB bytearrays) so SRAM2 is full, then `ml.Model("/flash/repro.tflite")` again → loads (`model_addr 0x903…`, DRAM).

Script that does exactly this (run from the IDE or over the REPL):

```python
import gc, os, ml, uctypes

ROM = "/rom/fomo_face_detection.tflite"
TMP = "/flash/repro.tflite"

# Same bytes, different filesystem: copy the ROMFS model to /flash.
with open(ROM, "rb") as s, open(TMP, "wb") as d:
    while True:
        b = s.read(4096)
        if not b:
            break
        d.write(b)

gc.collect()
a = bytearray(64 * 1024);       print("64 KB allocation lands at 0x%08X" % uctypes.addressof(a)); del a
a = bytearray(2 * 1024 * 1024); print(" 2 MB allocation lands at 0x%08X" % uctypes.addressof(a)); del a
gc.collect()

def attempt(tag, path, fill_sram2=False):
    # 64 x 16 KB = 1 MB, the size of GC block 0, so the model has to go elsewhere.
    hold = [bytearray(16 * 1024) for _ in range(64)] if fill_sram2 else []
    try:
        m = ml.Model(path)
        print("%-40s OK   %s" % (tag, str(m).split(",")[1].strip()))
        del m
    except Exception as e:
        print("%-40s FAIL %s" % (tag, e))
    del hold
    gc.collect()

attempt("1. from /rom (zero-copy)", ROM)
attempt("2. from /flash, heap untouched", TMP)
attempt("3. from /flash, SRAM2 filled first", TMP, fill_sram2=True)
os.remove(TMP)
```

Output, verbatim, from that script on the board described below:

```
64 KB allocation lands at 0x3411B580        <- SRAM2 (GC block 0)
 2 MB allocation lands at 0x9018CB80        <- DRAM  (GC block 1)
1. from /rom (zero-copy)                 OK   model_addr: 0x70800040
2. from /flash, heap untouched           FAIL Failed to load network
3. from /flash, SRAM2 filled first       OK   model_addr: 0x9039cb80
```

While the ballast is held, a fresh 64 KB allocation lands at `0x9019…`, i.e. the allocator has moved to DRAM — and the same file then loads.

**Expected behavior**
Step 3 loads. A model that installs from ROMFS and installs from DRAM should install from wherever `ml.Model()` puts it — or the load path should put it somewhere it can be installed from.

**Screenshots**
Not applicable — terminal output above; complete logs of two independent runs attached (`sram2-repro-n6-2026-09-19.txt`).

 - IDE version: 5.0.0
 - Firmware version: 5.0.0 (`omv.version_string()`), MicroPython `v1.28.0-49 on 2026-07-02`, build `OPENMV_N6`; board freshly re-initialised from the IDE, which reports it as latest; stock ROMFS
 - Board: OpenMV N6 (STM32N657X0)
 - OS: macOS 26.6.2 (IDE); the repro above was driven from Linux (Debian 12, Raspberry Pi 5) over the serial REPL with no SD card inserted and the host mass-storage driver detached. The failure is on the board and independent of the host.

---

**Analysis**

`boards/OPENMV_N6/board_config.h`:

```c
#define OMV_GC_BLOCK0_MEMORY  SRAM2   // Main GC block
#define OMV_GC_BLOCK0_SIZE    (1M)
#define OMV_GC_BLOCK1_MEMORY  DRAM    // Main GC block
#define OMV_GC_BLOCK1_SIZE    (24M)
```

MicroPython fills GC blocks in order, so any allocation that fits goes to SRAM2 first. `modules/py_ml.c` reads a file-backed model with:

```c
// Try allocating model using GC memory (typically fast SRAM).
model->_raw = m_malloc_maybe(size + IMLIB_ML_MODEL_ALIGN - 1);
```

so a model under ~1 MB lands in SRAM2. `lib/stai/stai_backend.c` then hands that address to `ll_aton_reloc_install()` with `.mode = AI_RELOC_RT_LOAD_MODE_XIP`, which fails for an image in SRAM2 (the region does not appear to be usable by the relocatable runtime / NPU for in-place execution), and the failure surfaces as the generic `RuntimeError("Failed to load network")`. Models larger than the SRAM2 block cannot be placed there and spill to DRAM, which is why the two big MediaPipe models work from a file and every smaller one does not. Alignment is not the cause — the load path already aligns to `IMLIB_ML_MODEL_ALIGN`, and shifting the heap by small amounts does not change the outcome; only leaving SRAM2 does.

The ROMFS path (`mp_get_buffer()` succeeds → `model->data = bufinfo.buf`) never allocates, which is why the bundled models are unaffected.

**Suggested fix**

Allocate file-backed NPU models somewhere the runtime can install from rather than "typically fast SRAM" — e.g. `uma_malign(size, IMLIB_ML_MODEL_ALIGN, UMA_PERSIST)` against the DRAM UMA pool, or fall back to that when the GC allocation comes back inside SRAM2. Separately, surfacing the `ll_aton_reloc_install()` return code instead of a bare "Failed to load network" would have made this a one-minute diagnosis.

**Notes**

- In one run, after step 2 failed, step 3 **hung inside `ml.Model()`** (no exception, board unresponsive) instead of completing; in a fresh session it passes as shown. Observed once — mentioned in case a failed install leaves state behind.
- `ll_aton_reloc_network.c` `_ai_reloc_rt_checking()` tests the runtime version with `&&` (`major != X && minor != Y`), so a mismatch in only one of the two passes the check. Unrelated to this failure; noticed while reading the path.
- Also reproduced on a third-party STM32N657 board with the same GC block layout (fomo 64 KB, blazeface 220 KB, yolo_lc 318 KB all fail from SD; palm 1.47 MB and hand_landmarks 3.32 MB load), so it is not one unit.

[sram2-repro-n6-2026-09-19.txt](https://github.com/user-attachments/files/32414970/sram2-repro-n6-2026-09-19.txt)

kwagyeman:
On the N6, a model loaded from a file is read into GC memory, which places anything under 1MB in SRAM2. The ST Neural ART code generator marks the weights pool cacheable, and the NPU only supports cacheable accesses to external memory, so `ll_aton_reloc_install()` rejects the image with `AI_RELOC_RT_ERR_PARAM_ADDR` (-7). Models larger than SRAM2 spill into SDRAM and load, which is why only the small ones fail.

Allocate file-backed models from the default (external RAM) UMA pool on STAI boards and include the runtime's return code in the error message.

The runtime executes the image in place and the allocator hands back the same address for every model, so before installing:

- clean the MCU DCache over the image, otherwise the NPU reads the previous model's bytes still sitting in dirty lines (wrong outputs, then a crash);
- invalidate the ICache over the image, otherwise the second distinct model loaded from a file executes the previous model's stale code and hard faults.

Bench (OPENMV_N6): the reporter's three-step repro passes; five models (fomo, blazeface, yolo_lc, person_detect, face_landmarks) loaded from `/flash` back to back at the same address produce byte-identical outputs to their ROMFS copies over three rounds; 40 load/free cycles and all eight ROMFS models (220 KB to 3.3 MB) load from a file. Removing either cache operation was tested and each one takes the board down. An NPU cache clean/invalidate over the image was also tested and made no difference, so it is not included.

ROMFS remains the intended place for N6 models: weights stream from the XSPI2 flash instead of sharing the PSRAM with the frame buffers.

Fixes #3272.


lib/stai/stai_backend.c
```diff
@@ -161,11 +161,19 @@ int ml_backend_init_model(py_ml_model_obj_t *model) {
         .mode = AI_RELOC_RT_LOAD_MODE_XIP,
     };
 
+    // The model's image may have just been written to RAM by the CPU, at an
```
iabdalkader, line 164: > The model's image may have just been written to RAM by the CPU

It's way more common for models to be loaded from romfs, rather than file-backed, so this will unconditionally clean/invalidate MBs of flash/romfs addresses.

modules/py_ml.c
```diff
@@ -365,6 +365,14 @@ mp_obj_t py_ml_model_make_new(const mp_obj_type_t *type, size_t n_args, size_t n
         }
 
         model->size = res;
+        #if MICROPY_PY_ML_STAI
```
iabdalkader, line 368: This module is meant to be generic, and I've put a lot of effort into making it so, you're now undoing this by re-introducing special cases here. Also, uma alloc here guarantees Nothing if blocks change or we add a new board.

iabdalkader (commented):
I'm redoing this, this is a bad fix.

iabdalkader:
Redoing this.

kwagyeman:
Okay, cool, yeah, I was on the fence about sending a PR to fix this given the models should be loaded via romfs. But, the changes aren't large to make it just work so we might as well.

PR #3278, opened 2026-09-21
common/umalloc: Minor umalloc fixes.

iabdalkader:
* Fix persistent allocs check for transient pools.
* Fix uma_malign size round-up.
* Add UMA unit-tests that reproduce the bugs above.


iabdalkader:
Added support for EXTERNAL flag.

common/umalloc.c
```diff
@@ -118,23 +118,26 @@ uma_pool_t *uma_pool_find(const void *ptr, size_t size, uint32_t flags) {
...
         }
         // Allow partial match unless strict matching was requested.
-        if (!strict && !partial && (uma_pools[i].flags & flags)) {
+        if (!strict && !partial && (attrs & flags)) {
             partial = &uma_pools[i];
         }
-        // Allow generic fallback unless strict matching was requested.
-        if (!strict && !fallback && !uma_pools[i].flags) {
+        // Fall back to the default pool unless strict matching was requested.
+        if (!strict && !fallback && (!attrs || (uma_pools[i].flags & UMA_DEFAULT))) {
```
kwagyeman: Is checking the `!attrs` here a fallback to handle boards that are missing `UMA_DEFAULT`?

common/umalloc.c
```diff
@@ -118,23 +118,26 @@ uma_pool_t *uma_pool_find(const void *ptr, size_t size, uint32_t flags) {
...
             continue;
         }
         // Never place persistent allocations in transient pools.
-        if (persist && (uma_pools[i].flags & UMA_TRANSIENT)) {
+        if (persist && (attrs & UMA_TRANSIENT)) {
             continue;
         }
-        // Exact attribute match (also handles flags==0 -> generic pool).
-        if (flags == uma_pools[i].flags) {
+        // Exact attribute match (also handles flags==0 -> default pool).
```
kwagyeman: // Exact attribute match (flags==0 matches attribute-less pools).

Nit on comment.

kwagyeman (approved):
Looks good , two minor comments.

common/umalloc.c
iabdalkader: Yes, though I ended up adding a default pool for every board, so it could be removed

kwagyeman: Okay

common/umalloc.c
iabdalkader: But the comment is accurate, with flags == 0 it matches the default pool.

iabdalkader: Ah right, because of EXTERNAL now. Okay.

kwagyeman (approved):

